PR #62, opened 2026-05-21
feat(examples/lambda-redact): promote skeleton to deployable Go submodule

sectoramen:
> **Stacked PR.** This branch is based on `feat/pkg-redact-public-api`
> (#61). When #61 lands on `main`, GitHub will auto-retarget this PR.
> The diff shown here will collapse to two commits at that point.

## What this PR does

Promotes `examples/lambda-redact/` from a build-tagged skeleton (where the `aws-lambda-go` import was commented out and the `//go:build examples_lambda` tag prevented the file from compiling as part of the main module) into a real, deployable Go submodule.

This is **PR 1 of the lambda-redact gateway work**. PR 2 will add a `terraform/` stack provisioning the supporting AWS infrastructure (API Gateway HTTP API, IAM role, log group, throttling, access logging with body-redacting format, etc.).

## Why split this from PR 2

PR 2's Terraform needs a stable handler shape and a verifiable build artifact to point at. Landing this PR first means PR 2's review surface stays focused on security-relevant infrastructure decisions (auth, throttling, log-format pinning, IAM least-privilege) without simultaneously reviewing handler code changes. The two changes are independently reviewable, and PR 1 is mechanical.

## What changed

| File | Change |
|---|---|
| `examples/lambda-redact/go.mod` | NEW — own module, depends on `aws-lambda-go` + `awslabs/ferret-scan`. `replace` directive points at `../..` so the submodule builds from a fresh monorepo clone. |
| `examples/lambda-redact/go.sum` | NEW |
| `examples/lambda-redact/handler.go` | REWRITE — drop `//go:build` tag, real `lambda.Start(handle)`, extract `buildResponse` as a pure function for unit testability, `FERRET_LOCAL=1` env var path for local invocation, replace hand-rolled `parseCSV` byte-loop with `strings.Split`+`TrimSpace`. |
| `examples/lambda-redact/handler_test.go` | NEW — 5 tests, 24 sub-cases including `TestBuildResponse_OmitsFindingsByDefault` pinning the side-channel guard. |
| `examples/lambda-redact/Makefile` | NEW — `build`, `test`, `vet`, `smoke`, `clean`, `help`. Lives at example root; invoked via `make -C examples/lambda-redact <target>`. |
| `examples/lambda-redact/.gitignore` | NEW — excludes `bootstrap`, `function.zip`, and the stray `lambda-redact` binary that a casual `go build` would produce. |
| `examples/lambda-redact/README.md` | REWRITE — real submodule documentation, `make -C` invocation pattern, env-var configuration table, customization recipe for downstream consumers. |
| `.github/workflows/go-test.yml` | New `lambda-redact-example` job parallel to the existing cross-platform test matrix. Runs `go vet` + `go test -race` + `make build` (cross-compile to `linux/arm64`) + verifies the build artifact exists and is non-empty. |
| `.pre-commit-config.yaml` | Broaden the `go.mod`/`go.sum` exclude regex from `^go\.(mod|sum)$` (root-only) to `(^|/)go\.(mod|sum)$` (any depth). The submodule's `examples/lambda-redact/go.mod` would otherwise have tripped the scanner's "unsupported file" exit code. |

### Engine reuse

The handler builds the `redact.Engine` exactly once in `init()` and reuses it for every invocation. Per-request setup cost is zero. With provisioned concurrency, the construction cost is paid during init phase rather than on the user's critical path.

### Side-channel guard

`Response.FindingsByType` is omitted from the JSON response by default. Per-type counts constitute a soft side-channel ("tenant X consistently submits PASSPORT data") even though no payload bytes are exposed. Operators who want them flip `FERRET_INCLUDE_FINDINGS=true`; the audit log always carries the full counts regardless.

The guard is implemented as a pure function:

```go
func buildResponse(
    redacted string,
    findingsByType map[string]int,
    durationMS int64,
    requestID string,
    includeFindings bool,
) Response {
    resp := Response{
        Redacted:   redacted,
        RequestID:  requestID,
        DurationMS: durationMS,
    }
    if includeFindings {
        resp.FindingsByType = findingsByType
    }
    return resp
}
```

Pinned by `TestBuildResponse_OmitsFindingsByDefault` and the symmetric `TestBuildResponse_IncludesFindingsWhenOptedIn`. A regression here fails CI rather than shipping.

### No payload logging

The handler writes only the audit record (counts + metadata) via `log.Printf` — never `req.Text` or `res.Redacted`. Sanitized error responses include a request ID for correlation but never echo the input or matched substring. Each error sentinel from `redact` is mapped to a generic user-facing category string.

### Safe-by-default suppressions

`AllowSuppressions` is intentionally empty. The public API's safe default; a multi-tenant gateway should never let a tenant pass through suppressions without per-tenant isolation. Code comment spells out the reasoning so a future contributor doesn't "helpfully" wire it up to a config file.

### Strategy validation

Unknown strategy strings produce a 400-style error rather than silent fallback. Pinned by `TestParseStrategy_RejectsInvalid`.

## Properties this change preserves

- **Parent `go.mod` has zero AWS dependencies**: `grep "aws-lambda-go\|aws-sdk-go" go.mod go.sum` returns nothing.
- **Parent `go build ./...` is identical** pre- and post-change. The submodule is invisible to the parent module's tooling.
- **`pkg/redact` is consumed only via its public API**; no `internal/` imports.

## Customization for downstream consumers

```bash
mkdir my-redact-gateway && cd my-redact-gateway
cp /path/to/ferret-scan/examples/lambda-redact/* .
go mod init my-redact-gateway
go get github.com/aws/aws-lambda-go@latest
go get github.com/awslabs/ferret-scan@latest
# Drop the `replace` directive at the bottom of go.mod — it's only
# there to keep the example buildable from a fresh monorepo clone.
```

## CI changes

A new `lambda-redact-example` job runs on every push/PR that touches `**.go`, `go.mod`, or `go.sum`:

1. `go vet ./...` against the submodule
2. `go test -race -count=1 ./...` against the submodule
3. `make build` cross-compiles to `linux/arm64` with the production flags (`lambda.norpc` tag, `-ldflags='-s -w'`, `-trimpath`)
4. Asserts the artifact exists and is non-empty (catches the case where the build silently succeeds but produces nothing)

The cross-compile is the most important step: a contributor who "improves" the imports without checking `arm64` can break the production target while local tests pass. CI catches that before deploy.

## What this PR does NOT do

- **No deployable Terraform stack.** That's PR 2.
- **No `make smoke` implementation.** The target ships in this PR but fails loud with a helpful message when `INVOKE_URL` is unset (because the URL only exists after PR 2's Terraform applies). The full smoke implementation (assertions on `.redacted` field, CC byte non-leak, email byte non-leak, `findings_by_type` absence) lands in PR 2.
- **No authentication / authorization, throttling, WAF, S3 async path.** All infrastructure concerns owned by PR 2 or a consumer's CDK / Terraform.

## Verification

- `go build ./...` (parent module): clean, no AWS deps leaked.
- `go test ./...` (parent module): green across all 24 packages.
- `make -C examples/lambda-redact vet`: clean.
- `make -C examples/lambda-redact test`: 5 test functions, 24 sub-cases, race-clean.
- `make -C examples/lambda-redact build`: produces `bootstrap` (~7.4 MB) + `function.zip` (~2.9 MB) for `linux/arm64`.
- `make -C examples/lambda-redact smoke`: fails loud with helpful message when `INVOKE_URL` unset (correct).
- `FERRET_LOCAL=1 ./examples/lambda-redact/bootstrap 'card 5500-...'`: runs end-to-end locally, redacts correctly.
- All pre-commit hooks pass. No skip flags used.

## Commits

| SHA | Subject |
|---|---|
| `a884255` | feat(examples/lambda-redact): convert from build-tagged skeleton to deployable Go submodule |
| `bc8bacb` | fixup(examples/lambda-redact): three small post-review fixes (`stat -c%s` → `wc -c < file` for cross-platform parity; README sizing-disclaimer note; `runLocal` usage message corrected from `./handler` to `./bootstrap`) |


sectoramen:
Closing — this work has been canceled and will not ship in this repo.

PR #165, opened 2026-07-22
feat(scan): UTF-16 + cp1252 support — Windows-native text files were invisible in file mode

sectoramen:
Follow-up to #164, closing the two Windows-encoding gaps found while testing it. Both verified **pre-existing** (reproduced on the pre-#164 binary).

## 1. UTF-16 was binary *by definition*
UTF-16-encoded ASCII carries a null byte in every other position — and both sniff sites rejected any null-bearing file before considering encoding. **Every PowerShell 5 `Out-File` transcript, regedit `.reg` export, and UTF-16 Windows log was silently invisible to all validators in file mode.**

Sniffing alone can't fix it (validators would receive null-riddled bytes no regex matches), so this adds a transcode layer (`internal/preprocessors/encoding.go`):
- **Detection**: BOMs (UTF-8, UTF-16 LE/BE) + a BOM-less UTF-16 heuristic (alternating-null pattern, strict wrong-side thresholds so UTF-32/binary don't slip through; UTF-32 BOM explicitly rejected)
- **Sniff**: transcodable encodings are decoded and judged as text *before* the null gate
- **Read path**: decodes to UTF-8 so validators see real text
- **Redaction — the critical half**: matches are produced against *decoded* text, so masking raw bytes would silently skip every match. Both redaction entry points (`RedactDocument`, `RedactContent`) decode before masking and **re-encode the output to the original encoding, BOM included** — a redacted `.reg` export stays a valid `.reg` export.

## 2. cp1252 smart quotes (0x80–0x9F)
Windows-1252 puts curly quotes/dashes/ellipsis below the 0xA0 floor the legacy fallback allowed — short smart-quoted lines from old Word/Outlook exports classified binary. Fallback now accepts the full high half, balanced by an ASCII-majority requirement so all-high-byte binary stays excluded.

## Verification (per user requirement: "thoroughly tested for everything, including masking and exceptions")
- **Detection matrix**: email+SSN+AWS-key content detected identically across utf-8, utf-8-bom, utf-16le±bom, utf-16be±bom, cp1252
- **Redaction matrix**: simple + format-preserving strategies — zero leaks, masking confirmed, **output encoding preserved** on every variant (locked by `TestRedactDocument_EncodingRoundTrip`, 6 encodings)
- **Exceptions**: empty file, 1-byte invalid file, BOM+truncated-rune file — clean exits, no panics; lone surrogates → U+FFFD; odd trailing byte dropped
- **Suppression flow + stdin gateway**: behavior identical across encodings (verified against utf-8 control)
- **Perf**: 12.6 MB UTF-16 file scans in 4.4 s
- **~45 unit subtests** (detection, round-trip incl. surrogate-pair emoji + CRLF, malformed inputs, sniff matrix); goldens byte-identical; full `./...` green

**Deliberately not supported** (documented in-code): UTF-32 (vanishingly rare), Shift-JIS/GBK/EUC-KR (need charset detection à la chardet — big dependency; save-as-UTF-8 is the workaround)

### DIFF
--- a/internal/preprocessors/encoding.go
+++ b/internal/preprocessors/encoding.go
@@ -1,0 +1,233 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package preprocessors
+
+import (
+	"strings"
+	"unicode/utf16"
+	"unicode/utf8"
+)
+
+// TextEncoding identifies the on-disk encoding of a text file, as detected
+// from its leading bytes. Only encodings ferret-scan can transparently decode
+// to UTF-8 (and re-encode on the redaction write path) are enumerated;
+// everything else is EncodingUnknown and handled by the legacy byte-level
+// heuristics.
+type TextEncoding int
+
+const (
+	// EncodingUTF8 is plain UTF-8 / ASCII — no transform needed.
+	EncodingUTF8 TextEncoding = iota
+	// EncodingUTF8BOM is UTF-8 with a leading BOM (Windows Notepad default).
+	// The BOM is stripped on decode and restored on encode.
+	EncodingUTF8BOM
+	// EncodingUTF16LE is UTF-16 little-endian with BOM (PowerShell 5
+	// Out-File default, regedit .reg exports, many Windows logs).
+	EncodingUTF16LE
+	// EncodingUTF16BE is UTF-16 big-endian with BOM.
+	EncodingUTF16BE
+	// EncodingUTF16LENoBOM is BOM-less UTF-16 LE, detected by the
+	// alternating-null heuristic.
+	EncodingUTF16LENoBOM
+	// EncodingUTF16BENoBOM is BOM-less UTF-16 BE.
+	EncodingUTF16BENoBOM
+	// EncodingUnknown means "not a transcodable encoding we recognize" —
+	// callers fall back to treating the bytes as-is.
+	EncodingUnknown
+)
+
+// String returns a short name for metadata/debug surfaces.
+func (e TextEncoding) String() string {
+	switch e {
+	case EncodingUTF8:
+		return "utf-8"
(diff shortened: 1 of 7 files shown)


sectoramen:
**Adversarial verification ran against this PR — two major regressions found and fixed in the second commit.**

Five parallel hostile verifiers (web path, public APIs, docker/build matrix, position tracking, blast radius) attacked the first commit. Results:

| Area | Verdict on commit 1 | Outcome |
|---|---|---|
| Position tracking / BOM offsets | ✅ clean — byte-exact across all hostile fixtures (line-1-col-0 after BOM, dup values, overlap case, all 3 strategies) | no action |
| Web UTF-16 handling, /export, panics | ✅ clean | no action |
| Docker/scratch, 6-target cross-compile, deps, Windows CI traps | ✅ clean | no action |
| Untouched pipelines (docx/wav/png routing, goldens, pre-commit, stdin) | ✅ byte-identical to main | no action |
| **Non-Latin codepages** | ❌ **cp1251/Greek/Hebrew/Arabic prose silently skipped** (my ASCII-majority gate — 4 of 5 verifiers converged on this independently) | **fixed**: historical floor restored; only the new 0x80–0x9F cp1252 allowance is ASCII-majority-gated |
| **FF FE-leading legacy text** | ❌ **decoded as UTF-16 → mojibake → reported CLEAN** (also a 2-byte scanner-evasion prefix) | **fixed**: BOM requires ≥1 null in body (real UTF-16 always has them) |

Accepted trade-off (minor): PCM-shaped binary now scans to 0 findings instead of being rejected — false-accept direction, no data loss.

All attacker repros re-run green post-fix (6 codepages × email/card, FF FE/FE FF prefixes), all original wins re-verified (UTF-16 ±BOM detect+redact+re-encode, cp1252 smart quotes, PNG exclusion), `TestAdversarialFindings` locks every repro, full suite green, goldens untouched.

PR #186, opened 2026-07-28
fix(preprocessors): emit audio and video metadata properties in sorted key order

sectoramen:
Audio and video metadata serialized a Go **map** straight to the extracted-text output by ranging over it. Map iteration order is randomized per run, so the extracted text — and therefore **every finding's line number and the byte-for-byte redaction output** — changed run to run on the same binary. For a redaction tool that is a correctness bug: unstable line numbers mean unstable redaction output and unstable suppression hashes over unchanged input.

> **Scope changed after a rebase onto `main` (`a8a4a39`).** The Office half of this fix landed independently via #181, which needed `FormatPropertiesMap` sorted to make container extraction reproducible. Both Office files are now **byte-identical to `main`** on this branch; what this PR adds for them is the regression test #181 shipped without. See [the rebase comment](https://github.com/awslabs/ferret-scan/pull/186#issuecomment-5106925996) for the full re-verification.

## Verified on real files

- A real `.m4a` whose `EncodingTool` line jumped position between runs.
- A real `.mov` whose HIGH-confidence GPS finding moves **line 9 <-> 8** across runs of the same binary — `main` splits 2/8 over 10 runs, this PR is 1 distinct.
- A real `.mp4` whose entire output hashed differently across runs.

The line shift is not cosmetic: on `main` the **suppression hash** for that unchanged file changes with it (`3c2f01f5...` vs `db93a781...`), so a captured baseline silently stops matching on the next run.

## Fix

The image metadata path already sorted its keys (`GetSortedKeys`) for exactly this reason; audio and video never got the same treatment. This brings them in line:
- `audiolib` `AudioMetadata.ToProcessedContent`
- `videolib` `VideoMetadata.ToProcessedContent`

Plus the GPS coordinate fix in commit 2 (first-complete-pair-wins, so the container's authoritative `(c)xyz` atom is not overwritten by a free-text scrape), and the `FormatPropertiesMap` regression test.

**Why sort is the right tool here, and its cost:** these emit a map to output *once per file*, so `O(n log n)` on a few dozen keys is microseconds against multi-millisecond extraction. Interleaved best-of-5 fix/base ratios: `.m4a` x0.983, `.mov` x1.017, `.mp4` x1.006. Scaling at n=100/200/400/800 gives per-doubling ratios 1.79 / 1.80 / 1.80 — n log n, not quadratic. (Argmax and per-match hot-path map iteration are a different category with different fixes — tie-break comparators and natural-order preservation respectively — and are not touched here.)

## Tests

Non-vacuous, re-checked after the rebase by copying the new test files onto `main`: both `ToProcessedContent` tests and both GPS tests **fail on `main`**. The `FormatPropertiesMap` test passes there, which is the correct reason to keep it — it guards a fix that currently has no test.

`gofmt` / `go build` / `go vet` / `GOOS=windows` / `GOOS=linux` / `-race` / full suite (53 packages) all clean. Golden corpus 5/5 consecutive passes. **Golden impact: none.** Finding sets identical to `main` across all fixtures — 0 new FPs, 0 lost TPs.

## Out of scope (verified, not this PR)

- Redacted `.docx` **container** bytes still vary, but the unzipped payload is byte-identical (`diff -r` clean) — zip metadata, which is #190.
- `.mov`/`.m4a` redaction erases findings entirely — that is #201, confirmed fixed when stacked with it.
- `gitlab-sast` "Additional Information" bullet order — #189. Stacked #189 + this PR: all 7 formats 1 distinct / 10 runs.

## Scope note

This PR is the concrete, verified subset of a broader "PR-5" grab-bag. The phone country-code comparator I investigated and **proved is not a determinism bug** (distinct same-length codes cannot both prefix-match, and the length-desc sort handles nesting deterministically — confirmed empirically over 100 runs). Help-text items were speculative and dropped for lack of a demonstrable defect.

Merges clean against `main` and all 15 other open PR branches.

### DIFF
--- a/internal/preprocessors/meta-extractors/meta-extract-audiolib/audio-metadata.go
+++ b/internal/preprocessors/meta-extractors/meta-extract-audiolib/audio-metadata.go
@@ -6,6 +6,7 @@ package audiolib
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 )
@@ -144,9 +145,18 @@ func (am *AudioMetadata) ToProcessedContent() string {
 		content.WriteString(fmt.Sprintf("Engineer: %s\n", am.Engineer))
 	}
 
-	// Additional properties
-	for key, value := range am.Properties {
-		if value != "" {
+	// Additional properties, emitted in sorted key order. Ranging over the map
+	// directly made the extracted text (and therefore every finding's line
+	// number, and the byte-for-byte redaction output) vary run to run — verified
+	// on a real .m4a where EncodingTool jumped lines between runs. The image
+	// metadata path already sorts its keys for exactly this reason.
+	propKeys := make([]string, 0, len(am.Properties))
+	for key := range am.Properties {
+		propKeys = append(propKeys, key)
+	}
+	sort.Strings(propKeys)
+	for _, key := range propKeys {
+		if value := am.Properties[key]; value != "" {
 			content.WriteString(fmt.Sprintf("%s: %s\n", key, value))
 		}
 	}
--- a/internal/preprocessors/meta-extractors/meta-extract-audiolib/determinism_test.go
+++ b/internal/preprocessors/meta-extractors/meta-extract-audiolib/determinism_test.go
@@ -1,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package audiolib
+
+import (
+	"testing"
+)
+
+// TestToProcessedContent_DeterministicPropertyOrder is the regression test for
+// the audio-metadata emit nondeterminism. ToProcessedContent ranged over the
+// Properties map directly, so the extracted text — and therefore every finding's
+// line number and the byte-for-byte redaction output — varied run to run.
+// Verified originally on a real .m4a where EncodingTool jumped lines between runs.
+func TestToProcessedContent_DeterministicPropertyOrder(t *testing.T) {
+	am := &AudioMetadata{
+		Filename: "sample.m4a",
+		Properties: map[string]string{
+			"EncodingTool":    "com.apple.VoiceMemos",
+			"FileExtension":   ".m4a",
+			"FilePath":        "/tmp/sample.m4a",
+			"MajorBrand":      "M4A",
+			"MinorVersion":    "0",
+			"CompatibleBrand": "isom",
+			"CreationTime":    "2026-01-01",
+			"Encoder":         "example-encoder",
+			"HandlerType":     "soun",
+			"MediaLanguage":   "und",
+		},
+	}
+
+	first := am.ToProcessedContent()
+	for i := 0; i < 300; i++ {
+		if got := am.ToProcessedContent(); got != first {
+			t.Fatalf("iter %d: audio ToProcessedContent order is not stable:\n--- first ---\n%s\n--- iter %d ---\n%s", i, first, i, got)
+		}
+	}
+}
--- a/internal/preprocessors/meta-extractors/meta-extract-videolib/determinism_test.go
+++ b/internal/preprocessors/meta-extractors/meta-extract-videolib/determinism_test.go
@@ -1,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package metaextractvideolib
+
+import (
+	"testing"
+)
+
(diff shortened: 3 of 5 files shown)


sectoramen:
Heads-up on merge order: this branch conflicts with #181 (`fix/container-determinism`) — 1 hunk in `internal/preprocessors/office_metadata_preprocessor.go` and 2 in `internal/preprocessors/shared_utilities.go`. Whichever of the two merges second needs a rebase. Verified by merging each open PR into `5acc670` in sequence: every other open PR merges clean, this pair is the only collision.

Separately, see #195 — #178 + #182 + #189 merge with zero git conflicts but leave `TestGoldenScanFormats` failing on two fixtures, so the last of those three to merge must run `UPDATE_GOLDEN=1 go test ./internal/goldencorpus/...`.

sectoramen:
Rebased onto `main` (`a8a4a39`) — this PR had gone `CONFLICTING`.

**Why it conflicted, and what changed.** The Office half of the first commit landed independently on `main` via #181, which needed `FormatPropertiesMap` sorted to make container-file extraction reproducible. Both `shared_utilities.go` and `office_metadata_preprocessor.go` conflicted on **comment text only** — the code was already functionally identical. I resolved both to `main`'s version, so this branch now shows **zero diff** against `main` for those two files.

What that leaves this PR doing:

- **audiolib `AudioMetadata.ToProcessedContent`** — sorted key emit (unique to this PR)
- **videolib `VideoMetadata.ToProcessedContent`** — sorted key emit (unique to this PR)
- **the regression test `main` shipped without** for `FormatPropertiesMap` (#181 fixed the code but added no test for it)
- **the GPS coordinate fix** in commit 2, unchanged

Commit 1's message was reworded to say this accurately rather than still claiming the Office changes.

**Non-vacuity, re-checked after the rebase.** Copying the three new test files onto `main` unmodified:

```
ok    internal/preprocessors                          <- FormatPropertiesMap test PASSES (main has the fix)
FAIL  .../meta-extract-audiolib   TestToProcessedContent_DeterministicPropertyOrder
FAIL  .../meta-extract-videolib   TestToProcessedContent_DeterministicPropertyOrder
FAIL  .../meta-extract-videolib   TestSearchForGPSInMetadata_DeterministicCoordinate
FAIL  .../meta-extract-videolib   TestSearchForGPSInMetadata_DoesNotOverwriteContainerCoordinate
```

The audio/video tests fail on `main` (this PR's unique fix); the `FormatPropertiesMap` test passes there, which is the correct reason to keep it — it guards a fix that currently has no test.

**Runtime evidence on real media, same binary, 10 runs each** (timing fields excluded):

| file | `main` | this PR |
|---|---|---|
| `voice.m4a` | 1 distinct | 1 distinct |
| `clip.mov` | **2 / 8 split** | **1 distinct** |
| `out.mp4` | 1 distinct | 1 distinct |

The `clip.mov` flap is a HIGH-confidence GPS finding whose `line_number` moves 9 ↔ 8 between runs of the same binary on the same file.

**Downstream damage that shift causes on `main`** — the suppression rule for that unchanged file is not stable:

```diff
-      hash: 3c2f01f5d70f02b76502336af34be9334d1af8ce0d58eab0af380572cc1028b7
+      hash: db93a781158411d5fa0aa000caabb1ee69867248f6b4b2e17412e8c07477c9bd
-        line_number: "8"
+        line_number: "9"
```

Same 2/8 ratio. A baseline captured on one run silently stops matching on the next. With this PR: 10/10 identical.

**Interaction with #189.** Under #189 alone, `gitlab-sast` output still varies — the residual is exactly this line shift and the `id` derived from it. #189 + this PR stacked (verified conflict-free): all 7 formats **1 distinct / 10 runs**.

**Scope checks.** Two things I confirmed are *not* this PR's to fix:
- Redacted `.docx` container bytes still vary, but the **unzipped payload is byte-identical** (`diff -r` clean) — that's zip metadata, which is #190.
- `.mov`/`.m4a` redaction erases findings entirely — that's #201, verified fixed when stacked with it.

**Gates.** `gofmt` clean (`internal/config/config.go` is unformatted on `main` too — pre-existing, separate). `go build` / `go vet` / `GOOS=windows` / `GOOS=linux` all clean. `-race` clean. Full suite 53 packages exit 0. Golden corpus 5/5 consecutive passes, no golden updates. Finding sets identical to `main` across all five fixtures — 0 new FPs, 0 lost TPs.

**Timing**, interleaved best-of-5: `.m4a` ×0.983, `.mov` ×1.017, `.mp4` ×1.006. Scaling on the sorted emit at n=100/200/400/800 gives per-doubling ratios 1.79 / 1.80 / 1.80 — n log n, not quadratic.

Merges clean against `main` and all 15 other open PR branches.

PR #215, opened 2026-07-29
fix(personname): score address/company context by distance, not per line

sectoramen:
## A name and an address on the same line loses the name

The business (−20), product (−8) and geographic (−35) penalties in `analyzeSpecificPatternsLineGlobal` applied to the whole **line**, regardless of where the pattern sat relative to the name. Summed and clamped to −50 in `analyzeContextCached`, that takes a solid two-token name from 92 to **42** — below the ≥ 50 emit threshold.

```
"Sarah Brooks,Acme Inc,1425 Oak Drive,Springfield,IL"              -> 0 findings
"Patient Sarah Brooks, 42 River Road, admitted Monday"             -> 0 findings
"Contact Sarah Brooks at 1425 Elm Avenue for the Acme Inc account"  -> 0 findings
"Michael Rivera,Globex LLC,88 Lake Street,Portland,OR"             -> 0 findings
```

The same names score **92** with the address removed, and **100** with the address on its own line.

The geographic list is ordinary address vocabulary — `street road city state drive avenue boulevard` — so this fires on mailing records, shipping labels, clinical notes and CSV exports. That's the shape of the data this tool exists to find.

## It's a leak, not a scoring quirk

An unreported name is never handed to the redactor. On a two-row CSV of names and addresses the CLI reported nothing and wrote **no redacted file at all** — both names survived in cleartext. After the fix:

| strategy | output |
|---|---|
| `simple` | `[PERSON-NAME-REDACTED],Acme Inc,1425 Oak Drive,…` |
| `format_preserving` | `************,Acme Inc,1425 Oak Drive,…` |
| `synthetic` | `Jann Dunham,Acme Inc,1425 Oak Drive,…` |

## Why this is a bug and not tuning

The per-match **negative-keyword** penalty has always been proximity-gated (`distance < 15`). These three families were the only context signals applied to the whole line at any distance.

They were also largely **redundant**: 13 of the 19 geographic patterns and 17 of the business patterns are *also* `negativeKeywords`. Those words were scored twice — once correctly, gated, and once line-globally.

## The fix

Locate the patterns per line (work stays hoisted, validator stays linear) but **apply** each penalty only when the pattern sits within `specificPatternProximity` bytes of the match — the same shape the negative-keyword path already uses. Boosts stay line-global; a boost cannot hide a finding.

## The window is measured, not guessed

Swept against 10 real names carrying addresses and 24 place/org/address decoys:

| window | real names found | decoys reported |
|---|---|---|
| 15 | 10/10 | 8/24 |
| 22 | 10/10 | 6/24 |
| **23–25** | **10/10** | **5/24** ← plateau |
| 26 | 9/10 | 5/24 |
| 30 | 8/10 | 5/24 |
| line-global | 4/10 | 5/24 *(old behavior)* |

**5/24 is exactly what the old code reported, on the same five lines.** So this recovers every real name while introducing **zero** new false positives. 25 sits at the top of the plateau, one byte before recall starts dropping.

The five residual decoys (`Phoenix Arizona`, `Menlo Park`, …) are a **separate pre-existing issue**, deliberately not fixed here: those place words are themselves entries in the *first-name database*, so they score on their own merits and the geographic penalty was only masking them incidentally — at the cost of deleting real names. No window value separates the two classes, because the geographic word sits outside the matched span in both (20–21 bytes away in the decoys, 29–31 in the real names). Fixing that means curating the name database.

## Also: the geographic scan was map-ordered

It ranged `geoPatternsMap` directly and `break`ed on the first hit, so map order picked the winner. The score was identical either way — which is why it was invisible — but the recorded byte **offset** was not order-independent, and the offset now decides whether a name is penalized. Added `getSortedGeoPatterns` alongside the three existing sorted accessors.

## Tests

- **`TestAddressOnSameLineKeepsTheName`** — 10 leak cases. Verified **6/6 fail on the pre-change validator and 0/6 fail after**, using a behavioral-only copy that compiles against both (the mechanism tests reference new internals, so they can't prove this on their own).
- `TestPlaceNamesStaySuppressed` — the FP direction, documenting which five lines are excluded and why.
- `TestSpecificPatternPenaltyIsProximityGated` — pins the mechanism (same words, same name, only distance varies), so a revert to line-global fails even if the corpora happen to pass.
- `TestGeoPatternIterationIsDeterministic` — runs the scan 50×, and asserts the accessor is actually sorted.
- `TestFirstWordKeywordIndexMatchesContainsWordKeyword` — locks the invariant the gate depends on: the new index helper and the existing boolean helper must agree, including on substring traps (`state` in `estate`, `drive` in `driven`).

## Verification

- Full suite `exit=0` (60 packages); `-race` clean on `personname` and `goldencorpus`; `gofmt`/`vet`/`build` clean.
- `UPDATE_GOLDEN=1` → **zero golden diff**, non-vacuous against **98 `PERSON_NAME` assertions across 9 golden files**.
- The `personname` complexity guard still passes, so the per-line hoist is intact and this didn't reintroduce per-match line scanning.
- Merges clean in order behind #207 → #208 → #209 → #210 → #211 → #212 → #213 → #214; the combined 9-branch stack builds and passes.

### DIFF
--- a/internal/validators/personname/geo_proximity_test.go
+++ b/internal/validators/personname/geo_proximity_test.go
@@ -1,0 +1,225 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package personname
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestAddressOnSameLineKeepsTheName is the leak gate.
+//
+// The business (-20), product (-8) and geographic (-35) penalties used to apply to
+// the whole LINE regardless of where the pattern sat relative to the name. Summed
+// and clamped to -50, that took a solid two-token name from 92 to 42, below the
+// >= 50 emit threshold — so an ordinary mailing record or CSV row lost the person's
+// name entirely. A name absent from the report is never handed to the redactor
+// either, so the cleartext value survives into the output: on a two-row CSV of
+// names and addresses the CLI reported nothing and wrote NO redacted file at all.
+//
+// Every line here carries a real person name AND an address and/or company on the
+// same line, which is the normal shape of the data this tool exists to find.
+func TestAddressOnSameLineKeepsTheName(t *testing.T) {
+	cases := []struct {
+		name string
+		line string
+		want string
+	}{
+		{"csv export row", "Sarah Brooks,Acme Inc,1425 Oak Drive,Springfield,IL", "Sarah Brooks"},
+		{"clinical note", "Patient Sarah Brooks, 42 River Road, admitted Monday", "Sarah Brooks"},
+		{"prose with company", "Contact Sarah Brooks at 1425 Elm Avenue for the Acme Inc account", "Sarah Brooks"},
+		{"pipe-delimited", "Sarah Brooks | Acme Corp | 1425 Elm Street", "Sarah Brooks"},
+		{"second csv row", "Michael Rivera,Globex LLC,88 Lake Street,Portland,OR", "Michael Rivera"},
+		{"shipping label", "Ship to Jennifer Alvarez, 210 Mountain Road, Boulder CO", "Jennifer Alvarez"},
+		{"employee record", "Employee: Daniel Fischer, 9 Valley Drive, ext 4417", "Daniel Fischer"},
+		{"prose address", "Karen Whitfield lives at 77 Creek Boulevard in the city", "Karen Whitfield"},
+		{"invoice line", "invoice for Thomas Bennett, 5 County Road, Acme Industries", "Thomas Bennett"},
+		{"suite address", "Deborah Callahan, 1200 State Street, Suite 300", "Deborah Callahan"},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			v := NewValidator()
+			matches, err := v.ValidateContent(tc.line, "record.csv")
+			if err != nil {
+				t.Fatalf("ValidateContent: %v", err)
+			}
+			if len(matches) == 0 {
+				t.Fatalf("no findings: the name %q is on this line together with an "+
+					"address/company, and a line-global context penalty dropped it. An "+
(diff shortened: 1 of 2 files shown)


sectoramen:
## Follow-up commit: closed a penalty-evasion hole in this PR

Adversarially re-reviewed the proximity gate and found a real hole **in my own change**, now fixed in `49f2840`.

`specificPatternIndices` recorded only the **first** matching pattern per family (a `break` after the first hit). Harmless while the penalty was line-global — the score was the same whichever pattern won — but under proximity gating the recorded **offsets** decide whether a name is penalized. So one distant match could hide an adjacent one:

| line | geo offsets | penalty |
|---|---|---|
| `oak drive sarah brooks` | `[4]` | **fires**, −50 |
| `avenue` + 60 dots + ` oak drive sarah brooks` | `[0]` | **silent**, −15 |

Identical adjacent `oak drive` in both. Prepending the alphabetically-earlier `avenue` far away suppressed the penalty entirely; through the CLI the finding rose from **57 to 69**. Attacker-controllable, and the same class of hazard as the map-order issue this PR already fixed — what changed was never the score for a given pattern, only *which* offset got recorded. The first commit's own comment flagged the risk without closing it.

Fix: collect every offset per family. The penalty is still awarded at most once per family, so this can't double-penalize — it only stops one offset from hiding another.

**No precision cost**, re-measured on the same corpora: 10/10 real names still found, 19/24 decoys still suppressed, same five residual place-name decoys.

`TestDistantPatternCannotMaskAnAdjacentOne` is non-vacuous — with the `break` restored it fails with the measured numbers (`control=-50.0 padded=-15.0`).

Also independently re-verified the original fix while checking this: the leak reproduces on pre-fix code, `w=25` passes the leak gate and `w=26` fails it (so the window is exactly at the boundary the comment claims), and `UPDATE_GOLDEN=1` still produces zero diff.

Suite `exit=0` (58 packages), `-race` clean, complexity guard still passes.

rustic (approved):

PR #218, opened 2026-07-29
perf(preprocessors): hoist line offsets out of the mapping loop

sectoramen:
## Document extraction was quadratic

`createBasicLineMappings` called `CalculateAbsoluteOffset` **twice per line, with identical arguments**. That helper re-runs `splitLines` over the **entire document** on every call — so a 10k-row spreadsheet performed 20,000 full-document splits.

| lines | before | after |
|---|---|---|
| 1,250 | 166ms | 1ms |
| 2,500 | 607ms | 1ms |
| 5,000 | 2.79s | 2ms |
| **10,000** | **10.15s** | **3ms** — 3,385× |

~3.6–4.6× per doubling before, where linear is 2×. ~1.9× after.

## The fix

A prefix sum computed once per document — the same arithmetic the helper performs internally, hoisted out of the loop. The plaintext preprocessor **already does exactly this** (`preCalculateLineOffsets`), so this brings the two extraction paths into line rather than inventing a pattern.

Exactly equal, not approximately: for `charPos=0` and `1 <= line <= len(lines)`, `CalculateAbsoluteOffset` returns the sum of `len(lines[i])+1` for `i < line-1`. The `charPos` clamp is a no-op at 0, and the `line > len(lines)` early return is unreachable because `lineNum` indexes `lines`. Both call sites passed the same arguments, so collapsing them to one lookup is also a plain common-subexpression elimination.

## One fix covers every document format

`createPDFPositionMappings` and `createOfficePositionMappings` both delegate to this function, so this covers `.xlsx/.docx/.pptx/.odt/.ods/.odp/.pdf`. Measured end-to-end through the CLI, **byte-identical output in every case**:

| input | before | after | findings |
|---|---|---|---|
| 10k-row `.xlsx` | 10,100ms | **759ms** (13.3×) | 19,873 both |
| 2.5k-row `.xlsx` | 1,149ms | 1,000ms | 4,968 both |
| 4k-paragraph `.docx` | 1,314ms | **274ms** (4.8×) | 3,950 both |

## The reported location was wrong

This was filed as a quadratic in `content_router.go`'s repeated `strings.Split`, inside the `file_evaluation` timing span. Both are red herrings: `RouteContent` costs ~1ms on a 1MB sheet and its `Split` calls are once-per-invocation, while `file_evaluation` is an observability span wrapping a fan-out to the preprocessors. The cost was one layer down. **`content_router.go` is untouched here.**

## Tests

`TestLineMappingOffsetsMatchCalculateAbsoluteOffset` keeps `CalculateAbsoluteOffset` as the **oracle** — it has no production callers left, but it's the definition the fast path must match — and compares every mapping against it across the shapes where an off-by-one would hide: empty text, no trailing newline, consecutive blank lines, and multi-byte runes (offsets are *byte* offsets, so a rune-counting mistake surfaces there). Zero differences.

`TestCreateBasicLineMappingsIsLinear` is the regression guard, with a **non-vacuity floor** asserting one mapping per line at both sizes before the clock is read. Verified non-vacuous — against the pre-change code it fails on *both* assertions:

```
4x input took 10.335298792s (> 2s) — CalculateAbsoluteOffset is probably being
  called per line again, which re-splits the whole document each time
4x input took 16.4x longer (base=629.769958ms big=10.335298792s)
```

while the equivalence test passes on both versions — correct, since it's the oracle.

## Deliberately not in this PR

The `PositionMappings` this builds have **no consumer anywhere** outside `internal/preprocessors` (`GetPositionMappingsForRange` has zero callers), so the whole structure is arguably dead work and deleting it would beat optimizing it. That's a dead-code change touching a struct field with a `json` tag, and belongs with the dead-code audit sequence rather than a behavior-preserving perf fix. `CalculateAbsoluteOffset` is likewise left in place.

## Verification

- First-party suite `exit=0` (58 packages); `-race` clean on `preprocessors` and `goldencorpus`; `gofmt`/`vet`/`build` clean.
- Golden: the suite passes with the goldens **unchanged**, which is stronger than regenerating them — zero files modified.
- Merges clean in order across the full 12-branch stack; combined stack builds and passes.

### DIFF
--- a/docs/testing/TEST_PLAN.md
+++ b/docs/testing/TEST_PLAN.md
@@ -218,6 +218,38 @@ Work lands as a conflict-free sequence, so:
 When in doubt, run more. The cost of an extra `go test ./...` is seconds; the cost of a
 shipped redaction bypass is a cleartext leak.
 
+### Ask what SINK consumes what you changed
+
+The table above is easy to recite and easy to misapply. Reciting a dimension is not
+running it. The reliable question is not "which row am I in?" but **"what downstream sink
+consumes the thing I just changed?"**:
+
+| Changed | Sinks that must be exercised |
+|---|---|
+| Byte offsets / spans | redaction (writes at those offsets) **and** the suppression hash (embeds position) |
+| Match order | redaction — replacement is a destructive sequential rewrite, so order decides the result |
+| Confidence | suppression hash (embeds `%.2f`) **and** every band filter (`--confidence`, exit codes) |
+| Match text | `--show-match`, the suppression `context_hash`, redaction token length |
+| Detection (new/lost findings) | redaction output, `stats.suppressed` counts, all 7 formats |
+
+**Before opening any PR, write the checklist out and mark every dimension either
+`RAN <command> → <result>` or `N/A because <reason>`.** "Probably unaffected" is not a
+reason. This exists because a perf change that only moved offset *computation* shipped
+with no redaction or suppression run: scan output was byte-identical, which felt like
+proof, but the offsets it rewrote are exactly what the redactor writes at and what the
+suppression hash embeds.
+
+Two traps specific to this repo:
+
+- **Container formats.** Grepping a redacted `.xlsx`/`.docx` for cleartext PII searches
+  *compressed* bytes and returns zero — indistinguishable from a clean pass. Read the part
+  inside the zip (`xl/sharedStrings.xml`, `word/document.xml`) and also count redaction
+  tokens, to prove the redactor rewrote content instead of copying the container through.
+- **Suppression must be tested cross-binary.** Generate the suppression file with the
+  **parent** binary and apply it with the **fixed** one. Same-binary round-tripping passes
+  even when the hash inputs changed; the cross-binary run is what proves existing
+  suppression files in the wild still match.
+
 ---
 
 ## Security reporting
--- a/internal/preprocessors/line_mapping_offsets_test.go
+++ b/internal/preprocessors/line_mapping_offsets_test.go
@@ -1,0 +1,143 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package preprocessors
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+)
+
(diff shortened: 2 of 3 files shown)


sectoramen:
## Redaction + suppression verified (was missing — thanks for the catch)

My original verification proved byte-identical **scan** output but never exercised redaction or suppression. That was the wrong gap to leave on this change specifically: the offsets this PR rewrites feed the redactor and are hashed into suppression entries, so an off-by-one would corrupt output or silently invalidate every existing suppression file rather than fail a test.

Both now checked, on `.xlsx` **and** `.docx` (both Office paths funnel through the changed function), parent vs fixed:

### Redaction — byte-identical

| input | strategy | parent | fixed | identical |
|---|---|---|---|---|
| `.xlsx` | simple | 4973B | 4973B | **yes** |
| `.xlsx` | format_preserving | 5806B | 5806B | **yes** |
| `.docx` | simple | 1831B | 1831B | **yes** |
| `.docx` | format_preserving | 2444B | 2444B | **yes** |

Synthetic is non-deterministic by design, so compared by content instead: 896 findings and a redacted file on both, no cleartext card in either.

Verified the redactor actually rewrote the document internals rather than passing the container through — inside `xl/sharedStrings.xml`:

| | cards | surviving SSN-shaped | `REDACTED` tokens |
|---|---|---|---|
| original | 300 | 300 | 0 |
| parent | **0** | 4 | 896 |
| fixed | **0** | 4 | 896 |

The 4 survivors are identical on both and are **pre-existing and correct**: they all have a `00` group (`400-00-0000`, `440-00-0240`, …), which is invalid per SSA rules, so the validator never reports them and the redactor correctly leaves them. Confirmed directly — scanning those four plus one valid SSN yields exactly one finding.

### Suppression — cross-binary hash compatibility

The important direction: **generate with the parent binary, apply with the fixed one.**

```
rules generated by parent: 2688 (896 enabled)
ferret-parent: remaining_findings=0  [SUPP] lines=896
ferret:        remaining_findings=0  [SUPP] lines=896
```

Suppressions written before this change still match after it, and the findings are **counted as suppressed** rather than dropped. Since the suppression hash embeds position data, this is the check that proves the prefix sum reproduces the old offsets exactly in the place where a mismatch would be invisible to a unit test.

Net: this PR is behavior-neutral across scan output, all three redaction strategies, and suppression — it only removes the quadratic.

rustic (approved):

PR #222, opened 2026-07-29
test(ssn): cover labelled SSNs in prose, the class a proposed FP fix would delete

sectoramen:
Test-only. **No behavior change.** This is a gate, added before the fix it guards against.

## The false positives are real

Real public data — a [World Bank population CSV](https://raw.githubusercontent.com/datasets/population/main/data/population.csv) containing **no PII at all** — produces **2,161 HIGH-confidence SSN findings**, because the tabular path skips the LOW cap for any ≥3-field delimited line.

A tempting one-line fix is a "decimal guard": reject a digit run adjacent to a `.` on a word boundary, since `1.130075728` is a fraction.

## That guard is a cleartext leak

`\b` ends **before** a sentence-terminal period, so `the SSN is 130075728.` is indistinguishable from a decimal fraction under a boundary test. Measured on the shipped validator, then with the guard applied:

| input | shipped | with guard |
|---|---|---|
| `Employee record: the SSN is 130075728.` | HIGH 100 | **no finding** |
| `Employee SSN: 130-07-5728.` | HIGH 100 | **no finding** |
| `Social Security Number: 130 07 5728.` | HIGH 100 | **no finding** |
| `ssn=130075728.` | HIGH 95 | **no finding** |

These are the **highest-confidence shapes this validator has** — an explicit label adjacent to a structurally valid value — and the guard sits *before* `isValidSSN`, so it removes them outright rather than re-banding. An unreported value is never handed to the redactor, so it stays in cleartext.

## Why this must be a committed test, not a note

Neither the `ssn` package suite nor the golden corpus contained a **single** sentence-terminal SSN. I implemented the guard and ran all three:

```
new coverage        -> FAIL (names the leak)
ssn package suite   -> ok
golden corpus       -> ok
```

Both gates that would have been cited as safety evidence are blind to it. A regression gate cannot prove anything about a class it does not contain — the same defect family as the vacuous complexity guards fixed in #213.

## The four tests

- **`TestLabelledSSNInProseIsDetected`** — 8 prose terminations (period, comma, semicolon, parenthesis, quote, `key=value`) with an explicit label. Verified non-vacuous: fails on all 8 under the guard.
- **`TestLabelledProseSSNKeepsHighConfidence`** — pins the **band**, not just presence. A fix that demoted these HIGH→LOW would hide them under `--confidence high,medium` and stop them failing pre-commit: a quieter version of the same leak.
- **`TestDecimalFractionsAreNotSSNs`** — the other half of the contract, recorded as **current** behavior rather than an aspiration, so a future precision fix has a measured starting point. Asserts only that these stay out of the HIGH band; a `t.Logf` reports what is reported today.
- **`TestProseAndTabularSSNsCoexist`** — both shapes in one document. This is the exact blind spot that made the broken fix look safe: the recall corpus used to justify it was **entirely delimited rows**, and a columnar-only corpus cannot see a prose regression.

## The FPs are still worth fixing — just not this way

The direction is the **column header**, which is currently ignored: `ssn\n130075728` and `population\n130075728` both score MEDIUM 70. That same header signal also closes a cross-line detection gap (a CSV header/data-row split currently yields zero findings and leaves a passport number cleartext in the redacted output). This commit is the gate that has to stay green while that work happens.

## Verification

First-party suite `exit=0` (58 packages); `-race` clean on `ssn`; `gofmt`/`vet`/`build` clean; **zero golden diff**. Merges clean in order across the full 16-branch stack.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build clean vet fmt run install-config install check-go-version
+.PHONY: build clean vet fmt run install-config install check-go-version pr-checklist
 
 # Default target
 all: check-go-version fmt vet build
@@ -44,6 +44,7 @@ help:
 	@echo "  test-git-commit    - Test real git commit blocking"
 	@echo "  test-cleanup       - Test cleanup functionality"
 	@echo "  container-test     - Test container health"
+	@echo "  pr-checklist       - Pre-PR audit: prints RAN/N-A/NEEDS-MANUAL per TEST_PLAN dimension"
 	@echo ""
 	@echo "🔧 Development:"
 	@echo "  clean              - Clean build artifacts"
@@ -222,6 +223,16 @@ vet:
 fmt:
 	@echo "Formatting..."
 	@go fmt ./...
+
+# Pre-PR self-audit: run every TEST_PLAN dimension and print an explicit
+# RAN / N-A / NEEDS-MANUAL line for each, so a skipped dimension is visible
+# instead of silently absent. Paste the output in the pull request.
+#
+# Not a CI gate — CI already runs the suite. The value is that it enumerates the
+# dimensions a reviewer cannot otherwise tell were considered, particularly the
+# redaction and suppression sinks.
+pr-checklist:
+	@./scripts/pr-checklist.sh $(BASE)
 
 # Run the application
 run: build
--- a/docs/testing/TEST_PLAN.md
+++ b/docs/testing/TEST_PLAN.md
@@ -218,6 +218,23 @@ Work lands as a conflict-free sequence, so:
 When in doubt, run more. The cost of an extra `go test ./...` is seconds; the cost of a
 shipped redaction bypass is a cleartext leak.
 
+### Run the checklist mechanically
+
+Reciting these dimensions is not the same as running them, and a dimension that is simply absent
+from a PR description is indistinguishable from one that was considered and dismissed. So enumerate
+them:
+
+```bash
+make pr-checklist              # vs origin/main
+scripts/pr-checklist.sh <ref>  # vs another base
+```
+
+It executes what it can and prints one line per dimension — `RAN`, `N/A because …`, or
+`NEEDS-MANUAL`. The sink dimensions (redaction, suppression) and the timing curve print
+NEEDS-MANUAL whenever production code changed, because those need a fixture that exercises *your*
+change and no script can invent one. Paste the output in the PR; leaving a NEEDS-MANUAL item unrun
+is how a cleartext leak ships.
+
 ---
 
 ## Security reporting
--- a/docs/testing/TEST_PLAN.md
+++ b/docs/testing/TEST_PLAN.md
@@ -218,7 +218,19 @@ Work lands as a conflict-free sequence, so:
 When in doubt, run more. The cost of an extra `go test ./...` is seconds; the cost of a
 shipped redaction bypass is a cleartext leak.
 
-### Run the checklist mechanically
+---
+
+## Security reporting
+
+A newly discovered vulnerability is reported to AWS/Amazon Security via the
+[vulnerability reporting page](http://aws.amazon.com/security/vulnerability-reporting/)
+**before** any public write-up — never as a public GitHub issue (`CONTRIBUTING.md`).
+A remediation for an *already-documented* threat-model item (e.g. a `THREAT_MODEL.md`
+row) is ordinary tracked work and ships as a normal PR.
+
+---
+
+## Run the checklist mechanically
 
 Reciting these dimensions is not the same as running them, and a dimension that is simply absent
 from a PR description is indistinguishable from one that was considered and dismissed. So enumerate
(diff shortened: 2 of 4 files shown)


sectoramen:
## Full TEST_PLAN checklist (you were right — I had skipped several)

What I originally ran: gofmt, vet, one full suite pass, golden diff, `-race` on `ssn` only. That left `-race` on the module, uncached runs, `-short`, flake repeats, redaction, suppression, all-formats and cross-platform unrun — and the sink dimensions are exactly the ones my own hardened TEST_PLAN says are non-negotiable. Complete run below, every line marked `RAN` or `N/A because`.

| # | dimension | result |
|---|---|---|
| 1 | build / vet / gofmt | **RAN** — `gofmt -l ./cmd ./internal ./pkg` empty; `go vet ./...` clean; `go build ./...` ok |
| 2 | full suite, **uncached** | **RAN** — `go test -count=1 ./cmd/... ./internal/... ./pkg/...` → **exit=0, 58 ok**, 12 no-test pkgs |
| 3 | golden corpus | **RAN** — `UPDATE_GOLDEN=1` → **0 files changed** |
| 4 | **redaction, all 3 strategies** | **RAN** — all 4 prose SSNs masked, **0 cleartext remaining** in simple / format_preserving / synthetic |
| 5 | **suppression** | **RAN** — 4 rules generated → enabled → **0 remaining, 4 `[SUPP]` lines** (counted, not dropped) |
| 6 | timing / O(n²) | **N/A because** no production code changed (0 non-test files); the added tests are single-line inputs |
| 7 | all 7 output formats | **RAN** — text 4, json 8, yaml 8, csv 4, junit 2, gitlab-sast 12, sarif 15 mentions |
| 8 | web / CSP | **N/A because** no web, template or asset file is touched |
| 9 | TP/FP vs ground truth | **RAN** — the PR *is* the TP corpus; FP side pinned by `TestDecimalFractionsAreNotSSNs` |
| 10 | **non-vacuity** | **RAN** — see below |
| 11 | cross-platform | **RAN** — `GOOS=windows/linux/darwin go vet` all rc=0; grep confirms no path separators, no wall-clock, no CRLF literals in the new file |
| 12 | flake | **RAN** — 5 uncached repeats of the `ssn` package, all ok |
| 13 | **race, whole module** | **RAN** — `go test -race -count=1 ./cmd/... ./internal/... ./pkg/...` → **exit=0, 58 ok** (previously I'd only raced `ssn`) |
| — | `-short` mode | **RAN** — `ssn` and `goldencorpus` both ok |
| — | dead code | **N/A because** the change is additive test code with no new production symbols |

### Non-vacuity, re-proved (dimension 10)

Re-implemented the rejected decimal guard and ran all three gates:

```
new tests            -> 4 failing subtests
pre-existing ssn suite -> ok
golden corpus          -> ok
```

That asymmetry is the entire argument for this PR: the two gates a future author would cite as evidence are blind to the leak, and these tests are not.

Working tree left clean (`git status --porcelain` → 0 tracked modifications) after every mutation.

Also worth recording: 4/4 prose SSNs redact to `[SSN-REDACTED]` with zero cleartext, which confirms the leak claim in the PR body is about *detection* (an unreported value is never handed to the redactor), not about the redactor itself.

sectoramen:
## Added `make pr-checklist` (commit `1466c7a`)

You pointed out I keep forgetting the tests. Rather than just re-running them for this PR, I made the omission **visible** — because the failure mode isn't ignorance of the plan, it's that "run the tests" collapses to `go test ./...`, and a dimension **absent** from a PR description is indistinguishable from one considered and dismissed.

```
$ make pr-checklist
1 gofmt                     RAN - clean
1 go vet                    RAN - clean
1 go build                  RAN - ok
2 full suite (-count=1)     RAN - 58 packages ok
3 golden regen              RAN - 0 files changed
4 redaction                 N/A - no production code changed
5 suppression               N/A - no production code changed
6 timing / O(n^2)           N/A - no production code changed
7 all 7 formats             N/A - no production code changed
8 web / CSP                 N/A - no web/template/asset file touched
10 non-vacuity              NEEDS-MANUAL - new tests MUST fail on the parent commit
11 cross-platform vet       RAN - linux darwin windows
12 flake (3 repeats)        RAN - ./internal/validators/ssn
13 race (whole module)      RAN - 58 packages ok
-- -short mode              RAN - ok
-- tree clean               RAN - this run dirtied nothing

```

**What it refuses to claim.** Redaction, suppression, timing, formats and non-vacuity print NEEDS-MANUAL whenever production code changed — each needs a fixture exercising the specific change, and no generic script can invent one. The messages carry the *traps*, not just the dimension name:

- containers: read the part **inside** the zip (grepping compressed bytes returns 0 and looks like a pass)
- suppression: generate with the **parent** binary and apply with yours (same-binary round-tripping passes even when the hash inputs changed)
- timing: non-vacuity floor plus distinct values (identical repeats let a quadratic measure as linear)

It doesn't duplicate `make test-ci`, which already does the race + uncached pass. That tells you the suite is green; this tells you which dimensions were *considered*.

### I tested the script and the Makefile target rather than assuming

- `shellcheck` and `bash -n` clean. Two real findings fixed, including an SC2015 `A && B || C` that is not if-then-else.
- **Found and fixed a bug in it:** the "tree clean" check reported the *caller's* own uncommitted work as dirty. It now snapshots the dirty set before running and diffs, so it reports only what the run itself touched. Verified with two pre-existing modified files present, and it correctly said "this run dirtied nothing".
- **Failure paths verified by injection**, not inspection: an unformatted file gave `1 gofmt FAIL`; a deliberately failing test gave `2 full suite FAIL` **and** `12 flake FAIL`; both exit non-zero. Clean tree exits 0.
- `BASE=<ref>` override works via the script argument and the Makefile; base equal to HEAD exits early with "nothing to check".
- Added to `make help` under Testing, and to `.PHONY`.
- Verified the stack still merges: the TEST_PLAN pointer added here coexists with the sink-gate section added in #218.

The one NEEDS-MANUAL item I proved by hand: re-implemented the rejected guard, confirmed the new tests fail while the pre-existing `ssn` suite and the golden corpus both pass. That asymmetry is the argument for this PR.

rustic (approved):

PR #226, opened 2026-07-29
fix(phone,driverslicense): stop a nearby keyword deleting a real finding

sectoramen:
## Problem

Two validators had a context check that **suppressed a finding to nothing** when a keyword appeared near the match, instead of adjusting its confidence.

That is a leak, not a scoring nit. Only reported findings are handed to the redactor, and **a file that yields no findings has no redacted output written at all** — so the value survives in cleartext in the file a redaction pipeline was supposed to sanitize. It is also attacker-controllable: anyone who can influence a byte of the document can type the magic word next to a real value and delete it.

Measured on a 4-line document containing two real phone numbers and two real licence numbers:

```
main:  files produced by --enable-redaction: 0     <- all 4 values leak
fix:   Callback [PHONE-REDACTED] (session token expired)
       Customer [PHONE-REDACTED] hash mismatch reported
       Driver License Number: [DRIVERS_LICENSE-REDACTED], road test scheduled 2026-08-01
       DL: [DRIVERS_LICENSE-REDACTED] issued CA, demo vehicle assigned
```

## PHONE

`isEmbeddedInIdentifierAt` vetoes on `{session, token, request, response, hash, uuid, guid, build, version, revision, timestamp, instance}` within 10 bytes of the match. Those are ordinary English words, so the window also fires on prose that merely mentions them near a real number — `Customer 212-555-0142 hash mismatch reported` reported nothing.

**The keyword list is not removable.** Each of the 12 was measured to be load-bearing against a bare digit run: with the keyword the finding is suppressed, without it the same digits are reported. So deleting the list costs real precision.

What the veto ignored is the match's **own punctuation**. Resource IDs, epochs and build numbers are contiguous digit runs (`i-057034242931`, `1735689600`, `20260714093012`); none of them is ever written `415-555-0123`, `(415) 555-0123` or `+1 415 555 0123`. A punctuated match near one of these words is a phone that happens to sit next to them. Bare digit runs stay suppressed.

| suite | main | fix |
|---|---|---|
| real phones near identifier words (should report) | 3/6 | **6/6** |
| genuine identifiers (should suppress) | 0/6 | **0/6** |

Two new false positives appear — `token 415.555.0123 issued` and `build (415) 555-0123 label` — and they are the definitional trade: those strings are genuinely ambiguous, and the punctuation rule is exactly what recovers `Customer 212-555-0142 hash mismatch`. Note the bands do **not** cleanly separate them (the recovered real phone lands LOW 58, one FP lands MEDIUM 65), so the justification is not "filter the FPs by band" — it is that every value is now *reported and therefore redactable* rather than silently dropped.

## DRIVERS_LICENSE

`strongSuppressKeywords` was checked with `containsKeyword` over the **whole line** and returned `-20` immediately. Against the base of 20 that lands on exactly 0, and the emit gate is `confidence <= 0`, so the finding was deleted. Two things were wrong:

- **No positional bound at all.** A keyword 400 characters away killed the finding as effectively as an adjacent one.
- **The early return discarded the positive evidence before counting it.** `ValidateContentCtx` only scans lines that already contain a DL keyword (`lineHasPositiveKeyword`), so every candidate reaching this check is already labelled. `Driver License Number: D1234567, road test scheduled` scores **95** without the phrase and scored **0** with it.

`road test`, `vision test`, `drug test`, `skills test`, `breath sample` and `sample collection` are ordinary DMV and HR vocabulary.

Suppression is now positional — it fires when the marker modifies **the licence**:

- **before the label**: `test DL: D1234567`, `example driver license D1234567`, `fake dl number D1234567`
- **opening an aside on the value**: `D1234567 (placeholder)`, `D1234567 // sample data`, `D1234567 <- example value`

A clause with its own subject after the value does not: `D1234567, drug test negative`, `D1234567 -- vision test passed`.

Suppression stays **absolute** when it fires (still `-20` against the base of 20). Only *which lines qualify* changed, so an example value still stays undetected.

| suite | main | fix |
|---|---|---|
| real licences near test vocabulary (should report) | 0/4 | **4/4** |
| genuine test data (should suppress) | 0/4 | **0/4** |
| held-out, 12 real + 12 test, written after the rule | — | **11/12 real reported, 0/12 test leaked** |

### Two alternatives rejected by measurement first

- **A pure distance threshold.** No cutoff separates the classes — real records land at distance 1 and test data at 2, interleaved all the way out. Best case was 10/11 real at the cost of leaking 5/10 test values.
- **A compound-noun list of DMV terms** (`road`/`vision`/`drug` before the keyword). Scored 10/10 on the cases I designed it against and **6/12 on held-out data**, where every single miss was a REAL licence classified as test data. Overfit to my own examples.

The one known miss is `Driver License Number: D1234567 sample submitted to lab`, where `sample` both opens the following clause and is a bare noun. It suppresses — matching the old behavior — rather than inventing a finding.

## Keeping the hoist (no DoS regression)

`AnalyzeContext` is deliberately called once per line, not per match, because a per-match full-line scan is `O(matches × line length)` — the single-long-line CPU-exhaustion shape `dos_test.go` guards.

The rule is split to preserve that: `markerBeforeLabel` stays a per-line invariant, and `markerOpensAsideAfter` reads only the few bytes following the span, using the offset the caller already has. Measured on one line with N licence numbers:

```
N=250  0.085s      N=500  0.091s      N=1000  0.105s      N=2000  0.134s
findings 250 / 500 / 1000 / 2000  (grows 1:1, so the guard is non-vacuous)
main 0.171s vs fix 0.126s at N=2000 — no regression
```

## Test plan

`scripts/pr-checklist.sh`: gofmt / vet / build clean, 58 packages, golden corpus **moves no snapshot**, cross-platform vet (linux/darwin/windows), 3× flake, whole-module `-race`, `-short`.

- **Redaction (dim 4)** — all 3 strategies, 0/4 cleartext remaining. Output shown above.
- **Suppression (dim 5)** — parent-generated rules still suppress under the new binary (2 rules → 0 surviving); the 4 newly-recovered findings generate 4 rules that suppress cleanly.
- **All 7 formats (dim 7)** — text/json/yaml/csv/junit/gitlab-sast/sarif each carry all 4 findings.
- **Timing (dim 6)** — multi-line curve to 8,000 findings and the single-long-line shape above, distinct values with a non-vacuity floor at every size.
- **Non-vacuity (dim 10)** — disabling the punctuation rescue puts every phone leak case red; reverting the DL rule to line-global puts every licence case red.
- Bounds cases for `markerOpensAsideAfter` (negative and past-end offsets) and `hasPhonePunctuation` (empty, `-`, `()`, `abc`).

### DIFF
--- a/internal/validators/driverslicense/suppress_position_test.go
+++ b/internal/validators/driverslicense/suppress_position_test.go
@@ -1,0 +1,278 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package driverslicense
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/awslabs/ferret-scan/v2/internal/detector"
+)
+
+// realRecordLines are labelled licence numbers on lines that also contain a
+// test/placeholder keyword in ordinary DMV, HR or lab vocabulary. Every one of
+// these reported NOTHING before the suppression rule was made positional.
+//
+// The consequence is a leak, not a scoring nit: only reported findings reach the
+// redactor, and a file with no findings has no redacted output written at all,
+// so the licence number survived in cleartext.
+var realRecordLines = []string{
+	"Driver License Number: D1234567, road test scheduled 2026-08-01",
+	"Driver License Number: D1234567 -- vision test passed",
+	"Driver License Number: D1234567 for the sample collection unit",
+	"DL: D1234567 issued CA, demo vehicle assigned",
+	"Applicant DL: D1234567 -- retest required in 30 days",
+	"Driver License Number: D1234567; hearing test waived",
+	"DL D1234567 passed the practical test on 2026-06-02",
+	"license number: D1234567, urine sample chain of custody",
+	"Driver License Number: D1234567, drug test negative",
+	"Driver License Number: D1234567 -- DOT physical and drug test on file",
+	"DL: D1234567, CDL skills test appointment 08/14",
+	"license number: D1234567 (CA) breath sample refused",
+	"Driver License Number: D1234567 issued 2019, road test waiver granted",
+	"Driver License Number: D1234567 for demo day shuttle driver",
+	"driver license D1234567 assigned to the testing fleet",
+}
+
+// testDataLines are genuine test/placeholder values. The keyword modifies the
+// licence itself — it sits before the label, or opens an aside on the value — and
+// these must stay unreported.
+var testDataLines = []string{
+	"test DL: D1234567",
+	"example driver license D1234567",
+	"sample license number: D1234567",
+	"DL: D1234567 placeholder value",
+	"mock DL D1234567 for demo",
+	"DL: D1234567 -- test record do not use",
+	"fake dl number D1234567",
+	"driver license D1234567 (example only)",
+	"DL: D1234567 // sample data",
+	"demo account driver license D1234567",
+	"DL: D1234567 <- example value",
+	"driver license D1234567 [test data]",
+	"# sample: driver license D1234567",
+	"DL: D1234567 fake",
+	"placeholder driver license D1234567",
+	"DL: D1234567 -- mock up only",
+	"uuid driver license D1234567",
+}
+
+// TestRealLicenceNearTestVocabularyIsReported is the leak this file exists for.
+func TestRealLicenceNearTestVocabularyIsReported(t *testing.T) {
+	v := NewValidator()
+
+	for _, line := range realRecordLines {
+		t.Run(line, func(t *testing.T) {
+			matches, err := v.ValidateContent(line, "records.txt")
+			if err != nil {
+				t.Fatalf("ValidateContent: %v", err)
+			}
+			if len(matches) == 0 {
+				t.Fatalf("a labelled licence was deleted by a test-vocabulary keyword "+
+					"elsewhere on the line; an unreported value is never redacted.\nline: %s", line)
+			}
+		})
+	}
+}
+
+// TestGenuineTestDataStaysUnreported is the other direction, and the constraint
+// that makes the fix acceptable: widening recall must not start reporting sample
+// data. An example value should stay undetected because it is not real.
+func TestGenuineTestDataStaysUnreported(t *testing.T) {
+	v := NewValidator()
+
+	for _, line := range testDataLines {
+		t.Run(line, func(t *testing.T) {
+			matches, err := v.ValidateContent(line, "fixtures.txt")
+			if err != nil {
+				t.Fatalf("ValidateContent: %v", err)
+			}
+			if len(matches) != 0 {
+				t.Errorf("genuine test data was reported as a real licence: %s\ngot %d match(es)",
+					line, len(matches))
+			}
+		})
+	}
+}
+
+// TestSuppressionIsPositionalNotLineGlobal pins the specific defect: the old
+// rule used containsKeyword over the whole line, so a keyword arbitrarily far
+// from the value killed the finding just as effectively as an adjacent one.
+func TestSuppressionIsPositionalNotLineGlobal(t *testing.T) {
+	v := NewValidator()
+
+	// 400 characters of filler between the licence and the keyword.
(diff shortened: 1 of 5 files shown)


sectoramen:
## Added: the adjacent-character path (`688b1d5`)

Reviewing my own PR body found it overstated. It claims *"a punctuated match near one of these words is a phone → 6/6"*, but `isEmbeddedInIdentifierAt` has a **second** path that runs before the keyword window, so punctuated phones were still being deleted:

```
Branch line Tel:415-267-1234                    0 findings
Escalation contact Madrid: +34 91 496 0345      0 findings
Dial 1-415-267-1234 for the after-hours desk    0 findings
```

Three defects, each with its own control:

**1. A leading `:` vetoed before the label vocabulary was consulted** — so the phone *label itself* was the veto, the worst possible polarity. `Tel:415-267-1234` reported nothing; `Tel: 415-267-1234` (one space) reported HIGH 100. It hit every label form: Tel, Phone, Fax, Mobile, Cell, Telephone, Contact. Directory listings, vCards and CSV exports routinely omit that space.

**2. `strings.HasSuffix(word, label)` matched city names.** `madrid` ends in `id`, so `Escalation contact Madrid: +34 91 496 0345` was read as an identifier label and dropped — likewise Cupid, Enid, Sid, Valladolid, and by symmetry any word ending in a *phone* label. Replaced with whole-word-or-separator-delimited-tail matching, which keeps `work-mobile` / `trace_id` / `x.session` working.

**3. A leading `-` swallowed NANP long-distance.** `ami-050451375729` should suppress and still does, but `1-415-267-1234` is how North American numbers are printed. Rescued only when the dash follows a **standalone** `1` — `sku-401-415-267-1234` still suppresses.

| | before | after |
|---|---|---|
| recall | 4/11 | **11/11** |
| false positives | 0/12 | **0/12** |

### A fourth fix, measured out rather than shipped

Allowing a **trailing** colon recovers `Reception 415-267-1234: ask for Dana` (1 finding) and admits **4 false positives**: `ratio 415-267-1234: 1`, `error 415-267-1234: connection refused`, `key 415-267-1234: value`, `range 100-200-3000: allowed` — every one a log line where `<value>:` is a key, which is exactly what this check exists to reject.

Gating it on `hasPhonePunctuation` doesn't separate them (the log keys are punctuated identically), and all five score **LOW 5.00%**, so confidence can't triage them either. A *leading* colon is rescuable because the preceding word says which kind of value follows; nothing after the match carries equivalent information. `TestTrailingColonStaysSuppressed` records this so it isn't "fixed" back.

### The sink, side by side

```
parent (this PR before 688b1d5):
  Branch line Tel:415-267-1234                    <- CLEARTEXT
  Escalation contact Madrid: +34 91 496 0345      <- CLEARTEXT
  Dial 1-415-267-1234 for the after-hours desk    <- CLEARTEXT
  Callback [PHONE-REDACTED] (session token expired)

with 688b1d5:
  Branch line Tel:[PHONE-REDACTED]
  Escalation contact Madrid: [PHONE-REDACTED]
  Dial 1-[PHONE-REDACTED] for the after-hours desk
  Callback [PHONE-REDACTED] (session token expired)
```

1 of 4 redacted became 4 of 4, across all three strategies, 0 cleartext.

### Test plan

- **Suppression** — parent-generated rules still apply under the new binary; the 4 newly-recovered findings generate 4 rules that suppress cleanly.
- **All 7 formats** — each carries all 4 findings.
- **Timing** — single long line stays linear, including an adversarial 4000-char unbroken prefix before the colon (the helpers walk left, so that's the shape that could compound): 250/500/1000/2000 matches at 0.089/0.097/0.133/0.155s, findings growing 1:1.
- **Non-vacuity** — each of the three fixes neutralized in turn; each put exactly its own test group red.
- `TestWordGoverningBounds` caught a real **index-out-of-range panic** on an empty line before it shipped.

Golden corpus moves no snapshot. gofmt / vet / build clean, 58 packages, cross-platform vet, 3× flake, whole-module `-race`, `-short`.

---

These came out of a validator-wide sweep for the same defect shape, which found it in **8 more validators** (email, physicaladdress, medicalid, passport, dob, personname, secrets). Those are deliberately **not** in this PR — different mechanisms needing separate measurement — and will follow individually.

rustic (approved):

PR #230, opened 2026-07-29
perf(dob): hoist the keyword scan out of the per-candidate loop

sectoramen:
## The sweep

Measured single-line scaling for **all 18 measurable validators** with distinct values at n = 250 / 500 / 1000 / 2000. `DATE_OF_BIRTH` was the only quadratic, and it was severe:

| n | before | ratio |
|---|---|---|
| 250 | 0.14s | |
| 500 | 0.30s | 2.1× |
| 1000 | 0.86s | 2.9× |
| 2000 | 3.34s | 3.9× |
| 4000 | **15.68s** | 4.7× |

`analyzeContext` recomputed every keyword tally for **each candidate date**, and each recomputation scanned the whole line against ~40 positive keywords, ~20 negative keywords and the non-human indicator list — `O(candidates × line length × keywords)`, the single-long-line CPU-exhaustion shape several other validators have already had hoisted out. `findKeywords` then did the same whole-line work twice more per candidate to fill `ContextInfo`.

## After

```
0.09 / 0.11 / 0.14 / 0.13s   at 250 / 500 / 1000 / 2000
15.68s -> 0.13s at n=4000  (88x), curve flat
```

`BeforeText` and `AfterText` are windows **inside the same line**, so the old `fullContext` (`BeforeText + line + AfterText`) contained no token the line didn't already contain. Scanning the line alone is equivalent, not an approximation — and findings are **byte-identical** before and after at n = 250 / 1000 / 2000 (line, text and confidence compared per finding). `analyzeContext` keeps its original signature and delegates.

## The rest of the sweep — all linear

Worst ratio per doubling: SSN 1.31, PHONE 1.20, EMAIL 1.31, CREDIT_CARD 1.11, IP_ADDRESS 1.20, MEDICAL_ID 1.33, DRIVERS_LICENSE 1.18, PASSPORT 1.27, BANK_ACCOUNT 1.11, VIN 1.12, PERSON_NAME flat, PHYSICAL_ADDRESS 1.20, SECRETS 1.25, OTP linear, CLOUD_RESOURCES 1.12, INTELLECTUAL_PROPERTY 1.33.

**SOCIAL_MEDIA could not be measured** — it's config-gated and reports nothing without a config file (tracked separately, see #211/#216). Saying so rather than counting it as a pass.

## Four fixtures that produced a false PASS

Each of these silently reports "linear" while measuring nothing, so they're worth recording:

- **Repeated values** — a quadratic measures linear once candidates dedup. My first DOB run used repeated dates, reported a flat 80 findings at every size, and hid the quadratic completely.
- **Only 400 distinct name combinations**, so PERSON_NAME plateaued and looked capped.
- **A Luhn generator that truncated the check digit it had just computed**, so 90% of the "cards" were invalid.
- **Adjacent PANs separated by one space**, where the pattern's own delimiter consumes the separator (documented in `creditcard/validator.go`).

PERSON_NAME also needed **real names** from a public dataset (717 distinct Titanic passenger names) because it matches against a name database — synthetic `Aana Bson`-style names match at 5%.

## Tests

- `TestDenseLineFindingsGrowWithInput` — asserts findings **grow** with input, so the timing guard cannot pass by measuring nothing.
- `TestKeywordScanIsHoistedOutOfTheCandidateLoop` — structural, not wall-clock: asserts the tally is a pure function of the line. A future change that reintroduces genuinely per-match context breaks that equality rather than just getting slow.
- `TestAnalyzeContextMatchesTheHoistedForm` — pins that both entry points agree.
- `TestDenseLineCompletesPromptly` — coarse backstop with a deliberately loose 2s threshold (quadratic was 3.3s, linear is 0.14s) so it isn't a flaky benchmark, plus its own non-vacuity floor.

**Non-vacuity:** reverting the call site to the per-candidate form puts n=2000 back to **3.23s** and the guard fails with the measured duration.

The non-vacuity floor also caught a *second* fixture bug during development — a trailing `"sample"` reaching the disqualifier path and zeroing the whole line in-process even though the CLI reported 2000 findings on the same shape.

gofmt / vet / build clean, 58 packages, golden corpus **moves no snapshot**, cross-platform vet, 3× flake, whole-module `-race`, `-short`.

### DIFF
--- a/internal/validators/dob/complexity_test.go
+++ b/internal/validators/dob/complexity_test.go
@@ -1,0 +1,162 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package dob
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/awslabs/ferret-scan/v2/internal/detector"
+)
+
+// denseDateLine builds ONE line carrying n distinct dates plus a keyword-rich
+// prefix and suffix.
+//
+// Distinct dates matter: with repeated values the candidate set collapses and a
+// quadratic implementation measures as linear. The first version of this
+// measurement used repeated dates, reported a flat 80 findings at every size, and
+// hid the quadratic completely.
+func denseDateLine(n int) string {
+	var sb strings.Builder
+	sb.WriteString("Patient DOB:")
+	base := time.Date(1950, 1, 1, 0, 0, 0, 0, time.UTC)
+	for i := 0; i < n; i++ {
+		sb.WriteByte(' ')
+		sb.WriteString(base.AddDate(0, 0, i*7).Format("01/02/2006"))
+		sb.WriteByte(',')
+	}
+	// A keyword-rich tail WITHOUT a disqualifier. An earlier version ended with
+	// "blood sample collected at intake", and although the CLI reports 2000
+	// findings on that shape, in-process the trailing "sample" reaches the
+	// disqualifier path for candidates far from the label and the whole line
+	// returned nothing -- which the non-vacuity floor below caught.
+	sb.WriteString(" recorded at intake by the attending physician")
+	return sb.String()
+}
+
+// TestDenseLineFindingsGrowWithInput is the non-vacuity floor for the timing
+// guard below. If the finding count stops tracking the input, the timing
+// assertion is measuring nothing and would pass no matter how bad the scaling is.
+func TestDenseLineFindingsGrowWithInput(t *testing.T) {
+	v := NewValidator()
+
+	var prev int
+	for _, n := range []int{250, 500, 1000} {
+		matches, err := v.ValidateContent(denseDateLine(n), "chart.txt")
+		if err != nil {
+			t.Fatalf("ValidateContent at n=%d: %v", n, err)
+		}
+		if len(matches) == 0 {
+			t.Fatalf("zero findings at n=%d: the fixture does not exercise the scan", n)
+		}
+		if len(matches) <= prev {
+			t.Errorf("findings did not grow with input at n=%d: got %d, previous %d",
+				n, len(matches), prev)
+		}
+		prev = len(matches)
+	}
+}
+
+// TestKeywordScanIsHoistedOutOfTheCandidateLoop is the structural guard, and it
+// is deliberately NOT a wall-clock assertion.
+//
+// The defect was that analyzeContext recomputed every keyword tally for each
+// candidate date, and each recomputation scanned the whole line against ~40
+// keywords, giving O(candidates x line length x keywords). Measured on a single
+// line of distinct dates before the hoist: 250 -> 0.14s, 500 -> 0.30s,
+// 1000 -> 0.86s, 2000 -> 3.34s, 4000 -> 15.68s (~4x per doubling). After:
+// 0.09 / 0.11 / 0.14 / 0.13s.
+//
+// This test asserts the invariant that makes it linear instead of timing it: the
+// tallies are a pure function of the line, so computing them once per line and
+// once per candidate must produce identical results. A future change that
+// reintroduces per-candidate context (something that varies per match) breaks
+// this equality, which is the actual regression to catch.
+func TestKeywordScanIsHoistedOutOfTheCandidateLoop(t *testing.T) {
+	v := NewValidator()
+
+	lines := []string{
+		"Patient DOB: 03/14/1987, blood sample collected at intake",
+		"date of birth 03/14/1987 and 07/22/1979 on file",
+		"born 03/14/1987, server age 5 years",
+		"Test DOB: 01/01/2000",
+		denseDateLine(50),
+	}
+
+	for _, line := range lines {
+		lower := strings.ToLower(line)
+		hoisted := v.newDOBLineKeywords(lower)
+
+		// Recomputing must be identical -- that is what makes the hoist sound.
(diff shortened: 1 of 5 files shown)


sectoramen:
## Follow-up: the OTHER quadratic, and why it needs no code here

The sweep above covered validators. I also measured the **file/preprocessing path**, because a validator-only sweep can't see costs upstream of it. Real `.xlsx` files (generated spreadsheets, 2 populated columns per row):

```
rows      500     1000    2000    4000
secs      0.13    0.23    0.55    1.63     <- 1.8x / 2.4x / 3.0x per doubling
```

Quadratic, repeatable across 3 runs each. But it is **not** in the validators, and not where I first assumed:

- **Extraction is linear.** Benchmarked `ExtractText` directly: 4.3 / 7.7 / 16.1 / 32.9 ms — a clean 2× per doubling.
- **The router is linear**, and negligible: `RouteContent` on the same content is 0.35 / 0.58 / 1.16 / 2.37 **ms** — under 0.2% of the 1.63s.
- Identical content as plain `.txt` scans in **0.41s vs 1.65s** as `.xlsx`, with byte-identical findings (7995 either way), and the penalty is the same whether you enable one check or all — so it's a fixed per-document cost, not per-validator work.

A CPU profile named it: **`preprocessors.splitLines`, 1.45s flat**, the only application function in the hot list. Its caller is the problem, not the function:

```
CalculateAbsoluteOffset (inline)  ---> splitLines   1.93s cum
```

`CalculateAbsoluteOffset` re-splits the **entire document** on every call, and `createBasicLineMappings` calls it **twice per line**.

### That is already fixed by #218

Same function, same call sites, same mechanism — and #218's commit message carries the same measurements I reproduced independently before finding it. Verified by building both:

| rows | this branch | #218 | |
|---|---|---|---|
| 500 | 0.16 | 0.14 | |
| 1000 | 0.24 | 0.14 | 1.7× |
| 2000 | 0.55 | 0.21 | 2.6× |
| 4000 | 1.52 | 0.32 | **4.8×** |

#218's curve is 0.14 / 0.21 / 0.32 — linear.

### The two are independent

With #218 alone, the DOB quadratic in this PR is still there and still severe:

```
n                1000     2000     4000
#218 only        1.62s    3.68s    12.59s     <- still quadratic
+ this PR        0.09s    0.11s    0.15s      <- 84x at n=4000
```

So: **two distinct quadratics, both covered** — the preprocessing one by #218, the DOB keyword scan by this PR. No additional code needed here, and no third quadratic found anywhere in the scan path.

Worth noting I was wrong twice while chasing this before profiling: I read `--preprocess-only` timings as "extraction is quadratic" (it includes more than extraction), and I suspected a `documentLines = append([]string{line}, ...)` prepend in the router (real O(n²) shape, but the router is 2.4ms — irrelevant at this scale). The profile settled it; code reading didn't.

sectoramen:
Closing: **the DOB fix here is redundant with #212**, which I found only after building the union.

#212 already hoists `analyzeContext` out of the per-candidate loop — and does it better, since it also renames `containsKeyword` → `containsKeywordLower` and removes the per-match `ContextInfo` that was fabricating keywords. Measured on my own quadratic fixture, built from #212's branch:

```
n            1000     2000     4000
#212         0.84s*   0.13s    0.17s     <- already linear (*first-run warmup)
```

So the quadratic this PR set out to fix is fixed on a branch that has been open the whole time. Shipping mine would have meant two competing implementations of the same hoist.

## What the sweep still established

The measurement work stands and is worth keeping on the record:

- **All 18 measurable validators swept** for single-line quadratic scaling with distinct values. Worst ratio per doubling: SSN 1.31, PHONE 1.20, EMAIL 1.31, CREDIT_CARD 1.11, IP_ADDRESS 1.20, MEDICAL_ID 1.33, DRIVERS_LICENSE 1.18, PASSPORT 1.27, BANK_ACCOUNT 1.11, VIN 1.12, PERSON_NAME flat, PHYSICAL_ADDRESS 1.20, SECRETS 1.25, OTP linear, CLOUD_RESOURCES 1.12, INTELLECTUAL_PROPERTY 1.33. **DATE_OF_BIRTH was the only quadratic** — and #212 owns it.
- **SOCIAL_MEDIA remains unmeasurable** (config-gated, reports nothing without a config file). Stating that rather than counting it as a pass.
- **A second, independent quadratic in the preprocessing path**, in `.xlsx` scans: `CalculateAbsoluteOffset` re-splits the whole document and `createBasicLineMappings` calls it twice per line. A CPU profile named `preprocessors.splitLines` at 1.45s flat. **Also already fixed — by #218**, verified 1.52s → 0.32s at 4000 rows. Details in the comment above.
- **Four fixture shapes that produce a false PASS** (repeated values, too few distinct name combinations, a Luhn generator truncating its own check digit, adjacent PANs whose separator the pattern consumes). Recorded in the comment above so the next sweep doesn't repeat them.

Net: **no quadratic in the scan path is unaddressed**, across validators, extraction, routing and preprocessing — but the two that exist are covered by #212 and #218, not by new code.

## Where the tests went

The DOB positional-disqualifier work (a separate leak, which #212 does *not* fix — verified: its leak cases still report 0 on #212 alone) moved to **#229**, reformulated on top of #212 to read the line at the match's own offset instead of an `AfterText` window. That matters: #212 documents a `"test"` sliced out of `"contest"` deleting a real DOB, and my original version read exactly that window.

The complexity tests from this branch aren't carried over — #212 and #218 each ship their own timing guards for the code they change, and duplicating them against an implementation that isn't landing would be dead weight.

PR #231, opened 2026-07-30
feat(passport): detect MRZ line 2 and verify its ICAO check digits

sectoramen:
> **Stacked.** Built on `work/all-through-229` (all of #207–#229). It touches the same emit loop and `determineCountry` region as #220 and #224, so it does not cherry-pick onto bare `main`. Review after those land; the diff below is MRZ-only.

## Problem

TD3 line 1 carries the holder's **name**. Line 2 carries the passport **number**, date of birth, sex, expiry and personal number — and **nothing matched it at all**.

That's a cleartext leak on the most ordinary passport shape there is. A scan, an OCR pipeline or any ICAO-compliant record emits *both* lines, and redaction covered only the first:

```
before:  [PASSPORT-REDACTED]                            <- line 1 redacted
         L898902C36GBR7408122M1204159ZE184226B<<<<<10   <- line 2 CLEARTEXT

after:   [PASSPORT-REDACTED]
         [PASSPORT-REDACTED]
```

0 of 4 values (document number, DOB, expiry, personal number) survive in cleartext, across all three redaction strategies.

## What makes a 44-character pattern safe

The shape gate is `[A-Z0-9<]{44}` — which any base32 secret or hash of that length also satisfies. On its own it would be a false-positive machine.

Every candidate must then pass `mrzLine2ChecksValid`: **five independent ICAO 9303 7-3-1 check digits** (document number, DOB, expiry, personal number, and the composite over all of them), plus range checks on the date fields and the sex field.

That's a **value-intrinsic** test, which is the whole point. Every suppression defect fixed in #226–#229 shared one root cause: a keyword *near* the value decided whether a real identifier was reported, so anyone who could influence a byte of the document could delete a finding. A check digit cannot be talked out of.

## Precision — measured, not asserted

| probe | result |
|---|---|
| 20,000 random 44-char MRZ-alphabet tokens | **0 reported** |
| doubled AWS key id, base32, sequential digits, all-A, all-filler | **0 reported** |
| the repo's own source tree (~2000 files, recursive) | **114 → 114**, zero delta |
| each of the five check digits corrupted in turn | finding correctly dropped |

## Three call sites, each found by measurement rather than reading

- **emit loop** — rejects a line-2 candidate whose checks fail.
- **`determineCountry`** — gains an `MRZ_TD3_L2` arm. Line 2 has no `P` prefix so it fell through to `""`, and the caller skips a match whose country is `""` — so every line-2 finding was **silently dropped after passing the checks**.
- **`hasStrongPassportContextWith`** — gains the bypass line 1 already has. Line 2 is pure machine-readable data with no `passport` keyword, so the prose-context requirement dropped it even with all five digits valid. Line 1 earns its bypass *structurally*; line 2 earns it *arithmetically*.
- **`CalculateConfidence`** — gains an arm: +25, and reads the issuing state from `[10:13]` (line 2's position) instead of the 2-letter lookup used elsewhere, which was reporting a valid `GBR` as an invalid country code and docking the score below the surfacing threshold.

## Suppression is compatible

A parent-generated rule for the line-1 finding **still suppresses line 1** under the new binary, and line 1's confidence is unchanged at **80.00** — so no existing suppression file is invalidated. The line-2 finding survives that rule set only because the parent could not generate a rule for a finding it never reported.

Contrast the passport *proximity* fix I held back from #229 for exactly this reason: it **raised** confidence on already-reported findings (70 → 90) and measurably broke stale suppression files. That one still waits on the hash-version migration.

## Performance

The pattern runs on every line, so the worst case is dense uppercase text:

```
lines     2000    4000    8000   16000
before    0.14    0.17    0.28    0.50
after     0.13    0.18    0.30    0.52
```

Within noise, linear.

## Test plan

- **Non-vacuity** — removing the pattern, disabling the check-digit gate, and removing the context bypass each put a **distinct** test group red.
- **Golden case** pairs the positive with the *same* line 2 carrying one corrupted check digit, which must not be reported. A snapshot of the positive alone would still pass if check-digit validation were deleted. The redaction snapshot is the gate that catches the leak returning.
- `TestMRZCheckDigit` pins the 7-3-1 arithmetic against ICAO 9303 specimen values, so a refactor can't quietly change the weighting.
- Invalid characters return `-1` rather than scoring as a filler — otherwise a lowercase byte could help a non-MRZ token satisfy the arithmetic.

gofmt / vet / build clean, 59 packages, golden regen moves nothing beyond the new case, cross-platform vet, 3× flake, whole-module `-race`, `-short`.

---

One process note worth recording: the first fixture I built by hand had a wrong personal-number check digit, which made *correct* production code look broken and cost a detour. The test helper now computes all five digits from the fields (`buildTD3Line2`) so a fixture cannot lie.


sectoramen:
## CI: diagnosed and fixed (`d3f8e67`)

All three OS jobs failed on one subtest, `TestValidatorComplexityIsSubQuadratic/socialmedia`. **It is not the MRZ change, and not a quadratic regression** — it is the timing guard's absolute ceiling colliding with `-race`.

```
socialmedia, 4x input, identical tree:
  without -race    0.33s
  with -race       4.97s      <- 15x instrumentation, 30ms UNDER the 5s ceiling
  CI runner       12.61s      <- same code, slower hardware, over the ceiling
```

CI runs `go test -race`, which instruments every memory access (10–20× on scan-heavy code), and the runner is a further 2–3× slower than a developer machine. The guard was passing locally by **thirty milliseconds** — this failure was waiting for any slower host.

### Fix

The ceiling is multiplied by 8 when `-race` is active, detected with the standard build-tag pair (Go has no runtime API for it). Normal runs are unchanged — multiplier 1, every declared threshold exactly as written.

The **ratio** check is what actually detects quadratic scaling, and it is unaffected: `-race` inflates the base and the 4× measurement equally, so the growth factor is preserved. The absolute ceiling stays as a backstop against a pathological blow-up, which 8× still catches — a genuine quadratic is 16× per 4× of input.

Verified still non-vacuous: forcing the ceiling to 1ns makes **all 18** subtests report it. And the exact CI command now passes locally — 59 packages, 0 failures, 0 data races.

### Why this surfaced here and nowhere else

The *configured* (non-vacuous) socialmedia target comes from #213. #226–#229 are single-commit branches that don't contain it, so their CI measures a 42ns early return instead of a real scan. **#231 is stacked on all 23 open PRs, so it's the only branch where that target does real work under `-race`.**

A property of the stack, not of this change. Worth noting I started from the assumption my own commit was at fault and let the measurement decide; two other hypotheses were checked and **refuted** on the way:

- that #216's handle-regex fix changed the match count — it does not: 66ms/243ms with it, 70ms/239ms without
- that the inputs straddled `maxClusterMatches` — they do not; both sizes stay above it

I also briefly mis-measured this as "passes locally, 1 match" because my probe called the generator once instead of per-repetition. The corrected measurement is 1200→4800 matches at 67ms→249ms — clean 3.7× for 4× input.

sectoramen:
## Correction: my first fix was incomplete (`91e4e89` fixes it properly)

CI came back still failing, and reading the message carefully showed why — it said `> 5s ceiling` **without** the `, scaled 8x for -race` suffix my change adds. So the multiplier wasn't being applied at all.

Cause was mine: `d3f8e67` added the constants (`raceDetectorEnabled` / `raceCeilingMultiplier` / `raceNote`) but **not the code that uses them**. I had reverted the guard file with `git checkout --` to undo a temporary 1ns-ceiling mutation from a non-vacuity check, and that discarded the real edit in the same file. Only the two new build-tagged files made it into the commit — a dead constant pair.

The diagnosis in my previous comment was right; the patch was half-applied.

### Now measured properly

On the same tree, with the threshold temporarily lowered to 1s to stand in for a slower host:

```
multiplier 8 (shipped)   socialmedia passes
multiplier 1 (mutated)   socialmedia FAILS: 3.56s > 1s ceiling, scaled 8x for -race
```

That substitution was necessary: at the real 5s threshold the *mutated* form still passes locally at ~4.97s. Which is precisely why CI caught this and my machine didn't — I was passing by 30 milliseconds. The error message now carries the multiplier, so a future failure states whether the ceiling was scaled.

Whole-module `go test -race -count=1` (the exact CI command) passes locally: 59 packages, 0 failures, 0 data races.

rustic (approved):

PR #239, opened 2026-07-31
fix(router): take document structure from the extractor, not from the document

sectoramen:
**Stacked on #238.** The structural fix — removes the attacker's capability rather than containing its consequence. Verified **SOUND** by an independent adversarial pass.

## The mechanism

`file_router.go` flattened every extractor's output into one `ProcessedContent.Text` using literal `\n\n--- name ---\n` separators, and `content_router.go` then **re-parsed that text** to recover the structure. Document authors control the text, so they controlled the "structure". This is in-band signaling — the same class as SQL injection — and the answer is the same: carry structure out of band.

**Producer:** `Sections []ContentSection` (plus `SectionKind`, `ClassifySection`) on `ProcessedContent`, appended in the same loop that already builds the flat concatenation, so the description cannot drift from what it describes. `Kind`/`Type` come from `Preprocessor.GetName()` — our own constant, never a document byte. `Text` stays the byte-identical flat concatenation, so the field is purely additive.

**Consumer:** `RouteContent` reads only `ContentSection` fields via a new `routeSections`, never the section text. With no declared structure it **fails closed** — the whole extraction becomes `DocumentBody`, so the document path runs all validators and the worst case is a lost metadata *label*, never lost coverage.

**Deletions:** `content_router.go` goes **1129 → 349 lines**. 24 symbols removed, including `separateCombinedPreprocessorOutput`, `splitByPreprocessorSeparators`, `detectMetadataSection`, `isMetadataSectionEnd`, `containsMetadataSections`, `looksLikeMetadata`, `extractEmbeddedMediaPath`, the four `contains*MetadataPatterns` sniffers, the `preprocessorSection` type, and `PreprocessorTypeMixedContent` (a category only the deleted classifier returned).

## Two things not in the original plan

**Embedded media is a sub-section inside one preprocessor's text**, and it routes to a *different* rule set — a `.wav` in a `.docx` has `Artist:`, on the audio sensitive-field list but not the office one. The first cut regressed `embedded.docx` from 7 findings to 6, losing `AUTHOR_INFO` for the clip's artist address. Fixed by having `ProcessEmbeddedMedia` return sections with its text and the router re-anchor rather than overwrite them. Caught by measurement, not by a test written to pass.

**`identifyPreprocessorType` also had to go**, though it wasn't on the list. It survived only as a timing label — but it sniffs document text, so reporting an attacker-derived label would have kept the in-band read path alive. Replaced with `processedContent.ProcessorType` in the observability payload.

**Docs:** both content-routing docs instructed maintainers to fix routing by regex-matching `--- image_metadata ---` against section text — i.e. to reintroduce this exact vulnerability — against an API (`ContentSection.Header`) that never existed. Rewritten to describe the declared-section mechanism.

## Testing

**Golden: ZERO change.** `git diff sec/pr2-extraction-part-names -- internal/goldencorpus/testdata/golden/` is empty. No regeneration, `UPDATE_GOLDEN` never used. That is the load-bearing result: the mechanism for determining structure was replaced wholesale and nothing about what gets detected moved.

Beyond the corpus, base-vs-fix equivalence on **14 fixtures × 7 formats**, identical modulo timestamps — with one disclosed exception: `forged_embed.docx` differs by one *removed* forged-attribution false positive, which is an improvement. The second commit here exists specifically to disclose that rather than let it hide inside "identical".

**Non-vacuity, isolated properly.** Reverting all production files gives only a compile error, which proves nothing about routing. So the type and producer were restored and **only the consumer** reverted:

```
--- FAIL: TestDeclaredSectionsDecideRouting_NotTheText
    a forged section header in the BODY changed the metadata split:
    2 items with the header vs 1 without.
--- FAIL: TestNoDeclaredSections_FailsClosedToDocumentBody
    DocumentBody is not the whole extraction.
```

Separately, replacing `Text: pResult.result.Text` with a copy fails `TestSectionTextAliasesNotCopies` — *"every scan now holds a second copy of the whole extracted document"*.

Note **5 of 7 new tests pass on a consumer-only revert**, because PR1's union already prevents the deletion. That is defence-in-depth working, not vacuity, and the test file documents it in a header comment rather than presenting seven independent proofs.

**PR1's union is intact and deliberately kept**, with a comment warning against "simplifying" it away now that sections are trustworthy — removing the capability and containing the consequence are different jobs. A forged first-line header still yields the SSN.

**Adversarial verify: no bypass survives.** Forged headers at first-line / mid-body / interleaved, odd casing and internal whitespace, the legacy `=== METADATA ===` form, all five metadata names forged at once, an `.xlsx` member named literally `xl/worksheets/--- office_metadata ---.xml`, and an embedded `.wav` named `a --- office_metadata --- b.wav` — none change section structure.

61 packages pass · `-race` clean on all five touched packages · cross-platform vet clean · timing flat within noise (0.96–1.02× over 12 points, linear on all three fixture families).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build clean vet fmt run install-config install check-go-version pr-checklist integration-branch
+.PHONY: build clean vet vet-host fmt run install-config install check-go-version pr-checklist integration-branch test-compile-all-platforms
 
 # Default target
 all: check-go-version fmt vet build
@@ -36,6 +36,7 @@ help:
 	@echo ""
 	@echo "🧪 Testing:"
 	@echo "  test               - Run all tests"
+	@echo "  test-compile-all-platforms - Cross-compile test binaries for every platform"
 	@echo "  test-ci            - Run CI-equivalent tests (race detector, no cache)"
 	@echo "  test-windows       - Run Windows-specific tests"
 	@echo "  test-cross-platform - Run cross-platform compatibility tests"
@@ -55,7 +56,8 @@ help:
 	@echo "  uninstall          - Remove system-wide installation"
 	@echo "  fmt                - Format code"
 
-	@echo "  vet                - Run go vet"
+	@echo "  vet                - Run go vet for darwin, linux and windows (compile-only)"
+	@echo "  vet-host           - Run go vet for this platform only (fast inner loop)"
 	@echo ""
 	@echo "🐳 Container:"
 	@echo "  container-run      - Run container"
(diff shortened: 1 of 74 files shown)


rustic:
Superseded by #244. This PR was auto-closed by GitHub when its base branch `sec/pr0-office-golden-fixtures` (merged into main via #243) was deleted. #244 carries the identical commits, rebased onto `main`.

PR #240, opened 2026-07-31
fix(metadata): take a finding's reported source from the package, not from the text

sectoramen:
**Stacked on #239.** Last of the sequence. Closes the attribution half that no earlier PR reaches.

## The defect

`extractEmbeddedMediaPath` parsed whatever sat between the first `(` and the first `)` of an `--- Embedded Media N (name) ---` line and built `MetadataContent.SourceFile` from it. That becomes `Match.Filename`. A document author can type that line as body text or supply the media part name, so **the reported source of a finding was attacker-chosen**:

| typed in a Word paragraph | reported as |
|---|---|
| `--- Embedded Media 1 (/etc/passwd) ---` | `report.docx -> passwd` |
| `--- Embedded Media 1 (../../secrets.txt) ---` | `report.docx -> secrets.txt` |
| `--- Embedded Media 1 (totally-different.docx) ---` | `report.docx -> totally-different.docx` |

`Match.Filename` feeds the suppression hash (`generateFindingHash` folds `filepath.Base(Filename)`), the SARIF `uri`, the gitlab-sast `file`, the JUnit name, and `internal/explain`'s `looksLikeTestPath` — where a name like `evil_test.go` flips the verdict toward "likely test data", i.e. advice to ignore a real finding.

`filepath.Base` neutralised traversal in the *displayed* value; it did nothing about the value being attacker-chosen at all.

Provenance now travels out of band in `ContentSection.SourceFile`, set by the preprocessor that actually opened the archive member. Both parse sites are gone — `extractEmbeddedMediaPath` (with #239's read side) and the METADATA validator's line loop. The `--- Embedded Media N (name) ---` line is now **display only**, with a comment saying so and saying not to reintroduce a parser or "fix" a wrong attribution by escaping the parens: an author still controls whatever a parser would read.

## What the adversarial pass caught

The verify agent returned **DEFECTIVE** on the first cut, with two regressions confined to the bridge's legacy fallback (`processContentLegacy`; `EnableFallbackMode` defaults **true**, and the arm runs whenever routing or dual-path processing fails). The final commit here fixes both.

That path hands the whole concatenation to the metadata validator with only the container path and has no `ContentRouter` to consult — so the emitted header was its **only** provenance carrier. Deleting the reader while leaving the emitter in place silently re-attributed real embedded findings to the container:

| binary | `real_embed.docx` (embedded `.wav`) |
|---|---|
| parent | `AUDIO_ARTIST_IDENTITY`, `EMAIL`, `IMAGE_AUTHOR` → `real_embed.docx -> audio1.wav` |
| first cut | all three → the container path ❌ |
| final | all three → `real_embed.docx -> audio1.wav` ✅ |

And because the suppression hash folds the filename, rules generated with the **parent** binary (11 enabled) suppressed only **8**; three legitimate findings escaped. Now **11/11**.

The fix validates each *declared section* separately, so each finding inherits the `SourceFile` its preprocessor recorded. Sections are produced before any routing decision, so they are available even though this arm only runs when routing failed. With no declared sections it falls closed to the previous behaviour — whole text, container path — because losing a precise label is acceptable and losing the scan is not.

**Fixing attribution alone was not enough**, and the second half is easy to miss. A validator handed one section counts lines from 1 inside it, so per-section validation made reported lines section-relative (21 → 6, 8 → 1, 9 → 2). The suppression hash embeds the line number, so that traded one hash invalidation for another — suppression went **8/11 → 6/11** before rebasing by `ContentSection.LineOffset`. Line numbers are now byte-identical to the parent.

## Testing

**Non-vacuity.** The three forgery tests: 10/10 subcases plus the suppression and GPS tests fail on the parent with the exact quoted output, and both new tests carry explicit floors (fatal if no finding lands after the forged header; fatal if no GPS pair is found). The legacy fix is mutation-proven in **both** halves separately — forcing `lineOffset` to 0 fails on the line-number assertion; replacing `section.SourceFile` with the container path fails on the attribution assertion.

**Forgery closed on every reachable path.** 15 shapes across the direct validator, the adapter fallback, and all 8 preprocessor arms all report the caller's path.

**Honest residual, stated rather than claimed away:** the replacement provenance is a zip member name, which the same document author also writes. So the specific harms are *narrowed*, not eliminated — a member at `word/media/evil_test.go` still reaches `looksLikeTestPath`. What is gone is forging attribution from **body text**, which needed no control of the archive at all.

61 packages pass · gofmt clean · vet clean for darwin/linux/windows · goldens unchanged vs #239 · SSN/PAN removal identical base-vs-fix on all 5 containers.

### DIFF
--- a/docs/development/content-routing-troubleshooting.md
+++ b/docs/development/content-routing-troubleshooting.md
@@ -141,35 +141,27 @@ ferret-scan --file document.pdf --debug-content-routing --dry-run
 
 **Solutions:**
 
-1. **Update Preprocessor Configuration:**
-```yaml
-# Ensure preprocessors include proper section markers
-preprocessors:
-  metadata_extractor:
-    include_section_markers: true
-    marker_format: "--- {type}_metadata ---"
-```
+> **Do not fix this by matching section markers in the extracted text.** Section
+> boundaries are *declared* by the producing preprocessor in
+> `ProcessedContent.Sections`, not recovered from the text. Matching
+> `--- image_metadata ---` in content lets a document author choose the routing,
+> because the author writes the text — and because the metadata path runs one
+> field-name scanner rather than the full validator set, a forged marker deletes
+> findings instead of relabelling them. A finding that is never reported is never
+> redacted.
 
-2. **Fix Section Pattern Matching:**
-```go
-// Update pattern matching logic if needed
-func (cr *ContentRouter) identifyPreprocessorType(section ContentSection) string {
-    patterns := map[string]string{
-        "image_metadata":    `(?i)---\s*image_metadata\s*---`,
-        "document_metadata": `(?i)---\s*document_metadata\s*---`,
-        "audio_metadata":    `(?i)---\s*audio_metadata\s*---`,
-        "video_metadata":    `(?i)---\s*video_metadata\s*---`,
-    }
+1. **Check that the preprocessor populates `Sections`:**
+   A preprocessor that returns text but no sections is treated as *all document
+   body* (fail-closed). Coverage is intact, but metadata labelling is lost, which
+   is what "missing preprocessor type information" looks like. The fix belongs in
+   the preprocessor: return one `ContentSection` per distinct run of its output,
+   as `office_metadata` does for its property block and each embedded media item.
 
-    for preprocessorType, pattern := range patterns {
-        if matched, _ := regexp.MatchString(pattern, section.Header); matched {
-            return preprocessorType
-        }
-    }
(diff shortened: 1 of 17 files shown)


rustic:
Superseded by #245, rebased onto `main`. GitHub blocks retargeting a stacked PR's base, and this PR's old base branch predates the #244 rebase-merge into `main`.

rustic:
All commits landed on `main` via #245 (rebase-merged). Closing this superseded stacked PR.

PR #241, opened 2026-07-31
fix(secrets): stop document context promoting a bare hex identifier to a secret

sectoramen:
**Stacked on #240.** Fixes the false positive surfaced (not caused) by #237, and adds a small generic mechanism other validators can reuse.

## The finding

An EXIF `ImageUniqueID` reported as `API_KEY_OR_SECRET` at **80 (MEDIUM)** in a real 10 MB `.docx`:

```
ImageUniqueID: "14f6364997257b9170c016a13d1f1127"
```

That's an image GUID. It reaches the generic entropy path only because `containsSecretIndicators` admits *any* quoted value of 20+ bytes — no secret keyword is involved. Digests, ETags, commit ids and content hashes share the shape.

## Why the obvious fix doesn't work

The secrets validator scores this value **55**. The 80 is:

```
55 (validator) + 20 (document-context adjustment) + 5 (cross-path correlation boost)
```

Both applied **after** the validator returns, at three separate sites in the bridge. So a validator clamping its own output has no way to make a bound stick — I tried that first and the reported score stayed at 80.

## The mechanism

A small generic contract instead: a validator sets `Match.Metadata["confidence_ceiling"]`, and the bridge clamps to it after **every** adjustment that can raise confidence.

The key is declared in the bridge rather than imported from a validator package, so the dependency keeps pointing one way. A test fails if the two literals drift, since a silent mismatch would turn every ceiling into a no-op.

Any validator with a value-intrinsic reason to say *"no amount of surrounding context should promote this"* can now use it without new plumbing. The secrets validator declares a ceiling of 55 for a value that is bare hex at an identifier length (32/40/64) on a line carrying no secret-ish word.

## Two decisions worth reviewing

**It is a ceiling, not a veto.** A 32-hex string genuinely can be a secret, and only reported findings reach the redactor — dropping one would leave the value in the "redacted" output, which is the failure mode this whole area keeps producing. Verified: the demoted finding is still redacted, output reads `ImageUniqueID: "[API_KEY_OR_SECRET-REDACTED]"` with zero cleartext occurrences.

**The discriminator is the line's label as a POSITIVE signal** — the ceiling applies when nothing says "secret" — rather than a negative list of identifier field names like `ImageUniqueID` or `SerialNumber`. That direction is the point. A negative list would let a document author suppress a **real** secret by writing `ImageUniqueID: "AKIA..."`: attacker-controlled suppression, the same class of defect #237–#240 have been removing. Here relabelling can only *remove* the ceiling.

Measured on one identical value:

| line | confidence |
|---|---|
| `api_key = "14f6364997257b9170c016a13d1f1127"` | **75** (untouched) |
| `ImageUniqueID: "14f6364997257b9170c016a13d1f1127"` | **55** (capped) |

## Result

| | before | after |
|---|---|---|
| the reported row | `[MEDIUM] 80.00%` | `[LOW] 55.00%` |
| total findings | 78 | **78** — count deliberately unchanged |

## Testing

Per `docs/testing/TEST_PLAN.md`.

**D10 non-vacuity, both halves separately:** removing the clamp calls restores 80 on the real document; dropping the label check fails **8 of 20** shape subtests.

**D3** goldens 0 changed — the corpus has no `API_KEY_OR_SECRET` row at all, so this moves nothing. **D4** linear on all four fixture families (1.70–1.88× per doubling), within 1.03× of base. **D5** sink verified above. **D11** 10 runs stable at 78 findings / 55 confidence. **D1/D2/D12/D13** 61 packages pass, gofmt clean, vet clean *and* test binaries compile for darwin/linux/windows, `-race` clean on both touched packages.

## One accepted cost

The suppression hash embeds `%.2f` confidence, so demoting a finding invalidates a rule written against it. Of 78 rules generated by the parent binary on that document, **77 still match and 1 escapes** — the demoted row, and nothing else. Any confidence change has this property; this is the minimum blast radius for it.

## Not addressed here

The **+5 cross-path correlation boost** applies to *every* finding whenever both validation paths hit — so adding one finding to a document can shift an unrelated finding's confidence and invalidate its suppression rule. That touches every finding in every container and doesn't belong in a one-row fix. Its own issue.

### DIFF
--- a/internal/preprocessors/shared_utilities.go
+++ b/internal/preprocessors/shared_utilities.go
@@ -270,7 +270,21 @@ func (rih *RouterIntegrationHelper) CreateEmbeddedMediaPath(originalFilePath, em
 	return fmt.Sprintf("%s -> %s", filepath.Base(originalFilePath), embeddedFileName)
 }
 
-// FormatEmbeddedMediaSection formats embedded media content for inclusion in metadata text
+// FormatEmbeddedMediaSection formats embedded media content for inclusion in
+// metadata text.
+//
+// The "--- Embedded Media N (name) ---" line is now DISPLAY ONLY: it tells a
+// human reading the extracted text where an embedded item begins, and nothing
+// parses it back. Two readers used to, and both let a document author choose the
+// reported source of a finding just by typing this line as body text — the
+// content router's extractEmbeddedMediaPath (deleted with the rest of the
+// text-sniffing read side) and the METADATA validator's line loop. The real
+// provenance travels out of band in ContentSection.SourceFile, set by the
+// preprocessor that actually opened the archive member.
+//
+// So do not reintroduce a parser for this text, and do not "fix" a wrong
+// attribution by escaping the parentheses or the name: an author still controls
+// whatever a parser would read, and the structure carries the answer already.
 func (rih *RouterIntegrationHelper) FormatEmbeddedMediaSection(index int, mediaName, content string) string {
 	var result strings.Builder
 	result.WriteString(fmt.Sprintf("\n--- Embedded Media %d (%s) ---\n", index+1, mediaName))
--- a/internal/router/attribution_provenance_test.go
+++ b/internal/router/attribution_provenance_test.go
@@ -1,0 +1,252 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package router
+
+import (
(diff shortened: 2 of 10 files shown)


rustic:
Superseded by #246, rebased onto `main`. GitHub blocks retargeting a stacked PR's base, and this PR's old base branch predates the rest of the stack landing on `main`.

rustic:
All commits landed on `main` via #246 (rebase-merged). Closing this superseded stacked PR.

PR #248, opened 2026-07-31
fix(validators): stop a finding's confidence depending on what else is in the file

sectoramen:
**2 of 3 confidence-integrity fixes.** Based on #247 — merge that first, then this. GitHub will retarget this to `main` automatically once #247 lands.

## The defect

The bridge added a flat **+5 to every finding** whenever both validation paths reported something, on the theory that a document with both body and metadata findings is more likely to hold real data. The consequence: a finding's confidence described the **file**, not the finding. Scanning the same value in two documents gave two answers.

Two `.docx` files with identical content and identical basenames, differing only by an unrelated author name in the metadata:

| fixture | `API_KEY_OR_SECRET` |
|---|---|
| no metadata | **55** |
| unrelated metadata | **60** |

On a real 10 MB `.docx`, **all 78** findings carried the boost and **14** had their score genuinely shifted by it (the rest were already clamped at 100), across 9 types: `PERSON_NAME`, `IMAGE_SOFTWARE`, `AUTHOR_INFO`, `LAST_MODIFIED_BY`, `COMPANY_INFO`, `APPLICATION_INFO`, `DOCUMENT_COMMENTS`, `API_KEY_OR_SECRET`, `TEMPLATE_INFO`.

## Why it's wrong

A confidence answers *"how sure are we about **this** value"* — so the same secret in two documents has to score the same.

And while the suppression hash folded confidence, adding one unrelated finding to a file silently invalidated the saved rules of **every other finding in it**. An operator's reviewed-and-accepted finding came back as noise, and in a pre-commit gate as a block. #247 removes confidence from that hash, which defuses the second consequence — but the scores themselves were still arbitrary.

## What else went

Removing the boost left `applyCrossPathConfidenceAdjustments` grouping matches by path and then returning them unchanged, so the function is **deleted** rather than kept as an identity call a reader has to verify does nothing (TEST_PLAN D10: a fix that makes a branch unreachable must remove it).

Per-path context adjustments still happen inside each bridge, where the match's own validator is known. The confidence-ceiling clamp from #241 survives at both remaining raise sites; the one inside the deleted function went with the raise it guarded.

## Cross-finding correlation is still worth having

Just not like this. A MEDIUM name next to a HIGH SSN probably *is* a person. Doing it properly needs **proximity** (this applied document-wide), a **stated per-type policy** (this hit every type equally, including metadata types whose confidence means something different), and an **opt-in flag with a benchmark**. That's the co-occurrence reranker — a separate change, and one this unblocks: you can't evaluate a co-occurrence reranker while an undocumented co-occurrence boost is already firing.

## Golden impact

**7 cases move, all dual-path containers; no text-only case moves** — the expected shape, since the boost only fired when both paths reported.

Diffs are confidence-only at exactly **−5** per finding, counts unchanged:

| case | before → after |
|---|---|
| `file_docx_body_and_metadata` | 4 → 4 findings |
| `file_wav_metadata_pii` | 5 → 5 findings |
| `file_xlsx_sheet_cells` | 3 → 3 findings |

The extra removed lines are the `cross_path_correlation`, `correlation_boost` and `original_confidence` metadata keys the boost used to attach.

## Testing

Per `docs/testing/TEST_PLAN.md`.

Tests drive `core.ScanFile` with real `.docx` bytes rather than the bridge directly, because the defect lived in *how the two paths were combined* — a hand-wired bridge would test the test's own wiring.

**D10 non-vacuity:** restoring the boost fails both tests. Each carries a floor that fails if its fixture produces no finding at all, so neither can pass by finding nothing. Coverage is symmetric: a body finding must not move because metadata exists, **and** a metadata finding must not move because the body has one.

61 packages pass · gofmt clean · vet clean for darwin/linux/windows · `-race` clean on validators and core.

### DIFF
--- a/internal/core/confidence_independence_test.go
+++ b/internal/core/confidence_independence_test.go
@@ -1,0 +1,168 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package core_test
+
+import (
+	"archive/zip"
+	"bytes"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/awslabs/ferret-scan/v2/internal/core"
+)
+
+// A finding's confidence must not depend on what else is in the file.
+//
+// The bridge used to add a flat +5 to EVERY finding whenever both validation paths
+// reported something. So scanning the same value in two documents gave two answers.
+// Measured on two .docx files with identical content and identical basenames, differing
+// only by an unrelated author name in the metadata, the same API_KEY_OR_SECRET scored 55
+// in one and 60 in the other. On a real 10 MB .docx all 78 findings carried the boost and
+// 14 had their score genuinely shifted by it, across 9 types.
+//
+// That is wrong on its own terms — a confidence answers "how sure are we about THIS
+// value" — and it had a second-order cost: while the suppression hash folded confidence,
+// adding one unrelated finding to a file silently invalidated the saved rules of every
+// other finding in it, turning a reviewed-and-accepted finding back into noise.
+//
+// Driven through core.ScanFile rather than the bridge directly: the defect lived in how
+// the paths were combined, and a hand-wired bridge would test the test's own wiring.
+
+const unlabelledSecret = `value "wJalrXUtnFEMI7K7MDENGbPxRfiCYzz1"`
+
+// TestConfidenceIsIndependentOfUnrelatedFindings is the regression.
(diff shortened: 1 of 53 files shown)


sectoramen:
Superseded by #249, which is the same change cherry-picked directly onto `main` instead of onto #247's branch.

Both PRs touch zero files in common (#247: `internal/suppressions/*`; this change: `internal/validators/dual_path_bridge.go`, `internal/core/*_test.go`, and 7 golden cases), so #247 and #249 can now merge into `main` in either order with no conflict and no dependency between them.

Closing this rather than retargeting because a PR's head branch cannot be changed after creation.

PR #252, opened 2026-08-01
perf(redact): stop rescanning the whole document once per match

sectoramen:
## The defect

Redaction located every match by scanning the **entire document**, so cost grew with (matches × content bytes) — while scanning the same input stayed linear.

Measured end to end, N distinct SSNs one per line, `--checks ssn`:

| n | scan | redact |
|---|---|---|
| 2,000 | 0.51s | 0.49s |
| 4,000 | 0.30s | 1.36s |
| 8,000 | 0.48s | 4.59s |
| 16,000 | 0.88s | **17.45s** |

The adversarial review took the same shape to **78s on ~1MB** of dense matches. `file_router.go` `MaxFileSize` permits **100MB**.

## Isolating the mechanism

A fixed-match / growing-content family separates O(m) from O(m·n): with the match count **pinned at 100**, cost still tracked content — the signature of a per-match whole-document scan rather than per-match work.

Profiling (`-cpuprofile` + `pprof -peek`) attributed it to four sites, each scanning the full document once per match:

| site | operation |
|---|---|
| `plaintext/redactor.go` `findMatchPosition` | `strings.Index(document)` |
| `position/correlator.go` `extractTextAtPosition` | `strings.Split(document)` |
| `position/correlator.go` `calculateDocumentPosition` | `strings.Split(prefix)` |
| `position/correlator.go` `tryExactMatch` + `calculateExactMatchConfidence` | `strings.Count(document)` **twice** |

## The fix

- A **line-offset table computed once** per document, so a match is located within its own line. Safe because matches are applied in **descending line order**: every rewrite so far sits at a strictly later offset than any line still to be processed. A whole-document search remains as the fallback, so a match whose recorded line number does not hold its text is still located and still redacted.
- `lineAt` / `lineAndColumnAt` walk newlines instead of allocating every line.
- The occurrence count is computed **once** and shared instead of twice per match.

**Result: 17.45s → 11.32s at n=16,000 (1.54×)**, identical finding counts, redaction sink byte-identical before and after at n=2,000/8,000/16,000.

## Stated plainly: this is a constant-factor fix, not a complexity fix

Redaction is **still superlinear**. Measured growth ratios before and after **overlap** (12.6–13.7 vs 10.0–13.2). Making it truly linear requires carrying each match's byte offset from detection, which touches every validator and belongs in its own change. I am not claiming more than I measured.

## Complexity guard: the reason this survived

The guard **already existed** — 18 validator targets, non-vacuity `minMatches` floors, `wantMatchGrowth` assertions, ratio ceilings, a `-race` multiplier — but every target calls `validator.ValidateContent` directly to isolate validator cost from orchestration. `grep -ci redact` on it returned **0**. Redaction, suppression and `ResolveOverlaps` had no targets at all.

This PR adds the first redaction targets. Two design points learned the hard way:

1. **The sink is asserted on every timing sample.** A redaction that leaves the value in place is not a faster redaction, it is a leak — so an "optimisation" that skips matches cannot pass.
2. **My first version of this guard was vacuous** — it passed on the known-quadratic code, because at n=500→2000 the ratio sits just under the 12× ceiling. I measured the actual ratios on both versions, found they overlap, and rewrote the ceilings as an explicitly documented **regression backstop rather than a proof of linearity**. A guard that passes on the bug it was written for is worse than no guard.

## Plan correction

`FINAL_PLAN.md` Phase 2 proposed "add a complexity dimension". That was **wrong** — the gate exists. The document now records that the real gap was coverage past the validator boundary, plus the honest caveat about what a ratio ceiling can and cannot prove.

## Testing

| Gate | Result |
|---|---|
| Golden corpus | ok |
| `-race` (redactors + goldencorpus) | ok, no races |
| 3-platform vet (darwin/linux/windows) | ok |
| Redaction sink | byte-identical vs `main` at n=2,000/8,000/16,000 |
| Finding counts | identical at every size |
| Determinism | 1 distinct redacted output over 10 runs |
| End-to-end | 17.45s → 11.32s at n=16,000 |

### Union with open PRs

Base is `main`. Verified conflict-free **and** semantically green when merged with #247, #249, #250 and #251 — 0 test failures in each union, run in an isolated worktree.

### DIFF
--- a/docs/proposals/FINAL_PLAN.md
+++ b/docs/proposals/FINAL_PLAN.md
@@ -1,0 +1,164 @@
+<!--
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Final plan: what to fix, in what order, and why
+
+Status: 2026-08-01. Supersedes the ordering in `ARCHITECTURE_AND_PLAN.md`; that document remains
+the architectural description and the evidence record.
+
+This plan was written after an independent verification pass in which I re-tested the claims I had
+been repeating. **Three of them were wrong or incomplete**, and correcting them changed the
+ordering. Those corrections are in §5, because a plan built on unverified claims is worth less than
+the claims it rests on.
+
+---
+
+## 1. What changed from my own verification
+
+| Claim I had been carrying | What testing showed |
+|---|---|
+| "`--limit` is ignored by junit/csv/gitlab-sast/sarif" | **Wrong.** All four honour it. The real defect is worse: **four formats truncate SILENTLY at the DEFAULT limit of 200** — `sarif`, `csv`, `junit` emit 200 of 300 findings and never disclose the true total. `json`, `yaml`, `text`, `gitlab-sast` do disclose it. A CI gate consuming SARIF sees 200 of 300 by default. |
+| "reported ≠ redacted at exit 0, all matches log `match_redaction_failed`" | **Incomplete and understated.** Reproduced with a worse shape: a `.docx` where **every part comes out byte-identical** — zero redaction applied — while 5 findings are reported at rc=0 and **nothing** logs `match_redaction_failed`. The output md5 differs only from zip re-compression, so a `cmp`/md5 check reports "differs" and misleads. |
+| "the phone veto is a phone problem" | **Wrong — it is systemic.** The same prefix-a-character trick suppresses `VISA`, `DRIVERS_LICENSE` and `MEDICARE_MBI`, and the sink confirms all three sit in cleartext in the redacted output. This is 3+ validators, not one, which changes the fix from a one-file change to a pattern change. |
+
+And one hypothesis I would have accepted that turned out **false**:
+
+- "confidence is saturated at 100, so the bands carry little information." **Not true.** Over varied
+  content: 11 findings, **10 distinct confidence values** spanning 33–100, only 2 at the ceiling.
+  The scale is working. I have removed the recalibration item that assumed otherwise.
+
+One reassuring measurement worth stating, because it is the counterweight to the veto findings:
+
+- The negatives that the vetoes exist to suppress **are** suppressed correctly. A git SHA, a
+  version string `1.2.3.4`, and an `ami-050451375729` resource id all produce **zero** findings.
+  Any veto fix must preserve this. Removing the vetoes outright would be a bad trade.
+
+---
+
+## 2. The confirmed defect list, ranked
+
+Ranked by (likelihood × impact). "Ordinary" means no attacker and no unusual document.
+
+| # | Defect | Who it affects | Sink verified |
+|---|---|---|---|
+| **1** | **Office redactor silent no-op** — a `.docx` can come out with every part byte-identical, findings reported, rc=0, no error | **Ordinary.** Any `.docx` whose XML the redactor's matcher does not handle | Yes — SSN + PAN cleartext in "redacted" output |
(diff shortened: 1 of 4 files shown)


sectoramen:
## CI caught my own guard — fixed in `4e98229`

The new redaction guard failed on macOS and ubuntu under `-race`: **16.6x against a 16.0 ceiling**. It passed locally, so the ceiling had no margin for a slower shared runner.

**Raising the number would have been the wrong fix.** Measured spread for this exact fixture pair on one machine:

| mode | ratios |
|---|---|
| no `-race` | 9.3, 12.6, 12.7, 12.5, 12.9 |
| `-race` | 10.3, 9.5, 11.9, 12.2, 11.8 |
| CI runner | **16.6** |

And the **known-quadratic** code, before the per-match rescan was removed, measured **12.6–13.7**.

So any threshold high enough not to flake on 16.6 sits *above* the behaviour it exists to catch. The first version of this test managed to be both **vacuous** (it passed on the quadratic code) and **flaky** (it failed CI) — the worst of both.

### What it does now

- The **ratio is logged**, not asserted, for a human comparing runs.
- The assertion is an **absolute wall-time ceiling** — immune to noise in the base measurement, still fails on an order-of-magnitude regression. Scaled by `raceCeilingMultiplier`, reusing the build-tag helpers the validator guard already has rather than a second mechanism.
- Sized from measurement with headroom: big fixture ~150–250ms locally, ~440ms on a loaded runner, ceiling **8s**.

### The honest limit, stated in the file

This is a **regression backstop, not a proof of linearity**. I verified it still passes with the perf fix reverted — so it does not detect today's superlinear behaviour, only a large future regression. Making redaction genuinely linear needs match byte offsets carried from detection, which is its own change.

The assertions carrying the real weight are unchanged: mappings-produced == matches-in, match count must grow, count pinned in the fixed-match family, and **the redaction sink verified on every timing sample** so a "faster" redactor that leaks cannot pass.

Verified: 61 packages, `-race` x3 stable, 3-platform vet.

sectoramen:
## Correction: this PR had reverted its own fix. Restored in `e20c7a2`.

An external review pointed out that #252 "only changes 2 files", which contradicted what I believed was in it. They were right and I was wrong.

**What happened:** while fixing the guard's CI failure, I had copied this PR's files into another branch's worktree for a mutation test, then ran `git checkout --` on them here to clean up — which reset `redactor.go` and `correlator.go` to `main`. The next commit (`4e98229`) picked that up as part of its diff: **186 deletions**, stripping the entire perf fix.

Net effect: this PR shipped a complexity guard with **no perf fix**. Its own `findMatchPosition` was byte-identical to main's 2-arg version.

Verified three ways before fixing:
- `gh pr diff 252 --name-only` → 2 files
- `git diff origin/main...` → 2 files
- `git show --stat 4e98229` → the 186 deletions

**Restored** from `501c442`, keeping the guard changes. Re-measured rather than trusting the earlier numbers:

| n | main | restored |
|---|---|---|
| 8,000 | 4.65s | 3.75s |
| 16,000 | **16.11s** | **10.44s** (1.54×) |

Identical finding counts, and the redacted output is **byte-identical** to main's. 61 packages, `-race` clean, 3-platform vet ok.

This PR now contains all four intended files:
```
docs/proposals/FINAL_PLAN.md
internal/goldencorpus/complexity_guard_redaction_test.go
internal/redactors/plaintext/redactor.go
internal/redactors/position/correlator.go
```

**Lesson worth recording:** `git checkout --` to clean up after a mutation test silently reverted production code, and a commit whose message said it "only touched the test" carried that revert. Check `git diff --stat` against the PR's own subject before committing.

rustic (approved):

PR #253, opened 2026-08-01
fix(metadata): score a field's name as a name, not as its content

sectoramen:
## The bug

One root cause, two symptoms: the metadata validator scored a field's **NAME** as if it were the field's **content**. A field name is chosen by the file format (or, for a custom property, by the document's author). It says what *kind* of thing the field holds and nothing at all about whether the value is real or sensitive.

### Symptom 1 — `Template:` penalized its own value

`AnalyzeContext` matched its test-data denylist (`test, example, sample, demo, placeholder, dummy, template, default, anonymous, unknown`) against the whole line, so a field whose *name* collides with that list demoted itself. Measured with the value held byte-identical and only the label changed:

| line | ctx | net |
|---|---|---|
| `Template: \\corp-fs01\confidential\nightjar\t.dotx` | **-0.05** | **75** |
| `Tmpl: \\corp-fs01\confidential\nightjar\t.dotx` | +0.10 | 90 |
| `Template: Normal.dotm` (mundane) | -0.15 | 25 |

A UNC path disclosing an internal fileserver lost 15 points for being stored under a field called "Template", ranking it **below** a mundane company name at 55. Over 120 real Office documents this fired on **every one of the 80** `TEMPLATE_INFO` findings, plus image metadata's `BitsPerSample` ("sample") and `Custom_db_template_reference`.

### Symptom 2 — a value containing a field name stole that field's type and boost

`determineMatchType` and `applyPreprocessorConfidenceBoosts` had the mirror bug, matching field-name vocabularies against the **VALUE** (`match.Text`):

| line | reported as | conf | spurious boost |
|---|---|---|---|
| `Company: Acme Corp` | COMPANY_INFO | 55 | *(control)* |
| `Company: Manager Tools LLC` | **MANAGER_INFO** | **95** | manager_boost 40 |
| `Company: Author Solutions Inc` | **AUTHOR_INFO** | **85** | author_boost 30 |
| `Subject: comments on Q3` | **DOCUMENT_COMMENTS** | **100** | comments_boost 50 |
| `Title: Company retreat` | METADATA | 85 | company_boost 40 |

Real organizations are named "Manager Tools LLC" and "Author Solutions", so this is ordinary input, not anything crafted. The wrong type reaches the report, the redaction path **and the suppression hash**, and inflates by up to +40. In the corpus it promoted a Nasdaq classification footer to `AUTHOR_INFO` at 100 on the strength of the word "auth*orized*" in its value.

## The fix

One helper, `splitMetadataField`, splits a line once into `(name, value)`. The **name decides the TYPE**, the **value decides the CONFIDENCE**. `extractSensitiveValue` now reuses it instead of open-coding the same split.

The email branch in `determineMatchType` deliberately still reads the value — "contains an @-address" is a genuine value-shape signal and no field is named "@" — and stays below the name branches, so `Author: john.doe@example.com` remains `AUTHOR_INFO`.

### What I deliberately did NOT do

The boost fix stops crediting a field name found in a value; it does **not** re-point the table at the field name. Those are different changes and only the first is a bug fix:

- Because a value almost never repeats its own field's name, this table has been **dormant for its intended purpose since the initial commit**. `Manager: Dana Reyes` collected *no* `manager_boost`; `Company: Manager Tools LLC` collected all of it.
- `CalculateConfidence` **already** scores these same field names from the same line a few calls earlier (author +0.20, company +0.15, manager +0.25, comments +0.30). Adding the table on top double-counts one signal — precisely the *"field boosts double-count → most TPs HIGH"* note already in `docs/proposals/CONFIDENCE_CONTRACT.md`.
- **Measured**: waking it moved the corpus from **8 HIGH to 143 HIGH**, and what arrived in HIGH was boilerplate — `"Microsoft Office User"`, `"python-docx"`, `"Microsoft Office Word"`. HIGH is the band that fails a commit by default (`precommit.ExitOnFindings = "high"`), so I rejected that as a regression and narrowed the fix.

## Corpus impact — 120 real Office documents (.docx/.xlsx/.pptx)

**488 findings before, 488 after. Nothing lost.**

| band | before | after | delta |
|---|---|---|---|
| HIGH | 8 | 10 | +2 |
| MEDIUM | 265 | 271 | +6 |
| LOW | 215 | 207 | -8 |

| TYPE | before | after | delta |
|---|---|---|---|
| APPLICATION_INFO | 108 | 108 | 0 |
| AUTHOR_INFO | 108 | 106 | **-2** |
| COMPANY_INFO | 28 | 28 | 0 |
| DOCUMENT_DESCRIPTION | 14 | 14 | 0 |
| EMAIL | 1 | 1 | 0 |
| IMAGE_SOFTWARE | 2 | 2 | 0 |
| LAST_MODIFIED_BY | 111 | 111 | 0 |
| METADATA | 36 | 32 | **-4** |
| TEMPLATE_INFO | 80 | 86 | **+6** |

**Every nonzero delta explained** — they are the same 6 findings retyped, no additions or removals:

- **TEMPLATE_INFO +6 / METADATA -4**: three custom properties named `Custom__TemplateID`, `Custom_db_template_reference`, `Custom_db_template_version` (2 occurrences each = 6). Their *names* literally contain "template", so `TEMPLATE_INFO` is the correct type; they were `METADATA` before because the name was never consulted. 4 of the 6 came from METADATA.
- **AUTHOR_INFO -2**: the Nasdaq classification footer, the remaining 2 of the 6. `Custom_ClassificationContentMarkingFooterText` is not authorship; it was typed `AUTHOR_INFO` only because its value contains "authorized". Now `METADATA` at 90 — **still reported, still HIGH**.
- **Band shift**: 77 confidence changes, all `Template` fields recovering the self-penalty (67 at 25→40, 8 at 45→60, 2 at **75→90** — the UNC-path disclosures now correctly HIGH).

## Golden corpus

One case changes, and **it is itself an instance of the bug**: the WAV `Comment` field collected audio metadata's `contact_boost` of 50 because its VALUE began `"contact 212-555-0142"`, reaching 100. It is now 70. The phone and AWS key on that line are still detected separately — **5 findings before, 5 after**.

## Mutation proof

All three production changes are independently mutation-proven. Each revert **still compiles** (an unused-import build failure is not a proof):

**M1** — revert `AnalyzeContext` to scoring the whole line:
```
--- FAIL: TestAnalyzeContext_FieldNameIsNotTestData
    field_name_vs_value_test.go:48: field name changed the context score for an identical value:
    field_name_vs_value_test.go:60: a field named "Template" still penalizes its own value: got -0.05, want >= 0
--- FAIL: TestAnalyzeContext_TestDataInValueStillSuppressed
    field_name_vs_value_test.go:99: field name was scored as test data for "Template: Normal.dotm": got -0.15, want >= 0
    field_name_vs_value_test.go:99: field name was scored as test data for "BitsPerSample: 8": got -0.15, want >= 0
```

**M2** — revert `determineMatchType` to matching the whole line:
```
--- FAIL: TestValueContainingFieldNameKeepsItsOwnFieldsIdentity
    "Company: Manager Tools LLC": value stole another field's TYPE: got MANAGER_INFO, want COMPANY_INFO
    "Company: Author Solutions Inc": value stole another field's TYPE: got AUTHOR_INFO, want COMPANY_INFO
--- FAIL: TestFieldNameDecidesTypeAcrossFields
    "Description: camera setup": got type DEVICE_INFO, want DOCUMENT_DESCRIPTION
    "Template: authored by hand": got type AUTHOR_INFO, want TEMPLATE_INFO
```

**M3** — neutralize the boost field-name guard (`if false && ...`, so it compiles):
```
--- FAIL: TestValueContainingFieldNameKeepsItsOwnFieldsIdentity
    "Company: Manager Tools LLC": value collected another field's confidence boost: map[manager_boost:40]
    "Company: Comments Media Group": value changed the confidence of its own field: got 100, want 55
--- FAIL: TestBoostRequiresTheFieldNameToAgree
    a boost key found only in the VALUE was credited: map[contact_boost:50] (field is named "Comment", not "contact")
```

**Vacuity guards**: `TestAnalyzeContext_TestDataInValueStillSuppressed` asserts 10 genuine value-borne markers *are* still demoted (`Author: test user`, `Tmpl: sample.dotx`, …), so deleting the suppression outright fails. `TestBoostRequiresTheFieldNameToAgree` asserts a field genuinely named `Artist` *keeps* its boost, so deleting the boost loop fails.

## Performance

Per-line hot path, so measured before/after (min of 3, `-benchtime 200x`):

| benchmark | before | after | ratio |
|---|---|---|---|
| determineMatchType | 441.0 ns | **88.1 ns** | **0.20x (5.0x faster)** |
| AnalyzeContext width=10 | 842.5 ns | 870.4 ns | 1.03x |
| AnalyzeContext width=40 | 3009 ns | 2972 ns | 0.99x |
| AnalyzeContext width=160 | 8664 ns | 8521 ns | 0.98x |
| AnalyzeContext width=640 | 31157 ns | 32492 ns | 1.04x |
| ValidateMetadataContent lines=100 | 523 µs | 531 µs | 1.02x |
| ValidateMetadataContent lines=800 | 4443 µs | 4488 µs | 1.01x |

`determineMatchType` got **5x faster** because it no longer lowercases the whole line, only the name. Growth stays linear (100→800 lines is 8.4x for 8x input).

**A regression I found and fixed**: my first version of `AnalyzeContext` cost **1.24–1.38x** from a second whole-line `strings.ToLower`. Splitting the *already-lowercased* string removes it — `strings.ToLower` cannot move an ASCII `:`/`=` byte offset. Now within noise.

## Test plan

| gate | result |
|---|---|
| `go test ./... -count=1` | **ok, 61 packages** |
| `go test ./internal/goldencorpus/...` | **ok** (1 case updated, explained above) |
| `go test -race` (metadata, goldencorpus, router) | **ok** |
| `make vet` (darwin / linux / windows) | **ok / ok / ok** |
| `gofmt -l` | **clean** |
| Determinism, 10 runs over 120 documents | **1 distinct result** |
| Corpus FP impact | **488 → 488, 0 lost**, every delta explained |
| Benchmark before/after | **no regression**; one function 5x faster |

Determinism note: hashing raw output gives 10 distinct results, but the *only* varying field is the wall-clock `duration_seconds` — unmodified `main` varies identically. Excluding it: **1 distinct result over 10 runs**, findings byte-identical.

## Merge check against every open PR

| PR | textual | semantic (`go test ./...` on merged scratch) |
|---|---|---|
| #252 perf/redaction-line-offset-table | 0 conflicts | **ok, 61** |
| #251 fix/metadata-value-shape-scoring | 0 conflicts | **BUILD FAILED** — see below |
| #250 fix/overlap-cross-part-line-collision | 0 conflicts | **ok, 61** |
| #249 fix/cross-path-correlation-boost-solo | golden-only conflicts | **ok, 61** after regenerating |
| #247 fix/suppression-hash-confidence | 0 conflicts | **ok, 61** |

**#251 is a real semantic collision that textual cleanliness missed** — exactly what this check exists for. #251 adds a `custom_` branch at the top of `determineMatchType` that reads `lineLower`; this PR renames that variable to `nameLower`. Git merges both hunks silently and the result does not compile:

```
internal/validators/metadata/metadata_validator.go:1351:23: undefined: lineLower
```

Resolution is one word — `lineLower` → `nameLower` — and it is also the *semantically correct* result, since #251's intent is to match the custom-property prefix against the field name. With that change the union builds and passes all 61 packages. Whichever of #251/this merges second should apply it.

**#249** conflicts only in the regenerable `file_wav_metadata_pii.*` snapshots, because both PRs legitimately change that case (#249 removes `correlation_boost`, this removes `contact_boost`). Resolved by regenerating; both fixes compose correctly and the case keeps all 5 findings.

## Not verified / out of scope

- **Office metadata is not redacted**, before or after this change: on a synthesized `.docx` the redacted output still contains `Manager Tools`, `corp-fs01` and `Jane Analyst` in `docProps/*.xml`. Identical before and after, so this PR neither causes nor fixes it. Pre-existing gap, separate PR.
- `determineImageMetadataType` still receives the whole line. Unlike the branches above it legitimately classifies by value as well as by name (`Software: Adobe Photoshop` is `IMAGE_SOFTWARE` because of its value), so narrowing it is a different question and I left it alone.
- The dormant boost table is left dormant, as argued above. Whether metadata should have field boosts *at all* is a scoring-design question for the confidence contract, not a bug fix.


sectoramen:
## Merge-order notes (verified in isolated worktrees, not inferred)

I independently re-verified the union situation for this PR. Two findings, both actionable:

### 1. The `lineLower` collision is FIXED in this branch

Confirmed the original problem first: merging #251 and this PR produced **zero textual conflicts** and then **failed to compile** — `undefined: lineLower` at `metadata_validator.go:1351`. #251 adds a `custom_` branch to `determineMatchType` that reads `lineLower`; this PR renamed that variable to `nameLower`. Git merged both hunks silently.

Commit `61a495c` makes the merge correct in **either direction** rather than relying on a manual resolution: `lineLower` stays bound as an alias and the `custom_` branch is carried here too. A field-name test spelled with either identifier is the same test, since the name is a prefix of the line.

Verified after the fix:

| union | build | tests |
|---|---|---|
| this PR ← #251 | OK | 61 packages, 0 failures |
| #251 ← this PR | OK | 61 packages, 0 failures |
| this PR + #247 | OK | 61 packages, 0 failures |
| this PR + #250 | OK | 61 packages, 0 failures |
| this PR + #252 | OK | 61 packages, 0 failures |

### 2. There IS a real conflict with #249 — golden files

Not reported earlier. Merging with #249 (`fix/cross-path-correlation-boost-solo`) conflicts on **7 golden files** for the same case:

```
internal/goldencorpus/testdata/golden/file_wav_metadata_pii.{txt,yaml,csv,json.json,junit.xml,sarif.json,gitlab-sast.json}
```

Both PRs legitimately move the same finding's confidence in opposite directions:

- this PR raises `AUTHOR_INFO` to **85** (name-based typing)
- #249 lowers it to **80** (removing the +5 cross-path correlation boost)

**Resolution, verified working:** take #249's side and regenerate — `UPDATE_GOLDEN=1 go test ./internal/goldencorpus/`. `AUTHOR_INFO` settles at **80**, which is both fixes composing correctly (name-based type, no phantom boost). Union then passes **61 packages, 0 failures**.

Whichever of these two lands second needs that one regeneration step. Nothing else.

sectoramen:
## Sequential-merge verification (simulated, in order, with **squash**)

Correcting my earlier comment. I re-verified using `git merge --squash` — which is what this repo actually does; every commit on `main` is single-parent — instead of a plain merge. **The plain-merge chain is clean and the squash chain is not**, so the earlier plain-merge result was the wrong thing to trust.

### Result: 5 of 6 merge clean; this PR conflicts

| step | squash | build | tests |
|---|---|---|---|
| #247 | clean | OK | 0 failures |
| #249 | clean | OK | 0 failures |
| #250 | clean | OK | 0 failures |
| #251 | clean | OK | 0 failures |
| #252 | clean | OK | 0 failures |
| **#253** | **CONFLICT (7 golden files)** | — | — |

### The conflict is inherent — I tried to engineer it away and could not

`internal/goldencorpus/testdata/golden/file_wav_metadata_pii.{txt,yaml,csv,json.json,junit.xml,sarif.json,gitlab-sast.json}`

Both PRs rewrite **all five findings of the same block**:

```
after #249                                    this PR
DOCUMENT_COMMENTS  100.00   <- position 1     AWS_ACCESS_KEY      89.00
AWS_ACCESS_KEY      84.00                     PHONE               87.00
PHONE               82.00                     AUTHOR_INFO         85.00
AUTHOR_INFO         80.00                     DOCUMENT_COMMENTS   70.00  <- position 4
BUSINESS            25.00                     BUSINESS            30.00
```

#249 changes the **numbers** (removing the +5 correlation boost); this PR **reorders the block**, because `DOCUMENT_COMMENTS` drops from first to fourth as its score falls. Overlapping rewrites of the same five lines cannot auto-merge.

Things I tested that do **not** help:
- **Reordering** — four different sequences; whichever of #249/#253 lands second conflicts.
- **Adopting #249's numbers here** — makes this PR fail its own golden test standalone (verified: `--- FAIL: TestGoldenFileScanFormats/file_wav_metadata_pii`). The two genuinely disagree until both are present.
- **A merge-free branch** (cherry-picking this PR's two commits onto `main`) — still conflicts, which disproved my merge-commit theory.

### Resolution: one mechanical step, verified end to end

```
git checkout --theirs internal/goldencorpus/testdata/golden/file_wav_metadata_pii.*
git add internal/goldencorpus/testdata/golden/file_wav_metadata_pii.*
UPDATE_GOLDEN=1 go test ./internal/goldencorpus/
```

Result — both fixes composing correctly, `DOCUMENT_COMMENTS` at **65** = this PR's 70 minus #249's removed 5:

```
AWS_ACCESS_KEY      84.00
PHONE               82.00
AUTHOR_INFO         80.00
DOCUMENT_COMMENTS   65.00
BUSINESS            25.00
```

**Full chain after resolution: `go build` OK, 61 packages, 0 test failures.**

Either side may be taken before regenerating — the regeneration is what makes it correct, and I verified the regenerated output rather than assuming it.

rustic (approved):

PR #260, opened 2026-08-03
fix(scan): stop files vanishing when every preprocessor fails

### LINKED ISSUE #257: A file whose preprocessors all fail vanishes from the scan: not processed, not skipped, no warning, exit 0
## Summary

When every preprocessor fails for a file, the router returns `nil` content and the file **disappears from the scan entirely** — it is counted as neither processed nor skipped, produces no warning, and does not affect the exit code even under `--fail-on-incomplete`.

Reproduced with a directory of three files — one valid `.docx` and two unparseable ones (a zero-byte `.docx` and a 1-byte `.docx`):

```
$ ferret-scan --config /dev/null --confidence all --format text --recursive <dir>
...
Files: 1 processed, 0 skipped | Findings: 7
```

**Three files in, one accounted for.** No warning naming the other two. With `--format json`, a directory containing only the two unparseable files returns `[]` and exit 0.

```
$ ferret-scan --format json --limit 0 --recursive <dir-with-2-corrupt-docx>
[]
$ echo $?
0
```

`--fail-on-incomplete` also returns 0.

## Why this matters

This is the *silent* failure shape, which is the worst one for a scanner: the operator gets a clean report and a success exit code for files that were never examined. A corrupt, truncated, or partially-downloaded document — exactly the kind of file that might be carrying data someone tried to hide — is indistinguishable from a directory that was fully scanned and clean.

Contrast the two diagnostics that already work correctly and are the model to follow here:

- **unreadable files** — `WARNING: scan incomplete — N of M file(s) could not be opened ... any sensitive data they contain was NOT detected`, and exit 3 under `--fail-on-incomplete`
- **empty extraction** — `WARNING: extraction empty — N of M file(s) yielded no document text`

Both name the files. This case names nothing.

## Cause

`internal/router/file_router.go:413`:

```go
return nil, fmt.Errorf("all preprocessors failed for file: %s", filePath)
```

Returning `nil` content means the worker has no `ProcessedContent` to attach a diagnostic to, so the downstream accounting in `internal/parallel/parallel_processor.go` increments neither the processed nor the skipped counter — the file falls out of both branches.

## This was independently predicted for ODF

A component review flagged the same code path for `.ods`/`.odt` files whose extraction yields empty text, noting that unlike `.docx` there is no second capable preprocessor to keep the file alive, so the `ExtractionWarning` that *was* correctly generated gets discarded by the router.

This report confirms the mechanism on a **different format** (`.docx`) via a different trigger (unparseable container rather than empty text), which suggests the defect is in the router's error path rather than in any one extractor.

## Suggested fix

Do not discard a `ProcessedContent` that carries a warning. When every preprocessor fails, return a non-`nil` `ProcessedContent` with `Success=false` and the joined extraction warnings, so the existing worker/collector plumbing can count the file and print a diagnostic. Independently: a file that cannot be processed at all should count as **skipped** rather than disappearing.

## Acceptance criteria

- [ ] A directory of N files where some are unparseable reports all N in `processed + skipped`
- [ ] The unparseable files are named in a warning, in the style of the existing two diagnostics
- [ ] `--fail-on-incomplete` returns a non-zero exit code when files were not scanned
- [ ] A regression test covering an unparseable container that **fails on today's code**
- [ ] Verify on `.ods`/`.odt` as well as `.docx`, since the ODF path has no fallback preprocessor

## Priority

High for a security tool. No detection is wrong here — the problem is that the *absence* of detection is invisible, which is the one failure mode a scanner must never have.

sectoramen:
Fixes #257.

## The defect

A file whose processing errored was counted as **neither processed nor skipped**. The result collector logged `Result.Error` and fell through without incrementing a counter or recording a diagnostic, so the file left no trace anywhere.

Measured on six files — five unparseable containers (0-byte `.docx`, 1-byte `.docx`, non-zip `.docx`, 0-byte `.ods`, 1-byte `.odt`) plus one valid `.docx`:

```
Files: 1 processed, 0 skipped | Findings: 7
(no warning, exit 0 even with --fail-on-incomplete)
```

**Six files in, one accounted for.** This is the worst failure shape for a scanner: a corrupt or truncated document — exactly the kind that might be hiding something — was indistinguishable from a directory that had been fully scanned and found clean.

## The fix

Follows the four diagnostics that already work rather than inventing a mechanism. `ProcessingStats.FailedFiles` is populated from `Result.Error`, surfaced by `writeFailedFilesWarning` (stderr only, pre-commit gated, like its siblings), and counted toward `--fail-on-incomplete`.

After:

```
WARNING: processing failed — 5 of 6 file(s) could not be processed, so they were
NOT scanned; any sensitive data in them was NOT detected:
  /tmp/van/empty.docx: all preprocessors failed for file: ...
  /tmp/van/empty.ods: all preprocessors failed for file: ...
  /tmp/van/notzip.docx: all preprocessors failed for file: ...
  /tmp/van/onebyte.docx: all preprocessors failed for file: ...
  /tmp/van/onebyte.odt: all preprocessors failed for file: ...
$ echo $?
3
```

**ODF is covered deliberately, not incidentally.** `.ods`/`.odt` have no second capable preprocessor to keep a file alive, so they are the formats most exposed to this. Both appear in the fixture and the test.

## Also fixes a pre-existing nondeterminism I found while testing this

The diagnostic lists are appended in worker-**completion** order, so the same scan printed the same files in a different sequence every run:

| list | runs | distinct orderings |
|---|---|---|
| empty-extraction (8 files) | 5 | **5** |
| failed files (5 files) | 6 | **6** |

That reaches operator-visible stderr, so diffing two scans of an unchanged tree showed spurious changes. All four lists are now sorted by path. They are per-run and small, so the cost is nil, and completion order carries no meaning to a reader.

To be clear about attribution: the nondeterminism was **already there** in the existing lists — I verified it on `empty-extraction` before adding anything. I'm fixing the family rather than adding a fifth member with the same flaw.

## Testing

Both tests **mutation-proven with compiling reverts** (a build failure is not a proof):

```
# no-op sort comparator
--- FAIL: TestDiagnosticListsAreSortedByPath
    diagnostics not sorted by path

# drop the FailedFiles append
--- FAIL: TestFailedFilesAreRecordedNotSilentlyDropped
    processed(1) + failed(0) = 1, want 4 — a file that is in neither bucket has
    vanished from the run with no warning and no effect on the exit code
```

The accounting test asserts the real invariant — **every file is either processed or recorded** — plus that each entry carries a path *and* a reason, since a nameless failure is not actionable.

| Gate | Result |
|---|---|
| `-race` (parallel, core, goldencorpus) | ok, 0 data races |
| `gofmt` | clean |
| 3-platform vet | darwin / linux / windows ok |
| Spurious warnings on healthy files | **0** |
| Finding counts | unchanged (10 MB docx 78 → 78) |
| Exit code on a clean scan | still 0 |
| Determinism after fix | 1 ordering over 6 runs |

### Merge safety

Branched off `main` (`113d370`, all 9 prior PRs landed). Touches `internal/parallel/`, `internal/core/scanner.go`, `cmd/main.go` — no overlap with any other open PR.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -747,6 +747,34 @@ func writeEmptyExtractionWarning(w io.Writer, emptyFiles []parallel.FileDiagnost
 	return true
 }
 
+// writeFailedFilesWarning emits a warning to w for every file whose processing
+// returned an error, so it was never scanned. It returns true if a warning was
+// written.
+//
+// The fifth member of the family, and it existed because of the worst gap of the
+// five: these files were counted as NEITHER processed nor skipped. The collector
+// logged the error and fell through without incrementing a counter or recording a
+// diagnostic, so the file left no trace at all. Measured on six files where five
+// were unparseable containers: "Files: 1 processed, 0 skipped", no warning, and
+// exit 0 even under --fail-on-incomplete. A corrupt or truncated document read
+// exactly like a directory that had been fully scanned and found clean.
+//
+// Same conventions as its siblings: gated only on pre-commit mode, stderr only
+// (never stdout/JSON), and it does not change the exit code by itself — these
+// files are counted toward --fail-on-incomplete alongside cut-short, unreadable
+// and empty-extraction ones.
+func writeFailedFilesWarning(w io.Writer, failedFiles []parallel.FileDiagnostic, totalFiles int) bool {
+	if len(failedFiles) == 0 {
+		return false
+	}
+	fmt.Fprintf(w, "WARNING: processing failed — %d of %d file(s) could not be processed, so they were NOT scanned; any sensitive data in them was NOT detected:\n",
+		len(failedFiles), totalFiles)
+	for _, fd := range failedFiles {
+		fmt.Fprintf(w, "  %s: %s\n", fd.FilePath, fd.Reason)
+	}
+	return true
+}
+
 // writeUnreadableFilesWarning emits a warning to w for every file that could not
 // be opened or stat'ed — a permission error, a dangling symlink, a file deleted
 // mid-scan. It returns true if a warning was written.
@@ -1598,6 +1626,12 @@ func main() {
 	// stderr warning and counted as a coverage gap: nothing was extracted, so no
 	// validator saw anything, so the file reported clean.
 	var emptyExtractionFiles []parallel.FileDiagnostic
+	// failedProcessingFiles captures files whose processing errored, so they were
+	// never scanned at all. Without this they were counted as neither processed
+	// nor skipped and vanished from the run entirely. Named distinctly from the
+	// local failedFiles int below, which means "processed but produced no
+	// findings" — a different and far less serious thing.
+	var failedProcessingFiles []parallel.FileDiagnostic
 
 	// Report config keys the schema does not recognize. Gated only on pre-commit
 	// mode, NOT on quiet/non-interactive — the same rule as the
@@ -1737,6 +1771,7 @@ func main() {
 			incompleteFiles = stats.IncompleteFiles
 			unredactedFiles = stats.UnredactedFiles
 			emptyExtractionFiles = stats.EmptyExtractionFiles
+			failedProcessingFiles = stats.FailedFiles
 
 			// Handle inline redaction results if redaction was enabled
 			if finalConfig.enableRedaction && redactionManager != nil {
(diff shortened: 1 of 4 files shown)


sectoramen:
## The Windows CI failure here is **not** from this PR

`windows-latest / Go` is failing on `TestUnreadableFileIsStillReported`, which lives in `internal/router/` — a file this PR does not touch.

It came from #256 (already merged) and has left **main's own "Go Tests" job red since `f245fdf`**, so every PR opened since inherits it. Cause: the test simulates an unreadable file with `os.Chmod(f, 0o000)`, which does not deny read access on Windows, and my `os.Geteuid() == 0` guard never fires there because `Geteuid` returns `-1`.

Fixed separately in **#261**, which touches only that one test file — no overlap with this PR.

Once #261 lands, this PR's Windows job should go green with no change here. Ubuntu and macOS already pass on this branch.

rustic (approved):

PR #262, opened 2026-08-04
fix(redact): redact Office document properties, and cover containers in golden

sectoramen:
## The defect

Metadata findings were **reported and never removed**. The scan named an author, a company, a template path — then wrote a "redacted" copy with every one of them still in cleartext inside `docProps/*.xml`.

That is the worst combination for a redaction tool: the report proves it saw the value, and the artifact you hand to someone else still contains it.

## Two independent causes, each separately necessary

Disabling either one alone puts **all 7** reported values back in the output:

1. `isTextContainingFile` required a body-part prefix (`word/`, `xl/worksheets/`, `ppt/slides/`) for every document type, so `docProps/*` was never selected and the redactor never rewrote it.
2. **Selecting the part is only half.** Its values live in `dc:title`, `dc:creator`, `cp:keywords`, `Company`, `Manager`, `Template` — none of which are `w:t` or a spreadsheet cell, so `isTextElement` rejected them. With the part selected but the elements unrecognized, the part extracts to nothing: the same silent no-op reached a different way.

Measured on a real fixture: **7 reported findings, 7 leaks before, 0 after.** The output stays a valid zip with all its parts, and body redaction is unaffected.

Structural and numeric properties are deliberately **excluded** (`Pages`, `Words`, `Characters`, `AppVersion`, `TotalTime`, `DocSecurity`, `dcterms` dates) — rewriting them would corrupt the document for no privacy benefit. A test pins both directions.

## Also lands the FileCases redaction coverage this depended on

`TestGoldenRedact` iterates `Cases` (in-memory strings) through `redact.Engine`, whose `Request` takes `Text` — no file path, so **no container ever reached it**. `FileCases` had *zero* redaction coverage, which is why every container leak (this one, #250, the Office repackaging no-op) was found by hand rather than by the suite.

The new guard asserts the **sink**, not a snapshot: for each case it redacts via `core.RedactFile` and checks that no reported value survives in any part, read **decompressed**. That detail matters — grepping the `.docx` bytes finds nothing whether or not redaction worked, which is how one of these leaks survived an md5 check that reported "differs".

**10 of 12 cases now assert fully and pass.** The 2 skips are `.go` and `.wav`, which have no registered redactor; the CLI already discloses that as `redaction incomplete — the original values remain in cleartext`. PDF is the same shape and its message is exemplary: *"refusing to produce an output that was not redacted but would falsely report success."*

## Complexity — checked, and honest about what it shows

Selecting `docProps` means it is now extracted and scanned, so this needed measuring:

| shape | ratios per doubling |
|---|---|
| growing docProps, **fixed** match count | 1.30 / 1.70 / 1.53 — **linear** |
| matches **and** content growing | 2.98 / 3.50 / 3.48 — superlinear |
| **body path on unmodified `main`** | 3.24 / 3.41 / 3.38 — **the same** |

So the superlinear shape is the Office redactor's **pre-existing** per-match whole-part scan, not something added here. It is the same defect #252 fixed for plaintext and deserves its own PR; this change does not make it worse.

## Testing

| Gate | Result |
|---|---|
| `-race` (redactors + goldencorpus) | ok, 0 data races |
| `gofmt` / 3-platform vet | clean; darwin / linux / windows ok |
| Golden corpus | unchanged |
| Determinism | redacted output byte-stable over 6 runs |
| Regression, 3 real documents | findings unchanged (78 / 7 / 7) |
| All 7 output formats | render |
| Every registered redactor | leak-free artifact |

Both new test files are **mutation-proven** against each half of the fix, with compiling reverts.

### Merge safety

Branched off `main`. **Zero shared files** with #260 or #261. Verified with a squash-merge simulation in the intended order — `main → #260 → #261 → this` — clean at every step, build OK, 0 test failures.

### DIFF
--- a/internal/goldencorpus/golden_file_redact_test.go
+++ b/internal/goldencorpus/golden_file_redact_test.go
@@ -1,0 +1,198 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package goldencorpus
+
+import (
+	"archive/zip"
+	"bytes"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/awslabs/ferret-scan/v2/internal/config"
+	"github.com/awslabs/ferret-scan/v2/internal/core"
+)
+
+// TestGoldenFileRedactionSink closes the coverage gap that let every container
+// redaction leak of the last several fixes hide.
+//
+// TestGoldenRedact iterates Cases (in-memory strings) through redact.Engine,
+// whose Request takes Text — there is no file path, so no container ever reaches
+// it. FileCases, which is where the .docx/.xlsx/.wav cases live, therefore had
+// ZERO redaction coverage: `grep -ci redact` over the FileCases test path
+// returned 0. Every leak in that area was found by hand, by unzipping an output
+// and grepping the parts, and fixed the same way:
+//
+//   - #250: ResolveOverlaps compared offsets across different source texts, so a
+//     metadata match silently dropped a body SSN from redaction
+//   - the Office redactor writing a package with every part byte-identical
+//   - a .docx inside a .docx whose inner body was never scanned
+//
+// None of those had a test that would have failed. This one would.
+//
+// It asserts the SINK, not a snapshot. A golden text file records what the
+// formatter said; this checks the thing that actually matters — whether the
+// sensitive bytes survive in the artifact handed to a third party. And it looks
+// INSIDE containers, because an OOXML part is deflated: grepping the .docx
+// bytes finds nothing whether or not redaction worked.
+func TestGoldenFileRedactionSink(t *testing.T) {
+	for _, fc := range FileCases {
+		fc := fc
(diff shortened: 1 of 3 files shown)


sectoramen:
## The Windows failure here is inherited, not from this PR

`windows-latest / Go` fails on `TestUnreadableFileIsStillReported`, which lives in `internal/router/` — a package this PR does not touch:

```
$ gh pr diff 262 --name-only
internal/goldencorpus/golden_file_redact_test.go
internal/redactors/office/docprops_redaction_test.go
internal/redactors/office/redactor.go
```

It is my own test bug from #256 (`os.Chmod(f, 0o000)` does not deny read access on Windows), which has left **main's "Go Tests" job red since `f245fdf`**. Every PR opened since inherits it.

**#261 is the fix and is currently OPEN / MERGEABLE / CLEAN.** Verified that it resolves this: merging #261 into this branch is clean, `GOOS=windows go vet ./internal/router/` passes, and the router tests pass.

No change needed here — merging #261 first clears it. Ubuntu and macOS already pass on this branch.

rustic (approved):

PR #275, opened 2026-08-05
feat(test): score detection quality across four layers, and fix a name-redaction leak it found

sectoramen:
The rest of the suite proves output did not **change**. Nothing proved output was **right**.

`internal/goldencorpus` says so itself ([corpus.go:10](https://github.com/awslabs/ferret-scan/blob/main/internal/goldencorpus/corpus.go#L10)): *"The purpose is NOT to assert that any particular detection is correct."* So a change that trades a real detection for a quieter report passes the golden net — regenerate the snapshot and move on — and passes every unit test. This adds the gate that refuses it, and ships the first bug it caught.

## Why one precision/recall number is not enough

A single detection score is blind to the failure class this tool actually ships. Three regressions, each measured on `main`:

| mutation | full suite | detection score | redacted artifact |
|---|---|---|---|
| a "fixed-width report noise" filter | **rc=0, 64 pkgs ok** | 111 → 108 TP | **3 whole-value leaks** |
| revert [#250](https://github.com/awslabs/ferret-scan/pull/250) (Office redactor) | rc=1 | **bit-identical** | **1 container leak** |
| band-demote a bare 9-digit SSN | **rc=0, 64 pkgs ok** | `recall_all` unchanged | pre-commit stops blocking |

Row two is the point: **reverting a real cleartext-leak fix leaves the detection score bit-for-bit identical**, because the value was detected correctly and then dropped during redaction. Rows one and three pass the entire existing suite today.

So the gate scores **four layers**, each of which can fail while the others look perfect:

| layer | question | what breaks it alone |
|---|---|---|
| **validator** | detected, right type, at or above its band? | a veto, a scoring change, a regex narrowing |
| **redaction** | are the value's bytes gone from the artifact? | overlap resolution, a container repack bug |
| **suppression** | does a rule silence *exactly* what it names? | an over-broad hash → a leak with an audit trail saying it was approved |
| **executable** | does the real CLI report it, exit right, write the file? | [#193](https://github.com/awslabs/ferret-scan/pull/193) silent output, [#270](https://github.com/awslabs/ferret-scan/pull/270) rc inversion |

The executable layer exists because every other layer calls the library in-process, and **that is not the product**: #193 shipped rc=1 with zero bytes on stdout *and* stderr. Neither that nor #270 moves a precision number.

**Recall is reported twice, on purpose.** `recall_all` (all bands) is the **redaction** surface — redaction is confidence-blind, measured: a bare 9-digit SSN at confidence 50 still redacts to `*****5728`. `recall_hm` (≥MEDIUM) is the **pre-commit exit-code** surface, the only one a band drop moves.

## What it measures today

```
scorecorpus  143 cases, 180 gated labels, 16 of 19 checks
TOTAL        180 TP   0 FN   47 FP   recall_all 1.0000  recall_hm 0.9944  prec_hm 0.7920
sink         simple whole_leak=0 residue4=0 | format_preserving whole_leak=0 residue4=829
suppression  57 rules, 57 silenced, 0 collateral, 0 ineffective
executable   180 labels verified through the real binary
```

Ground truth comes from bytes **planted** in each fixture, never from what the tool reports — otherwise the corpus certifies today's behaviour as correct by construction. Labels key on `(line, value, occurrence)` with **no byte span**, because `detector.Match` has no offset field; `TestValueOccursOnce` pins the invariant that makes that sound.

The **47 false positives are a real defect**, baselined so a later fix can be *measured* rather than argued. A baseline is a floor to ratchet, never a goal.

## The bug this found: hyphenated names left in cleartext

A name with a hyphen in **both** parts was never matched in full. Two patterns existed — `hyphenated_last_name` (`First Last-Name`) and `compound_first_name` (`First-Name Last`) — and each claimed a different **partial**, so nothing spanned the whole name. The shortfall reached the redacted file:

```
Anne-Marie Delacroix-Webb  ->  Reviewed by Anne-********************
Mary-Jane Watson-Parker    ->  Contact ****************-Parker today
Jean-Claude Van Damme      ->  Signed: Jean-****************
```

Real name fragments in cleartext, in a file the tool calls redacted. Fixed by adding the two missing combinations. All three now mask completely, and **PERSON_NAME went from 1 TP / 2 FP to 4 TP / 0 FP** — the overlapping partials *were* the false positives, so recall and precision improved together.

Hyphenated surnames are common and disproportionately non-Anglo, so this is not an edge case.

## Two settled questions

**24 national-ID fixtures are labelled TRUE POSITIVE and scored.** They carry an SSN-shaped value under an honest header for a non-US identifier (`sin`, `nino`, `personnummer`, `codice fiscale`). Decided on evidence: the SSN validator's own `positiveKeywords` already contain `"national id"`, `"government id"`, `"federal id"`, `"tax id"` — it was **built** for these. And the options were not symmetric: labelling them FP would make *deleting real PII detections* score as an improvement. Effect: SSN precision 0.7115 → **0.7739**; recall floor 111 → **155**.

**An improvement is reported and does NOT fail the build.** The alternative keeps the floor rising automatically but taxes every PR that makes the tool better. Stated tradeoff: a win never locked in with `make score-update` can be given back later without the gate noticing.

## Testing

- **26 tests. 21 guard the harness, not the product**: labels resolve to their own fixture; no aspirational labels (one the tool cannot satisfy would sit in the recall denominator forever); both polarities present; bands track `shared.GetConfidenceLevel` at 59.9/60/89.9/90; no machine dependence (`config.LoadConfig("")`, not `LoadConfigOrDefault`, which discovers `~/.ferret-scan` and changed the validator count 2 vs 0).
- **Non-vacuity, each proven with a COMPILING mutation** via `make score-mutation-check`: M1 fixed-width veto → recall **and** sink red; M2 revert #250 → detection unchanged, container sink red; reverting the PERSON_NAME fix → residue 0 → 17 bytes, red. Control green on a clean tree.
- **Determinism**: 6 in-process runs and 3 separate processes, identical.
- **Linear**: 2.00× per doubling at n=125…2000, with a vacuity floor asserting every label actually matched.
- **Cost**: 1.05s in-process, ~5s with the executable layer, 16s under `-race`.
- **Full module `-race`: rc=0, 64 packages, 0 FAIL. Golden 186 = 186, zero drift.** gofmt/vet clean.
- **Zero imports from `internal/goldencorpus`** (own canonical sort, own OOXML builder), so a rename there cannot produce a clean merge that does not compile.

## Two incidental fixes

**CI path filter.** `baseline.json` matches no `**.go` pattern, so a **baseline-only commit previously received zero checks** — exactly the commit shape that most needs them, since editing the baseline is how a quality regression gets blessed. Adds `internal/scorecorpus/testdata/**` to both path lists.

**`.gitignore`.** An unanchored `testing/` matched `docs/testing/`, the only directory of that name in the tree, so the pattern protected nothing and silently ignored committed documentation. A new file there was invisible to `git add` and would have been dropped from this PR without a word.

## Docs

[docs/testing/SCORE_CORPUS.md](docs/testing/SCORE_CORPUS.md), linked from TEST_PLAN.md as dimension 3b — including the traps that produced confidently wrong numbers while building this: a `.csv` extension alone is worth **+20 confidence**; my first container fixture was **vacuous** and passed the #250 mutation; and residue must be measured against the original or a neighbouring `A Morgan` reads as a leaked email.


sectoramen:
## Timing measurement for the PERSON_NAME pattern change

I reported TP/FP for this change but **not** timing when I opened the PR. Measured now, because the fix adds two regexes to a per-line pattern set and that has a cost worth stating rather than discovering later.

### Isolated A/B, interleaved on one binary (64KB single line, 5×20 iterations)

| | median | spread |
|---|---|---|
| without the two patterns | **61.3 ms** | 60.8 – 64.1 |
| with them | **75.1 ms** | 73.6 – 79.4 |

**+13.8 ms, +23%.** The ranges do not overlap, so this is a real signal, not noise — I checked the same-binary noise floor first (±10% at n=4000, 432–474 ms) and the earlier cross-build numbers were inside it. Two patterns added to a 17-pattern set is +12% patterns for +23% time on the worst-case shape.

### Still linear — no O(n²)

Per-doubling ratios, with the patterns present:

```
multi-line   n=500 63ms | 1000 117ms 1.84x | 2000 240ms 2.06x | 4000 478ms 1.99x
single line  64KB 83ms | 128KB 174ms 2.09x | 256KB 331ms 1.90x | 512KB 562ms 1.70x
```

Linear is ~2.0× per doubling, quadratic ~4.0×. Both stay at ~2.0× at every step, including the largest. The `personname` complexity guard ([complexity_guard_test.go:213](https://github.com/awslabs/ferret-scan/blob/main/internal/goldencorpus/complexity_guard_test.go#L213)) passes uncached against its 15 s sub-quadratic threshold, with the measured run at 0.07 s.

### Realistic input: ~+8%, not +23%

The 23% is a worst case — a line packed with nothing but hyphenated names. On a 4,000-line / 335 KB document where ~6% of lines carry a name, end-to-end (5 runs each):

```
before   0.18s 0.18s 0.19s 0.18s
after    0.19s 0.20s 0.20s 0.20s
```

### Assessment

The cost is real and I would rather it were zero, but it buys closing a cleartext leak — three name fragments (`Anne-`, `-Parker`, `Jean-`) previously survived in files the tool reported as redacted. A one-off ~14 ms on a pathological 64 KB line, and ~10 ms on a realistic 335 KB document, is the right trade for that. It is also linear, so it does not degrade with scale.

If the worst case ever matters, the fix is mechanical: the two new patterns are strictly wider than `compound_first_name` and `hyphenated_last_name`, so all four could collapse into one optional-hyphen pattern. I did not do that here because it would change the behaviour of the two existing patterns, and this PR is already carrying a corpus.

**TP/FP for the same change, unchanged from the PR body:** PERSON_NAME 1 TP / 2 FP → **4 TP / 0 FP** (the overlapping partials *were* the false positives), corpus total 180 TP / 0 FN / 47 FP, golden 186 = 186, full module `-race` rc=0.

rustic (approved):

PR #277, opened 2026-08-06
fix(cli): report files that were never examined, and stop them reading as clean

sectoramen:
A file the tool could not read must never be indistinguishable from a file that was read and found clean. Four separate defects made it exactly that.

## 1. The summary said `0 skipped` for files it never opened

The skip-classification loop does `continue` for an unreadable file **before** `skippedFiles++`, so it landed in no counter at all. Measured on a 7-file directory with 2 unreadable and 4 unparseable files:

```
Files: 2 processed, 0 skipped
```

Five files vanished from the accounting, and `2 processed` included a file that **failed**. The comment above that loop already documented the intent to separate unreadable from unsupported — only the count was dropped.

Now `Files: 2 scanned, 5 NOT examined`, which reconciles (2 + 5 = 7). "scanned", not "processed", because a failure is not a completed scan. `ScanStats` gains `FilesNotExamined` (omitempty, so json/yaml are unchanged when zero and no golden moves).

## 2. Four WARNING banners became one report

Before, each failure mode printed its own block, each line repeating the path then appending a raw Go error:

```
WARNING: scan incomplete — 1 of 1 file(s) could not be opened, so they were not scanned at all; any sensitive data they contain was NOT detected:
  /tmp/x/noperm.txt: Unreadable: open /tmp/x/noperm.txt: permission denied
```

The path appears **three times**, `Unreadable:` is internal vocabulary, and the consequence sits in a subordinate clause. Now:

```
═══ Scan Summary ══════════════════════════════════════════════════════════
Files: 2 scanned, 5 NOT examined | Findings: 1 (0 high, 1 medium, 0 low)
───────────────────────────────────────────────────────────────────────────
NOT EXAMINED: 5 of 7 files — contents were never read, so findings may be missing
  cannot read (2)
    /tmp/mixdir/hr_notes.txt   permission denied
  cannot parse (3)
    /tmp/mixdir/budget.xlsx    zip signature present but the archive could not be opened
  Add --fail-on-incomplete to make this a non-zero exit (3).
═══════════════════════════════════════════════════════════════════════════
```

It renders **inside** the block (`FormatterOptions.NotExaminedFooter`) rather than being printed separately: on stderr while the summary is on stdout, a piped report ended with a frame whose closing rule went to the terminal.

Rules are sized to content (floor 80, ceiling 120). The real-world summary line `Files: 276 scanned, 24 NOT examined | Findings: 4631 (...)` is **86 characters** and visibly overhung its 80-character frame. Width counts **runes** — the em-dash and middle dot are 3 and 2 bytes, so `len()` over-ran by several characters per glyph.

Detail comes from the **bytes, not the extension**: `describeUnparseable` reads the signature, so *"zip signature present but the archive could not be opened"* is an observation. Inferring *"corrupt Office document"* from `.xlsx` would be a guess, and a wrong guess sends the operator looking in the wrong place.

Over 8 files the list collapses to per-cause counts — 40 broken `.docx` produced **54 lines** of stderr, 40 identical apart from the filename, burying both the findings table and the one line that mattered.

## 3. Pre-commit exited 0 with ZERO bytes of output

Measured: `PRE_COMMIT=1` on a directory whose only file was `chmod 000` produced **no stdout, no stderr, rc=0** — byte-identical to a clean pass, on a file that may be full of PII. The commit is allowed and nothing says why. That is the [#193](https://github.com/awslabs/ferret-scan/pull/193) shape.

Pre-commit now emits one terse line (no frame; it is read mid-commit):

```
ferret-scan: 1 file NOT examined (contents unreadable) — findings may be missing. Re-run without --pre-commit for detail.
```

## 4. Unreadable and corrupt files exited DIFFERENTLY

| scenario | before (pre-commit) | after |
|---|---|---|
| clean | 0 | 0 |
| **unreadable, no findings** | **0**, silent | 0 + warning; **3** with `--fail-on-incomplete` |
| **corrupt, no findings** | **2** | 0 + warning; **3** with `--fail-on-incomplete` |
| MEDIUM finding | 0 · 1 at `EXIT_ON=medium` | unchanged |

Cause: `hasErrors` was `len(supportedFiles) - processedFiles`, and an unreadable file never *enters* `supportedFiles` — the same counter-skipping shape as defect 1. Nobody chose the split.

Exit code 2 is documented as *"the tool failed"*, so it is reclaimed for that: `hasErrors` is now fed by the `Parallel processing failed` branch, which previously only printed and never affected the exit code.

**Deliberately NOT changed:** unexamined files still need `--fail-on-incomplete` to be non-zero. Making that the default would break commits that pass today, and the first time it fires on a broken `.docx` mid-hotfix the hook gets disabled — and a disabled hook catches nothing. It is loud at rc=0 now, so the default can be flipped later on evidence rather than assertion.

## Also: the duplicate `--limit` note

The out-of-band *"output limited to N of M findings"* line was emitted for every format *"since it costs nothing to be consistent"* — my own reasoning in [#274](https://github.com/awslabs/ferret-scan/pull/274). It does cost something: measured on a 3-finding scan at `--limit 1`, **six of seven formats already disclose truncation in band** (text `and N more findings`, json/yaml `truncated:true`, junit `system-out`, sarif run properties, gitlab-sast scan block), so the note repeated the report's own disclosure immediately after it. Now scoped to **csv**, the one format with nowhere to put metadata, and worded to say why.

## Testing

- **6 output modes verified end to end:** `--limit 0`, `--limit N`, `--debug`, `--fail-on-incomplete`, all 7 formats, and a clean run.
- `--debug` no longer advises *"re-run with --debug"* when debug is already on.
- **`No matches found.` carried NO footer**, because four early returns bypass the summary — the same text a genuinely clean scan prints, on files never read. Fixed and covered.
- `TestSummaryFrameShape` pins double/single/double; a swap fails it. `TestSummaryRulesReachTheirContent` uses the real 4631-finding numbers.
- **3 new exit-code tests drive the real binary**, with a clean-scan control so a change marking every file unexamined cannot pass.
- **Non-vacuity via COMPILING mutations:** restoring the old `hasErrors` makes unreadable/corrupt disagree again (3 vs 2); removing the pre-commit line restores the zero-byte silence. Both fail with the intended message. A first attempt produced a build failure, which proves nothing, and was redone.
- Full module `-race`: **rc=0, 64 packages, 0 FAIL.** Golden **186 = 186**, zero drift. gofmt/vet clean, `GOOS=windows go vet` clean.
- **Zero file overlap with [#276](https://github.com/awslabs/ferret-scan/pull/276)**, so the two merge in either order.

## Known gap, next PR

Machine-readable formats carry only the **count** (json/yaml, free via `ScanStats`) and not the file list; **csv/junit/sarif/gitlab-sast carry nothing at all**. Those are the CI-facing formats and deserve their own diff — a SARIF consumer today sees findings with no indication that files went unexamined.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -1705,6 +1705,13 @@ func main() {
 	// the user no reason to look again at a file that was never scanned.
 	var supportedFiles []string
 	var unreadableFiles []string
+	unreadableCount := 0
+
+	// scanMalfunction records that the SCANNER failed, not that an input was bad.
+	// Exit code 2 means "the tool did not work"; a file it could not read is a
+	// coverage gap (code 3), which is the distinction the old failedFiles counter
+	// blurred.
+	scanMalfunction := false
 	for _, filePath := range filesToProcess {
 		canProcess, reason := fileRouter.CanProcessFile(filePath, finalConfig.enablePreprocessors)
 		if canProcess {
@@ -1713,6 +1720,12 @@ func main() {
 		}
 		if strings.HasPrefix(reason, router.ReasonUnreadable) {
 			unreadableFiles = append(unreadableFiles, fmt.Sprintf("%s: %s", filePath, reason))
+			// Counted as unreadable AND as not-processed. The `continue` here used to
+			// skip the increment below, so an unreadable file landed in NO counter at
+			// all: two permission-denied files produced "0 skipped" in the summary
+			// while the warning above said two files were never opened. Separating the
+			// two categories was right; dropping the count was not.
+			unreadableCount++
 			continue
 		}
 		skippedFiles++
@@ -1800,31 +1813,34 @@ func main() {
 					stats.ProcessedFiles, stats.TotalMatches, stats.WorkerCount, stats.TotalDuration.Milliseconds())
 			}
 		} else {
+			// A genuine tool malfunction, as opposed to an input the tool could not
+			// read. This is what exit code 2 is for, so record it rather than only
+			// printing it.
 			fmt.Fprintf(os.Stderr, "Parallel processing failed: %v\n", err)
+			scanMalfunction = true
 		}
 	}
 
 	elapsed := time.Since(progressStart)
 	finalSkippedCount := totalSkipped + skippedFiles
 
-	// Provide clearer reporting (suppress in pre-commit mode)
-	totalAttempted := len(supportedFiles)
-	failedFiles := totalAttempted - processedFiles
-
 	if !shouldSuppressProgressOutput(finalConfig, precommitConfig, isInteractive) {
-		if finalSkippedCount > 0 || failedFiles > 0 {
-			fmt.Fprintf(os.Stderr, "Scan complete: %d files processed successfully", processedFiles)
-			if failedFiles > 0 {
-				fmt.Fprintf(os.Stderr, ", %d files had no results", failedFiles)
-			}
-			if finalSkippedCount > 0 {
-				fmt.Fprintf(os.Stderr, ", %d files skipped (%d unsupported types)", finalSkippedCount, skippedFiles)
-			}
-			fmt.Fprintf(os.Stderr, " in %s\n", elapsed.Round(time.Millisecond))
-		} else {
-			fmt.Fprintf(os.Stderr, "Scan complete: %d files processed in %s\n",
-				processedFiles, elapsed.Round(time.Millisecond))
+		// Report only what this line uniquely owns: how many files were scanned, how
+		// many were an unsupported type the user did not expect a result for, and how
+		// long it took. Files that FAILED to parse are deliberately NOT mentioned here
+		// as "had no results" — that phrasing conflated a clean scan (findings=0) with
+		// a file that was never read, and the not-examined report below now owns that
+		// category in full. Reporting it in both places produced two different numbers
+		// for the same files (a 23-vs-24 split a reader had to reconcile).
+		fmt.Fprintf(os.Stderr, "Scan complete: %d files scanned", processedFiles)
+		if finalSkippedCount > 0 {
+			fmt.Fprintf(os.Stderr, ", %d skipped (unsupported type)", finalSkippedCount)
 		}
+		// Trailing blank line so the progress block (spinner, bar, this line) is
+		// visually separated from the findings table that follows it. Without it the
+		// table header butts directly against the progress output and the two read as
+		// one wall of text.
+		fmt.Fprintf(os.Stderr, " in %s\n\n", elapsed.Round(time.Millisecond))
 	}
 
 	// Incomplete-coverage warning (v2 Phase 4): a file whose validator coverage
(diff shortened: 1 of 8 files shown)


sectoramen:
## Schema verification for the follow-up (machine-format reporting)

Checked the authoritative schemas before building the next PR, since the maintainer flagged that a live GitLab integration must not break. Recording the results here because they change the design.

### GitLab SAST — extra keys are safe, but there is a better field

The [GitLab SAST schema](https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/raw/master/dist/sast-report-format.json) sets **`"additionalProperties": true`** at the top level and imposes **no `additionalProperties` restriction on the `scan` object**. Unknown keys cannot cause a rejection.

That is also retroactive evidence for this PR's predecessor: #274 already added `truncated` and `total_vulnerabilities` to the `scan` block, which is live on `main` today. Working GitLab reports are proof extra keys are tolerated.

**But the schema defines a purpose-built field I would otherwise have missed:** `scan.messages[]` — *"Communication intended for the initiator of a scan"* — with `level` ∈ `{info, warn, fatal}`. GitLab's own guidance says use `warn` for *"a potentially recoverable problem, or a partial error"*, which is exactly "N files could not be read". A standard field **renders in the GitLab UI**; a custom key is silently ignored.

So the follow-up will use `scan.messages[]`, not a bespoke `unscanned` key.

`scan.status` stays untouched — the schema pins it to `success`/`failure`, and a scan with unreadable files did succeed. (Already reasoned out in #274's code comments.)

### SARIF — the OPPOSITE, and it constrains the design

The [SARIF 2.1.0 schema](https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/schemas/sarif-schema-2.1.0.json) sets **`"additionalProperties": false`** on both the top-level object and the `run` object. Only `propertyBag` allows extras.

- This PR's `run.properties` addition (from #274) is **valid** — `properties` is a declared `run` field and is a propertyBag.
- A new invented key on `run` would make the document **schema-invalid**. The standard location is `run.invocations[].toolExecutionNotifications[]`, a `notification` requiring only `message`, with `level` ∈ `{none, note, warning, error}` and optional `locations` pointing at the affected artifacts.
- We currently emit **no** `invocations` array at all, so that has to be added.

### Baseline captured

Current `gitlab-sast` output uses exactly the six schema-required `scan` keys (`analyzer`, `end_time`, `scanner`, `start_time`, `status`, `type`), and on a directory with 5 unexamined files reports 1 vulnerability with **zero indication** anything went unread. That is the gap the follow-up closes, and the before/after diff will be in its PR body.

### Sequencing note

The follow-up needs a new `FormatterOptions` field and wiring at the scan-completion call site — the **same anchors this PR edits** (`internal/formatters/formatter.go` @@28, `cmd/main.go` @@1926). Building it now would guarantee an add/add conflict, so it waits until this PR merges rather than racing it.

Also worth adding then: validating a generated report against the live GitLab schema in CI, so "extra keys are safe" becomes a check that runs rather than a claim in a PR body.

rustic (approved):

PR #285, opened 2026-08-11
feat(scan): expose the match's surrounding text on Finding

rustic:
## What

Adds `ContextBefore`, `ContextAfter` and `FullLine` to `pkg/scan.Finding` and populates them in `mapResult`, so the text around a match is reachable from outside the module.

Requested by the **ferret-scan-mobile** team (iOS + Android). `pkg/scan` already covers everything else their facade needs; match context was the last thing only reachable via `internal/detector`, and therefore the last reason their build has to compile inside a checkout of this repo instead of importing it as a normal Go module. This package's own doc comment names the gomobile facade as an intended consumer.

Their use case: the surrounding line is what distinguishes a live value from a form example, and it's the prompt they give an on-device model for a second opinion on a finding the engine already flagged.

```
Corporate card ⟦value⟧ expires soon        → probably real
Example: ⟦value⟧ (do not use)              → probably a placeholder
```

## The change

Three fields, three assignments, one mapping point — `mapResult` is the only place internal matches become public findings, so `ScanText` and `ScanFile` are both covered by one edit. No new detection logic, no validator touched, no output format touched. The golden corpus is unchanged and needed no regeneration.

Two things measured while implementing it changed *how* the fields are exposed. Both are the reason this is a little larger than the 3+3 line diff that was requested.

### 1. The raw values can be invalid UTF-8

Validators cut their context windows with byte arithmetic — `line[matchStart-30 : matchStart]` in creditcard, ±50 in ssn — which slices through a multi-byte rune whenever the neighbouring text isn't ASCII. Forwarding it verbatim, measured on CJK text either side of a test card number:

| field | value | valid UTF-8 |
|---|---|---|
| `ContextBefore` | `"\x8f\xaf以外流給任何人知道 "` | ❌ |
| `ContextAfter` | `" 這是機密資料絕對不\xe5\x8f"` | ❌ |

Harmless internally (the value only feeds keyword matching), but a *public* field gets converted to a Java/Swift `String` by gomobile and to U+FFFD by `json.Marshal`. The fragment is trimmed at the mapping point, bounded to the ≤3 bytes a cut can leave — a boundary repair, not a sanitizer, so genuinely mis-encoded content further into the line still reaches the caller intact.

### 2. `Finding` has no JSON tags, so new fields would silently widen an existing sink

Nothing in this repo marshals `Finding`, but third-party callers are why the type is public, and Go serializes untagged exported fields by name. Adding three raw-content fields would have appended the surrounding line to whatever an existing caller already writes to a file or ships to a log — no code change on their side, nothing in a diff to notice.

The three fields carry `json:"-"`. The previously marshaled shape is byte-identical; a caller who wants context in JSON has to opt in. **Flagging this as the one judgment call a maintainer might want to reverse** — it's a one-word change if you'd rather they serialize by default.

The doc comment states only the contract the fields can actually keep: single line, window width owned by the validator and explicitly *not* part of the compatibility surface (30–50 bytes today, will move as validators are tuned), same sensitivity as `Text`, may be empty.

## What I deliberately left out, and why

**Three validators record no context at all**, so their findings report blank context however much text surrounds them: `SECRETS`, `PERSON_NAME`, `CLOUD_RESOURCES`. `Match.Context` is populated per-validator with no central backfill. Verified on both entry points:

```
ctx-OK    creditcard/VISA          before="Corporate card "  after=" expires soon"
ctx-OK    ssn/SSN, email, phone, dob, passport, vin, bank_account, address, ...
ctx-NONE  secrets/AWS_ACCESS_KEY   before="" after="" fullline=""
ctx-NONE  PERSON_NAME/PERSON_NAME  before="" after="" fullline=""
ctx-NONE  cloud_resources/AWS_ARN  before="" after="" fullline=""
```

SECRETS is the worst of the three to lose — "real key or documentation example" is precisely the question the field exists to answer. (The fixture in the test is `AKIAIOSFODNN7EXAMPLE`, the key from AWS's own docs.)

I intended to close this gap in the same PR. **It isn't the two-line change it looks like.** `suppressions.findingHashVersion` reads `Context.FullLine`, `BeforeText` and `AfterText` as part of a finding's *identity*:

```
AWS_ACCESS_KEY, no context (today)   ed3217d2a07c46a9...
AWS_ACCESS_KEY, context attached     54ba6fe7b2da4db6...
```

Suppression matching is purely hash-indexed (`rulesByHash`) with no structural fallback, and the version candidates cover only the confidence axis. So attaching context would make **every existing operator rule for a SECRETS, PERSON_NAME or CLOUD_RESOURCES finding stop matching** — turning reviewed-and-accepted findings back into noise, and in a pre-commit gate back into a block. That's the exact failure mode the comment on `findingHashVersion` exists to prevent ("an operator's suppression file is a record of decisions they already made, not a cache we may invalidate").

Closing it properly needs a hash-compatibility version, which loosens suppression identity for *every* finding — a maintainer call, and a security-relevant one, not a rider on an API addition. So instead:

- the gap is documented on the field, with empty defined as **"not recorded"**, never "the match stood alone" (a caller reading it the second way draws a conclusion about a secret that the data doesn't support);
- `TestValidatorsWithoutContextAreDocumented` fails if a validator starts populating context, and points the next person at the hash question before they ship it.

Tracked in #286, which lays out four options for closing it with the measurements behind each. Happy to implement whichever you pick — it needs a decision from you first.

## Also noticed (not addressed here)

1. **`Finding.SuppressedBy` is dead and silently lies.** `mapResult` never sets it, and it can't: `core.ScanFile` returns `Matches: unsuppressed`, so a suppressed finding never appears in `Findings` at all, and `SuppressedBy` lives on `detector.SuppressedMatch`, a different type. A caller checking `f.SuppressedBy != ""` always concludes nothing was suppressed. This is the same shape as the `SOCIAL_MEDIA` / `ValidCheckNames()` fix in the current changelog — a public field reporting a clean state it can't actually observe. Populating it means surfacing `SuppressedMatches` (an API addition); dropping it is breaking but honest. Left alone here since either direction is a separate decision; tracked in #287.
2. **`docs/upstream-asks.md` has two contradictory `## Status` tables** (lines 111–138) — ask #8 is `Done` in one and `Not started` in the other. Untouched to keep this diff focused.
3. **TM-11 applies to the consumer's design.** Worth passing back to the mobile team rather than fixing here: the threat model already documents that attacker-authored context is a suppression oracle. Feeding it to an LLM that decides keep-vs-discard rebuilds that oracle at the prompt layer, where `Example: ⟦value⟧ (do not use)` also becomes an injection surface. Their prompt needs the same "checksum-valid values keep a floor" discipline the engine adopted.

## Verification

- `go test ./...` clean (65 packages), `go vet ./...` clean, `make build` succeeds.
- Golden corpus unchanged, no regeneration — no detection, confidence or format change.
- Additive: existing callers' compiled behaviour *and* marshaled output are identical. No unkeyed `Finding` literals in the tree.
- Both non-obvious decisions are mutation-tested: removing the trim fails the UTF-8 test, removing the tags fails the JSON test.
- New tests: exposure on both entry points, single-line contract, valid UTF-8, trim helpers (table + bounded-not-sanitizing), JSON exclusion, and the documented-gap lock.

### DIFF
--- a/pkg/scan/context_test.go
+++ b/pkg/scan/context_test.go
@@ -1,0 +1,283 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package scan
+
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"unicode/utf8"
+)
+
+// findByValidator returns the first finding produced by the named validator.
+// Tests select on the validator rather than indexing Findings, because the
+// order of the slice is the order validators complete, not source order.
+func findByValidator(t *testing.T, findings []Finding, validator string) Finding {
+	t.Helper()
+	for _, f := range findings {
+		if f.Validator == validator {
+			return f
+		}
+	}
+	t.Fatalf("no finding from validator %q in %d findings", validator, len(findings))
+	return Finding{}
+}
+
+// TestContextExposedOnBothEntryPoints locks the context fields on the two public
+// scan paths. Both go through mapResult, so this is really asserting that the
+// mapping keeps the data rather than dropping it as it did before.
+func TestContextExposedOnBothEntryPoints(t *testing.T) {
+	const line = "Corporate card 4111-1111-1111-1111 expires soon\n"
+
+	path := filepath.Join(t.TempDir(), "ctx.txt")
+	if err := os.WriteFile(path, []byte(line), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	fileResult, err := ScanFile(context.Background(), path, FileOptions{Checks: []string{"CREDIT_CARD"}})
+	if err != nil {
+		t.Fatalf("ScanFile: %v", err)
+	}
+	textResult, err := ScanText(context.Background(), line, TextOptions{Checks: []string{"CREDIT_CARD"}})
+	if err != nil {
+		t.Fatalf("ScanText: %v", err)
+	}
+
+	for _, tc := range []struct {
+		entry    string
+		findings []Finding
+	}{
+		{"ScanFile", fileResult.Findings},
+		{"ScanText", textResult.Findings},
+	} {
+		t.Run(tc.entry, func(t *testing.T) {
+			f := findByValidator(t, tc.findings, "creditcard")
+
+			if !strings.Contains(f.ContextBefore, "Corporate card") {
+				t.Errorf("ContextBefore = %q, want it to contain the preceding text", f.ContextBefore)
+			}
+			if !strings.Contains(f.ContextAfter, "expires soon") {
+				t.Errorf("ContextAfter = %q, want it to contain the following text", f.ContextAfter)
+			}
+			if got, want := f.FullLine, strings.TrimSuffix(line, "\n"); got != want {
+				t.Errorf("FullLine = %q, want %q", got, want)
+			}
+
+			// The context brackets the match; it never restates it. A caller
+			// that concatenated before+text+after must not get the value twice.
+			if strings.Contains(f.ContextBefore, f.Text) {
+				t.Errorf("ContextBefore %q must exclude the matched value %q", f.ContextBefore, f.Text)
+			}
+			if strings.Contains(f.ContextAfter, f.Text) {
+				t.Errorf("ContextAfter %q must exclude the matched value %q", f.ContextAfter, f.Text)
+			}
+			if !strings.Contains(f.FullLine, f.Text) {
+				t.Errorf("FullLine %q must contain the matched value %q", f.FullLine, f.Text)
+			}
+			if got := f.ContextBefore + f.Text + f.ContextAfter; !strings.Contains(f.FullLine, got) {
+				t.Errorf("before+text+after = %q is not a span of FullLine %q", got, f.FullLine)
+			}
+		})
+	}
+}
+
+// TestContextDoesNotSpanLines guards the documented single-line contract. A
+// caller building an LLM prompt from context needs to know the blast radius of
+// what it is about to send; "the neighbouring lines too" is a different and much
+// larger promise than the one this package makes.
+func TestContextDoesNotSpanLines(t *testing.T) {
+	body := strings.Join([]string{
+		"UNRELATED-PRECEDING-LINE",
+		"Corporate card 4111-1111-1111-1111 expires soon",
+		"UNRELATED-FOLLOWING-LINE",
+	}, "\n") + "\n"
+
+	r, err := ScanText(context.Background(), body, TextOptions{Checks: []string{"CREDIT_CARD"}})
+	if err != nil {
+		t.Fatal(err)
+	}
+	f := findByValidator(t, r.Findings, "creditcard")
+
+	for name, got := range map[string]string{
+		"ContextBefore": f.ContextBefore,
+		"ContextAfter":  f.ContextAfter,
+		"FullLine":      f.FullLine,
+	} {
+		if strings.Contains(got, "\n") {
+			t.Errorf("%s = %q contains a newline; context is documented as single-line", name, got)
+		}
+		for _, leaked := range []string{"PRECEDING", "FOLLOWING"} {
+			if strings.Contains(got, leaked) {
+				t.Errorf("%s = %q leaked text from a neighbouring line", name, got)
+			}
+		}
+	}
+	if f.LineNumber != 2 {
+		t.Errorf("LineNumber = %d, want 2", f.LineNumber)
+	}
+}
+
+// TestContextIsValidUTF8 covers the reason the mapping trims rune fragments.
+// Validators cut their context windows at byte offsets, so a run of multi-byte
+// runes before the match gets sliced mid-rune. Handing that to a gomobile
+// consumer means a mangled Java/Swift string, and to json.Marshal means a silent
+// U+FFFD substitution, so the public fields are repaired at the boundary.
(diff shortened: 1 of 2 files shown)


sectoramen:
Reviewed this by re-deriving the three non-obvious claims rather than reading them, and they all hold. Recording what I verified, then two notes on the parts you flagged as open.

**The invalid-UTF-8 premise reproduces exactly.** Driving `core.ScanContent` directly on the CJK fixture: `BeforeText = "\x8f\xaf以外流給任何人知道 "`, `AfterText = " 這是機密資料絕對不\xe5\x8f"`, `FullLine` valid. So the trim is load-bearing rather than defensive. Trimming at the mapping point rather than in the validators is the right call for a second reason worth stating: it leaves the internal value untouched, so the suppression hash does not move — an in-validator rune-rounding fix would have moved it for every file with multi-byte runes near a match.

**The suppression-hash deferral is justified.** `findingHashVersion` folds trimmed `FullLine` and `sha256(BeforeText+AfterText)` into the identity, `IsSuppressed` matches purely through `rulesByHash` with no structural fallback, and both candidates are context-bearing. Attaching context to the three contextless validators does invalidate every existing rule for those types.

**Both mutations fail as advertised** — I ran them. Reverting the trim fails `TestContextIsValidUTF8` with the exact invalid bytes; removing the tags fails `TestContextExcludedFromJSON`. `TestValidatorsWithoutContextAreDocumented` is also non-vacuous: all three validators do produce findings on its fixtures, so it is asserting emptiness rather than absence.

Gates on the branch: `go build`, `go vet`, `go test ./...` (64 packages), golden corpus unchanged, `TestValidatorComplexityIsSubQuadratic` + generators non-vacuity, `make score` (`recall_all=1.0000 recall_hm=0.9947 prec_hm=0.9310 whole_leak=0`), timing 0.87s in-process, integration, `-race`, and `gosec ./pkg/scan/...` clean.

---

**On `json:"-"` — keep it, but the strongest argument is not the one in the description.** "Adding three raw-content fields would have appended the surrounding line to whatever an existing caller already writes to a file" overstates the delta: the pre-existing marshaled shape already contains `"Text":"4111-1111-1111-1111"`, so that sink already carries the secret itself, and context adds prose around it rather than a new class of data. The argument that does hold on its own is shape stability, and it also settles the naming question if the decision is ever reversed: this project already has a wire format for exactly these three fields — `full_line` / `before_text` / `after_text` (`internal/formatters/shared/structures.go:211-213`). Tagging them `json:"-"` now and those names later is strictly better than inventing PascalCase keys, so the current state is the right default.

**One line worth adding to the sensitivity note: unlike every CLI surface, these fields have no reveal gate.** Measured across the formatters, context is emitted by text, json, yaml, sarif, junit and gitlab-sast, and never by csv — and in every case it is gated: sarif and gitlab-sast require `--show-match`, text/json/yaml/junit require `--verbose --show-match`, and `--verbose` alone leaks nothing. `SanitizeSuppressedMatches` clears all three specifically because they "hold the raw surrounding line". `pkg/scan` has no `ShowMatch` equivalent, so `ContextBefore` is always populated. That is consistent with `Text` being ungated and the doc comment does call the sensitivity out, so I read it as by-design rather than a defect — but a caller porting from the CLI's mental model will expect a gate, and one sentence saying there isn't one would close that gap.

---

Two things this made me look at that are not blockers for the PR, both filed separately:

- #290 — the contextless enumeration is missing a fourth shape. `metadata` builds `ContextInfo{FullLine: line}` at ~26 sites and never sets before/after, so METADATA findings always report blank before/after with a populated `FullLine` (measured on a real `.docx`: `before="" after="" full="Author: Michael Thompson"`). Worth naming before #286 lands, since the natural doc edit then is "all validators populate context", which would be false. The issue also flags that #286's per-type option would exclude metadata from the hash-compatibility fix, and metadata sits on the same cliff (`hashSensitiveData("")` returns `""` today).
- #289 — `SOCIAL_MEDIA_CLUSTER` borrows the primary sub-match's context, so `FullLine ⊇ Text` is not an invariant. `TestContextExposedOnBothEntryPoints` asserts it for creditcard, which is true, but a reader will take it as global. Chasing the downstream consumers turned up something worse than a doc problem, hence the separate issue — the doc-facing part is just that `FullLine is the whole line the match sits on` needs a hedge for synthesized matches.

Neither changes my read of this PR: the gap is real, the mapping point is the right place, and the two judgment calls are the right defaults.

sectoramen:
One more, and it is the only thing I would ask you to change in this PR rather than defer: **`TestContextIsValidUTF8` can go vacuous without failing.**

It relies on creditcard's 30-byte window landing mid-rune inside a run of 3-byte CJK characters. The field doc immediately above it says the window width "is not part of this package's compatibility surface and will move as validators are tuned" — and if it moves to a multiple of 3, the cut lands on a rune boundary, the trim becomes a no-op, and the test passes while asserting nothing about the repair. Its guard catches empty context, not boundary-aligned context.

The blast radius is small because `TestTrimRuneFragments` locks the trim logic width-independently, so what is lost is only the end-to-end coverage — but that is the half that proves the mapping actually calls the trim, which is exactly what your first mutation demonstrates. A width-independent assertion would keep it honest, e.g. assert the mapped value differs from the raw internal one for this fixture, or that the fixture's untrimmed window is invalid UTF-8 before checking the mapped one is valid. Given `score-mutation-check` and `TestComplexityGeneratorsAreNonVacuous` exist for the same reason elsewhere in the tree, it seemed worth raising rather than leaving to drift.

Separately, filed #293 from something this PR led me to: `pkg/scan` inherits a config file discovered from the process working directory and has no way to opt out, which is why identical calls can return different findings. Not caused by this change, but it lands on the same public surface and on the same gomobile consumer.

Issue #294, opened 2026-08-12
A corrupt or truncated .pdf is counted as scanned and reported clean: pdf_metadata succeeds so the text extractor's error reaches no disclosure channel (exit 0 even with --fail-on-incomplete)


sectoramen:
## Summary

A `.pdf` whose text extraction fails is reported as **scanned and clean**. Because the `pdf_metadata` preprocessor succeeds on the same bytes — it succeeds even when the file has no `%PDF` header at all — the router stamps `Success: true`, and the text extractor's error is discarded before it can reach any coverage-disclosure channel.

The same corruption in a `.docx` or `.jpg` correctly prints `NOT FULLY EXAMINED ... cannot parse` and exits 3 under `--fail-on-incomplete`.

This is the residue of the family closed by #257 and #264: both trigger on "every preprocessor failed" or on a text sniff, and **neither condition holds here** — one preprocessor genuinely succeeded.

## Reproduction

Build a valid PDF carrying an SSN in a FlateDecode stream, then truncate it at the xref:

```python
import zlib
content = b"BT /F1 12 Tf 72 700 Td (Employee SSN: 536-90-4271) Tj ET\n"
comp = zlib.compress(content)
objs = [b"1 0 obj\n<< /Type /Catalog /Pages 2 0 R >>\nendobj\n",
        b"2 0 obj\n<< /Type /Pages /Kids [3 0 R] /Count 1 >>\nendobj\n",
        b"3 0 obj\n<< /Type /Page /Parent 2 0 R /MediaBox [0 0 612 792] /Contents 4 0 R /Resources << /Font << /F1 5 0 R >> >> >>\nendobj\n",
        b"4 0 obj\n<< /Length %d /Filter /FlateDecode >>\nstream\n" % len(comp) + comp + b"\nendstream\nendobj\n",
        b"5 0 obj\n<< /Type /Font /Subtype /Type1 /BaseFont /Helvetica >>\nendobj\n"]
out, offs = bytearray(b"%PDF-1.4\n"), []
for o in objs: offs.append(len(out)); out += o
x = len(out)
out += b"xref\n0 %d\n" % (len(objs)+1) + b"0000000000 65535 f \n" + b"".join(b"%010d 00000 n \n" % f for f in offs)
out += b"trailer\n<< /Size %d /Root 1 0 R >>\nstartxref\n%d\n%%%%EOF\n" % (len(objs)+1, x)
open("valid.pdf","wb").write(bytes(out))
[...]

PR #311, opened 2026-08-14
fix(redact): redact files embedded inside documents, and fail loudly when we cannot

### LINKED ISSUE #305: A finding inside an embedded part is reported but never redacted: exit 0, no warning, and the cleartext value stays in the "redacted" copy
## Summary

A finding inside an embedded part of a container is **reported but never redacted**, and the tool exits 0 with no warning — so the "redacted" copy still holds the value in cleartext while the report says the scan succeeded.

Split out of #304 (which closes the *detection* half of #297) because the gap is older and wider than that change.

## Measured

With #304 applied, an SSN in `word/embeddings/oleObject1.docx`:

```console
$ ferret-scan --config /dev/null --checks SSN --limit 0 --file outer.docx \
    --enable-redaction --redaction-output-dir ./out
[HIGH  ] ssn  SSN  100.00% line 10  [HIDDEN]  outer.docx
Files: 1 scanned | Findings: 1 (1 high, 0 medium, 0 low)
$ echo $?
0
```

Reading the embedded part **inside** the redacted output (grep on the container returns nothing either way, since the bytes are compressed):

```
SSN still cleartext in the redacted inner document: TRUE
```

## Why it is not specific to #304

No redactor rewrites an embedded part:

```console
$ grep -rln 'embeddings' --include='*.go' internal/redactors/
# 0 files
```

`internal/redactors/index.go:71` defines an `EmbeddedMediaRef` type, but nothing consumes it to rewrite content. Embedded **images** and **audio** have been routed and scanned for some time, so any finding from those was already unredactable. #304 adds embedded *documents* to that existing class — it does not create it.

## Why it matters

`internal/core/redact.go:118-130` already carries a guard for exactly this failure class, with a comment that describes it precisely:

> *"Returning success here would hand the caller a path its own doc comment calls safe to share while the sensitive values sit in it verbatim — and `RedactionCount` would read 0, indistinguishable from a genuinely clean file."*

That guard does **not** fire here: `stats.UnredactedFiles` is empty, because the worker pool *did* write an output file — it just wrote one with the embedded part copied through unchanged.

So the operator gets a report naming the finding, a redaction run that exits 0, and a file that still contains the value. That combination invites more confidence than a silent miss does.

## Two ways to close it

1. **Wire the existing guard** (smaller). Treat a finding whose source is an embedded part as unredacted unless the redactor rewrote that part, so `UnredactedFiles` fires and the run fails loudly. Does not recover the value, but makes the tool stop claiming success.
2. **Descend in the redactor** (complete). Have the Office redactor open the embedded part, redact it, and re-pack it — mirroring what the extractor now does on the read side. This is the "Office redactor" item already on the internal roadmap.

(1) is the honest minimum and could land on its own; (2) is the actual fix.

## Note for whoever picks this up

Verifying either one requires reading the part **inside** the zip (`word/embeddings/…` → `word/document.xml`). Grepping a redacted `.docx` for cleartext searches *compressed* bytes and returns zero — indistinguishable from a clean pass. `docs/testing/TEST_PLAN.md` calls this trap out explicitly.

sectoramen:
Closes #305.

Only reported findings reach the redactor, so a value the redactor cannot reach is a leak that **reports as success**. The read side already descends into embedded parts, so these values were FOUND. The write side did not, so they were shipped in cleartext inside a file sitting in a directory named `redacted`, at exit 0, with nothing on stderr.

## The leak, measured on the parent commit

Each containers OWN value was correctly redacted in every row, which is what made this invisible.

| fixture | before | after |
|---|---|---|
| `.docx` embedded in a `.docx` | wrote file, **SSN in cleartext** | wrote file, **0 leaks** |
| `.jpg` EXIF embedded in a `.docx` | wrote file, **SSN in cleartext** | wrote file, **0 leaks** |
| both children in one document | wrote file, **2 values in cleartext** | wrote file, **0 leaks** |
| 3-level nest (docx in docx in docx) | wrote file, **SSN in cleartext** | wrote file, **0 leaks** |
| `.wav` ICMT comment embedded | wrote file, **SSN in cleartext** | **no file + warning** |
| embedded `.pdf` | **unscanned**, value shipped | **detected**, no file + warning |
| embedded `.svg` | **unscanned**, value shipped | **detected**, no file + warning |
| unrelated clean photo alongside a body SSN | untouched | untouched (byte-identical) |
| clean document | no file | no file |

Every row verified by reading the bytes actually written and **descending into each nested archive member** — grepping a `.docx` searches compressed bytes and finds nothing, which is indistinguishable from a clean file and is the most common way a leak here gets certified as fixed.

## What changed

**Give the write side the recursion the read side has.** Each embedded part is dispatched to whichever registered redactor claims it, so embedded OOXML, legacy OLE and image metadata are covered by one loop, each by the code that already knows its format.

An earlier attempt taught the Office redactor to do zip surgery on nested OOXML parts. It fixed the embedded-`.docx` case and **could not fix the embedded-image case even in principle**, because removing an SSN from EXIF is not a string replacement in a zip member. That approach is deleted.

**Admission is derived from capability, not a hand-maintained table.** The table excluded **488 of 2,520 embedded parts (19%)** across 334 real Office files: 411 `.svg` (plain XML text that yields an SSN finding standalone), plus `.emf`, `.wdp`, and the `.bin` form Word uses for embedded OLE objects. Every part under `media/` or `embeddings/` is now offered to `RouterInterface.CanProcessFile`, so embedded coverage tracks top-level coverage and cannot silently fall behind again.

**Fail loudly rather than ship.** A container is refused when an embedded part still holds a reported value: no output file, and the existing `redaction incomplete ... the original values remain in cleartext` channel names the part and the reason. That converts three silent leaks into loud failures. The same policy already applied one level up, so nesting no longer changes the guarantee.

The childs self-report is **not trusted** — the bytes about to be written are re-scanned for every reported value, because a `RedactionCount` of 1 has previously been reported on an output byte-identical to its input.

**Do not touch clean parts.** Every redactor here is lossy (the image redactor decodes, re-encodes and strips all metadata), so a part is dispatched only when it can be shown to hold a reported value. Caught during review: an unrelated 706-byte photo in a document whose *body* held an SSN came back re-encoded to 664 bytes with its caption removed. It is now byte-identical. PDF is the deliberate exception and is always dispatched, because its text lives in FlateDecode streams so absence cannot be established by a byte scan.

**Bounds.** `MaxDepth` and `ErrTooDeep` move to a new leaf package `internal/embedded` shared by both halves, so the read and write bounds are one value rather than two that can drift — a shallower write side reports findings from a level it will not rewrite; a deeper one redacts where nothing scanned. The temp-file extension comes from a character-class allowlist, never from the producer-controlled entry name (BSC1), and the residue scan inflates archive members.

## Testing

Per `docs/testing/TEST_PLAN.md`:

- `go build ./...`, `go vet ./...`, `gofmt -l` — clean.
- `go test ./... -count=1` — **66 packages, 0 failures**, with #310 merged (union tested, not just conflict-checked).
- Golden corpus `-count=1` — pass, **zero goldens moved**.
- `make score` — **unchanged**: `recall_all=1.0000 recall_hm=0.9947 prec_hm=0.9310 whole_leak=0`.
- **Non-vacuity**: the six new end-to-end tests FAIL on the parent commit with the exact leak paths, including the 3-level nest.
- **Mutation**: nine mutations applied to the new guards (depth bound, allowlist, fail-closed, post-dispatch verification, residue gate, zip descent, temp cleanup, opaque set, child collection). **Every one compiled and every one was caught.**
- **Complexity** — redaction scaling is linear, binaries built outside `/tmp`:

| case | N=25 | 50 | 100 | 200 | per doubling |
|---|---|---|---|---|---|
| every part clean (skipped) | 0.130s | 0.163s | 0.181s | 0.344s | 1.1–1.9x |
| every part dispatched (worst case) | 0.126s | 0.210s | 0.406s | 0.819s | 1.66–2.02x |

  2x per doubling is linear. The clean case matches the parent almost exactly (0.344s vs 0.333s at N=200), so the residue gate keeps ordinary documents free; the worst case costs ~2–2.7x, which is the price of actually redacting parts previously skipped entirely.

## Behaviour changes worth calling out

- A document whose embedded part cannot be redacted now produces **no output file** instead of a silently incomplete one. Exit code is unchanged (0 with a stderr warning), matching the existing standalone behaviour — changing that is a separate decision.
- An embedded `.pdf` and an embedded `.svg` are now **scanned**, which is new detection; neither has a redactor, so such documents fail loudly.
- Corpus figures are aggregate counts only; every value shown is from a synthetic fixture.

## Follow-ups found while here, not in this PR

- A malformed WAV (odd-length chunk, unpadded) reports `No matches found` with no disclosure — same class as #294/#301.
- Redactor coverage lags preprocessor coverage at the **top** level too: the score corpus already lists `.tsv`, `.html` and `.sql` as scanned-but-unredactable, and #310 widened that set.

### DIFF
--- a/internal/core/redact.go
+++ b/internal/core/redact.go
@@ -204,10 +204,12 @@ func NewDefaultRedactionManager(outputDir string, strategy redactors.RedactionSt
 			FailureHandling:         redactors.FailureHandlingGraceful,
 		})
 
+	officeRedactor := office.NewOfficeRedactor(outputManager, observer)
+
 	for _, r := range []redactors.Redactor{
 		plaintext.NewPlainTextRedactor(outputManager, observer),
 		pdf.NewPDFRedactor(outputManager, observer),
-		office.NewOfficeRedactor(outputManager, observer),
+		officeRedactor,
 		legacyole.NewLegacyOLERedactor(outputManager, observer),
 		image.NewImageMetadataRedactor(outputManager, observer),
 	} {
@@ -215,5 +217,19 @@ func NewDefaultRedactionManager(outputDir string, strategy redactors.RedactionSt
 			return nil, nil, fmt.Errorf("failed to register redactor %s: %w", r.GetName(), err)
 		}
 	}
+
+	// Let the Office redactor descend into files embedded inside a document.
+	//
+	// Injected AFTER registration because the dispatcher is the manager that owns
+	// every redactor, and an embedded part is routed by type exactly as a top-level
+	// file is: an embedded .docx to this same redactor, a .jpg to the image
+	// redactor, a .doc to the legacy OLE one. Without this call the Office redactor
+	// still runs, but reports any embedded part holding a reported value as
+	// unredacted rather than silently shipping it.
+	//
+	// The manager bounds the recursion (embedded.MaxDepth), so registering the
+	// Office redactor with a dispatcher that can route back to it is not unbounded.
+	officeRedactor.SetEmbeddedRedactor(manager)
+
 	return manager, outputManager, nil
 }
--- a/internal/core/redact_embedded_test.go
+++ b/internal/core/redact_embedded_test.go
@@ -1,0 +1,648 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package core
+
+import (
+	"archive/zip"
+	"bytes"
+	"encoding/binary"
+	"image"
+	"image/color"
+	"image/jpeg"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/awslabs/ferret-scan/v2/internal/config"
+)
+
+// A value inside a file inside a file must be REDACTED, not merely reported.
+//
+// The read side already descends into embedded parts, so these values are found.
+// The write side did not, so they were reported and then shipped in cleartext.
+// Measured on the parent commit, with the container's OWN SSN correctly redacted
+// in every row:
+//
+//	outer.docx -> word/embeddings/oleObject1.docx   SSN in cleartext, exit 0, no warning
+//	outer.docx -> word/media/image1.jpg (EXIF)      SSN in cleartext, exit 0, no warning
+//	both of the above in one document               2 values in cleartext
+//
+// Only reported findings are redacted, so a value the redactor cannot reach is a
+// leak that reports as success — a file sitting in a directory named "redacted"
+// with the SSN still in it. See #305.
+//
+// These tests are the SINK test for that. They assert on the bytes actually
+// written, descending into every nested archive member, because grepping a .docx
+// searches COMPRESSED bytes and finds nothing — which is indistinguishable from a
+// clean file and is the most common way a leak in this area gets certified fixed.
+
+const (
+	// childSSN lives only inside the embedded part.
+	childSSN = "452-11-9384"
+	// outerSSN lives in the container's own body. It is the control: it was already
+	// redacted correctly before this change, so a test that stops finding it has
+	// broken the pre-existing behaviour rather than proving the fix.
+	outerSSN = "536-90-4271"
+)
+
+func TestEmbeddedDocumentValuesAreRedacted(t *testing.T) {
+	for _, tc := range []struct {
+		name  string
+		build func(t *testing.T, dir string) string
+	}{
+		{"office document embedded in a document", func(t *testing.T, dir string) string {
+			return writeDocx(t, dir, "outer_docx.docx", map[string][]byte{
+				"word/embeddings/oleObject1.docx": buildChildDocx(t, childSSN),
+			})
+		}},
+		{"image EXIF embedded in a document", func(t *testing.T, dir string) string {
+			return writeDocx(t, dir, "outer_img.docx", map[string][]byte{
+				"word/media/image1.jpg": buildJPEGWithEXIF(t, "Contact SSN "+childSSN),
+			})
+		}},
+		{"both kinds of child in one document", func(t *testing.T, dir string) string {
+			return writeDocx(t, dir, "outer_both.docx", map[string][]byte{
+				"word/embeddings/oleObject1.docx": buildChildDocx(t, childSSN),
+				"word/media/image1.jpg":           buildJPEGWithEXIF(t, "Contact SSN "+childSSN),
+			})
+		}},
+		{"child nested two levels deep", func(t *testing.T, dir string) string {
+			// A .docx inside a .docx inside a .docx. Well within embedded.MaxDepth,
+			// so nothing may be skipped and nothing may leak.
+			inner := buildChildDocx(t, childSSN)
+			mid := buildDocxBytes(t, "middle document", map[string][]byte{
+				"word/embeddings/inner.docx": inner,
+			})
+			return writeDocx(t, dir, "outer_deep.docx", map[string][]byte{
+				"word/embeddings/mid.docx": mid,
+			})
+		}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			dir := t.TempDir()
+			in := tc.build(t, dir)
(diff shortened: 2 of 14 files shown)


sectoramen:
Holding this as a draft until one measurement is redone.

The real-corpus delta I was about to report was computed against the **wrong baseline**. I built the "before" binary from `9fe22e7`, which predates #310, while this branch is based on `880f12a` which includes it. So the +44,827-finding delta is contaminated by #310s text-fallback change and cannot be attributed to this PR.

Re-running against `880f12a` — the actual parent — and will post the corrected new/lost breakdown by type before marking this ready.

The other gates were measured correctly and stand: build/vet/gofmt clean, 66-package suite green, golden corpus with zero moves, `make score` unchanged (`recall_all=1.0000 whole_leak=0`), the linear scaling curve, non-vacuity against the parent commit, and 9/9 mutations caught.

The contaminated comparison also showed 3 LOST `API_KEY_OR_SECRET` findings. Whether those are mine or #310s is exactly what the corrected run will settle; an unexplained loss is a blocker either way.

sectoramen:
Corrected corpus measurement, plus two fixes it drove. Pushed as `926afdb`.

## Corpus A/B against the ACTUAL parent (`880f12a`)

| | findings |
|---|---|
| parent `880f12a` | 150,430 |
| this branch | 150,430 |
| **delta** | **+0** |

New = 3, Lost = 3 with line number in the diff key; **0 and 0 with line removed** — those six are the same three `API_KEY_OR_SECRET` findings at line numbers shifted by the text the embedded parts add above them. So **zero real losses**, which clears the recall floor, and zero detection change on this corpus. Same-binary noise floor was exactly 0, so the comparison is trustworthy.

Zero detection delta is the expected and correct result here: the corpus embedded parts are overwhelmingly `.png` (1,857), which the parent already admitted. This PRs value is on the **redaction** side, which a detection count cannot show — that is precisely why the test plan scores the redaction layer separately, and it is covered by the fixtures and the score corpus residue layer.

## What the first measurement caught (a real defect, now fixed)

Deriving admission from pipeline capability admitted `.svg`, which is XML text, so the byte-sniffing text path claimed it — and its content is coordinate data. On one real deck with 171 embedded `.svg` parts, `PHONE` went **0 → 1,095, of which 826 were HIGH**, matching path geometry like `43.5968 15.4721 43.4928 15.7281`. 826 HIGH false positives in one file is not coverage; HIGH is the band operators triage first.

`embedded.SkipTextPipeline` now excludes `.svg`/`.emf`/`.wmf`/`.wdp` on **both** sides so the halves still admit the same set. That deck is back to 0 `PHONE`, and the embedded-`.docx` and image-EXIF leak fixtures still redact to 0 leaks.

This is a deferral, not a claim those parts are uninteresting: an `.svg`s `<text>`/`<title>`/`<desc>` nodes are real prose, and a fixture with `<text>Employee SSN: ...</text>` is detected at confidence 90 standalone. Recovering that needs an SVG-aware extractor that reads text nodes and ignores geometry — its own change, with its own precision evidence.

## Windows CI

Two **test** bugs, product unaffected:

- `TestSafeExtNeverYieldsAPathComponent` asserted a hardcoded `/tmp/sandbox/` prefix; Windows `filepath.Join` yields `\tmp\sandbox\...`, so it failed on every case while the extensions were all correct. Now compares `filepath.Dir`, separator-agnostic.
- `TestRedactEmbeddedTempFileIsNotWorldReadable` asserted `0600`/`0700`; Windows `os.FileMode` carries no Unix permission bits (it reports `-rw-rw-rw-`). Skipped on Windows only; the modes are still passed unconditionally and still asserted on Unix.

Both were green locally and red in CI.

## CodeQL check

Not from this PR. The two HIGH alerts are **Trivy** findings on `refs/heads/main` against the **Go stdlib v1.26.5 toolchain** (CVE-2026-46600, CVE-2026-39821; fixed in Go 1.26.6+). They will fail every PR until the toolchain is bumped — being handled separately.

rustic (approved):

Issue #326, opened 2026-08-17
A directory walk silently drops every symlink, so a symlinked file's contents are never scanned and never disclosed — and --fail-on-incomplete exits 0 despite the docs promising 3


sectoramen:
A symlink inside a scanned directory is dropped at discovery with no record anywhere. The same file **is** scanned when named directly, so the two discovery paths disagree, and the coverage loss is invisible in every channel.

## Reproduction

```console
$ ls -la d/
dangling.txt -> /tmp/nonexistent
link.txt     -> /tmp/target/secret.txt      # contains a card number
normal.txt                                   # contains an SSN

$ ferret-scan --file d --checks all --limit 0
[HIGH] ssn SSN 100.00% line 1 [HIDDEN] normal.txt

═══ Scan Summary ═══
Files: 1 scanned | Findings: 1 (1 high, 0 medium, 0 low)
```

`Files: 1 scanned` for a directory of 3 entries. No warning, no `NOT FULLY EXAMINED` block, and the symlinked card number is neither reported nor disclosed.

```console
$ ferret-scan --file d --checks all --fail-on-incomplete ; echo $?
0
```

## The same file IS scannable — only the walk drops it

| how the file is reached | result |
|---|---|
| target file named directly | `VISA` found |
| **symlink named directly** | **`VISA` found** — the single-file path follows the link |
| symlink via directory walk | **silently dropped** |

So this is not a deliberate "we do not follow symlinks" policy. It is an inconsistency: identical bytes are scanned or ignored depending on how the path is spelled.

## Mechanism

`cmd/main.go:2570`:

```go
// Only add regular files
if info.Mode().IsRegular() {
    ...
}
// <-- no else
```

`filepath.Walk` supplies `os.Lstat` info, so a symlinks mode is `ModeSymlink`, never regular. The branch has **no else**, so the entry is dropped without being recorded in `result.SkippedFiles`, without a warning, and without reaching any counter. It is absent from `total_files`, from `files_skipped`, and from `files_not_examined`.

The single-file path (`cmd/main.go:2494`) uses `os.Stat`, which follows the link and reports the target as regular — hence the divergence.

## The documented behaviour is the opposite

`cmd/main.go:47-50` states:

[...]

PR #332, opened 2026-08-18
fix(cli): count a file refused for size as unexamined, and decide the warning from capability

### LINKED ISSUE #324: A file refused for being >100MB is never counted as unexamined: absent from the summary, absent from files_not_examined, and --fail-on-incomplete exits 0
A file the scanner refuses for size is silently dropped at discovery, so it is missing from the coverage accounting entirely — including the denominator.

## Reproduction

One 105MB file with an unusual extension (a browser partial download), plus one small text file holding an SSN:

```console
$ ls -la d/
-rw-r--r--  110100480  .com.brave.Browser.ABCdef
-rw-r--r--         26  normal.txt

$ ferret-scan --file 'd/*' --checks SSN --limit 0
Warning: Skipping d/.com.brave.Browser.ABCdef: file too large (> 100MB)
[HIGH] ssn SSN 100.00% line 1 [HIDDEN] normal.txt

═══ Scan Summary ═══
Files: 1 scanned | Findings: 1 (1 high, 0 medium, 0 low)
```

**No `NOT FULLY EXAMINED` block at all.** The machine artifact is worse:

```
total_files        : 1     <-- the skipped file is not even in the denominator
files_processed    : 1
files_skipped      : 0
files_not_examined : (absent)
```

```console
$ ferret-scan --file 'd/*' --checks SSN --fail-on-incomplete ; echo $?
0
```

So the artifact reports a **complete, clean scan** over a directory containing a 105MB file that was never opened. The only trace is a stderr line, which is exactly what CI discards (`--format json --output report.json`).

This is strictly worse than the cases already closed by #277, #284 and #296: those at least appeared in the summary block and in `files_not_examined`. Here the file is removed from the accounting *before* the scan starts, so even the denominator hides it.

### (a) The silencing predicate is an allowlist, and it is backwards

`cmd/main.go:2455-2467` (glob path) and `isUnsupportedType` (`cmd/main.go:2294`) both carry the **same hardcoded 11-entry list**:

```go
".mp4", ".mov", ".avi", ".mkv", ".dmg", ".iso", ".img", ".zip", ".tar", ".gz", ".7z"
```

A too-large file is warned about unless its extension is in that list. So the tool is quiet about 11 known-big binary types and **noisy about everything else** — including files it could never have scanned anyway. Browser partial downloads (`.com.brave.Browser.HXWeCx`) have random suffixes, so no extension list will ever cover them, and every one produces a warning line.

The predicate should be inverted: **would the router have processed this file at all?** `router.CanProcessFile(path, enablePreprocessors)` already answers that. An unprocessable type needs no warning at any size — it was never going to be scanned. That also deletes the duplicated list, which is the same "derive from capability, not from a hand-maintained list" correction made for embedded parts in #311.

### (b) A size-refused file is not "skipped", it is UNEXAMINED

`ScanStats.FilesNotExamined` documents the distinction itself:

> counts files the tool could not read, parse or extract text from. They are NOT "skipped" (an unsupported type the user does not expect a result for)

A `.mp4` skipped as unsupported is legitimately "skipped" — nobody expected a finding. A **processable** file (`.docx`, `.txt`, `.pdf`) refused for size is the opposite: the user did expect a result and did not get one, and nothing says so. By the fields own definition that is `files_not_examined`.

Note the two paths are already inconsistent with each other: the **single-file** path (`cmd/main.go:2494+`) records size refusals in `result.SkippedFiles` with a `Silent` flag, while the **glob** path prints inline and records nothing — so glob-discovered refusals are not even counted by `totalSkipped`.

## The seam already exists

`collectUnscanned(unreadable, emptyExtraction, failed, incomplete)` in `cmd/unscanned_report.go:230` builds `[]unscannedEntry`, and `cmd/main.go:2050` sets `FilesNotExamined: len(unscannedEntries)`. There is already an `unscannedCause` enum (`causeUnreadable`, `causeUnparseable`, `causeNoText`, `causeCutShort`).

So the fix is shaped:

1. add a `causeTooLarge` ("too large to scan (>100MB)")
2. carry discovery-time size refusals of **processable** types through to `collectUnscanned` as a fifth input
3. they then flow automatically into the summary block, `files_not_examined` in every format, and `--fail-on-incomplete`
4. with the count in the summary, the per-file stderr warnings can be dropped — which is the reported annoyance

## Behaviour change to call out

Documents that currently exit 0 will exit **3** under `--fail-on-incomplete` once size refusals count as unexamined. That is the correct answer, but it will newly fail pipelines that scan directories containing large media, so it needs a real corpus measurement in the PR and a CHANGELOG note.

## Acceptance criteria

- a 105MB processable file appears in the `NOT FULLY EXAMINED` block, in `files_not_examined`, and in `total_files`
- `--fail-on-incomplete` exits 3 for it
- an unprocessable type (any size) produces **no** warning and is not counted as unexamined
- the duplicated 11-entry extension list is gone
- glob and single-file discovery paths agree
- real-corpus delta reported, including how many files newly count as unexamined

Reported from a real scan; the reproduction above is synthetic.

sectoramen:
Closes #324. Implements **defect (a)** and **defect (b) option (a)** — the exit-code change.

A file refused for being over the size limit reached **no counter at all**. Measured on a directory holding one 105MB text file plus a small file with an SSN:

```
Files: 1 scanned | Findings: 1
total_files 1, files_processed 1, files_skipped 0, files_not_examined absent
--fail-on-incomplete -> exit 0
```

The artifact described a complete, clean scan of a directory containing a file that was never opened. The only trace was a stderr line — exactly what CI discards when it redirects stdout to a report.

## Defect (a) — the silencing predicate was a hardcoded list, and backwards

It was an 11-extension list duplicated at **three** call sites, not two: the single-file path, the glob loop, and the directory walk (the latter two carrying it inline). The tool was quiet about a few known-big binary types and noisy about everything else, including files it could never have scanned at any size. Browser partial downloads have random suffixes, so no extension list can ever cover them.

Replaced with `router.CanProcessType`, which asks the only question that matters: *would we have processed this type at all?* `CanProcessFile` cannot answer it — its own size gate returns "File too large" before the type is considered, so asking it about an oversize file is circular. All three copies of the list are gone.

**Deliberate consequence:** oversize **video and audio** become disclosed rather than silent when preprocessors are on, because their metadata *is* extracted and scanned — so an oversize video that went unscanned is a genuine coverage loss. That follows from deriving the predicate from capability instead of from a list.

## Defect (b) — a size-refused processable file is UNEXAMINED, not skipped

`ScanStats` already draws the distinction: a skipped file is an unsupported type nobody expected a result for; an unexamined one was expected to produce something and did not.

```
Files: 1 scanned, 1 NOT examined, 1 skipped | Findings: 1
NOT FULLY EXAMINED: 1 of 2 files — findings may be missing
  file too large to scan (1)
    big_report.txt  file too large (max size: 100MB)

total_files 2, files_processed 1, files_skipped 1, files_not_examined 1
--fail-on-incomplete -> exit 3
```

An **unprocessable** type refused for size stays a silent skip, but is now *recorded* rather than dropped, so it stays in the denominator. Verified on all three routes (single file, glob, recursive walk).

⚠️ **This is an exit-code behaviour change**: `--fail-on-incomplete` now returns 3 where it returned 0. That is the flag doing what its name says, and the capability predicate bounds the blast radius to files the run could actually have scanned.

## Also fixed here, because it is in the switch this change had to touch

`causeNotFollowed` (added for #326) was **never mapped** in `toFormatterNotExamined` and fell through to the conservative default — so every refused symlink was reported as "cannot read" in machine formats while the text report said "symlink not followed", asserting an I/O failure that never happened. That is exactly the mislabelling the cmd-side comment warns a new cause would cause.

Both new causes are mapped, and a test asserts every cmd cause maps to a **distinct** formatter cause with identical wording on both sides.

The cause now travels on `SkippedFile`, whose zero value is `causeUnreadable` — so a producer that forgets to set it silently mislabels its entry. A test covers the symlink producer for exactly that.

One **stdout write on the scan path** moved to stderr (`getFilesToProcess`'s error path). stdout is the machine artifact; a diagnostic spliced into it leaves an unparseable document at exit 0. Same defect class as the media-error write already moved in #325.

## Non-vacuity

| mutation | failure |
|---|---|
| `CanProcessType` always false | `big_report.txt is not in UnexaminedFiles []` |
| oversize routed back to skipped | same, plus the preprocessors-off case |
| drop the `causeNotFollowed` mapping | `mapped to "cannot read"` |
| forget `Cause` on symlink disclosure | `cause = cannot read, want causeNotFollowed` |

A fixture note worth keeping: `Truncate` extends with NULs, so a sparse "text" file whose leading text is shorter than the 512-byte sniff window reads as **binary** and silently tests the wrong branch. Both helpers fill the window — I hit this and the test caught it.

`make all` green; full suite 66 packages, 0 failures; golden corpus unchanged.

---

**Not self-merging** — leaving this for @rustic to review.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -1549,6 +1549,33 @@ func main() {
 			}
 			os.Exit(2) // Use exit code 2 for no files to process as per design
 		} else {
+			// Disclose discovery-time coverage losses BEFORE exiting.
+			//
+			// This path used to print "No files to process" and exit 0 unconditionally,
+			// which is the exact artifact this report exists to prevent. A single
+			// oversize processable file is refused at discovery, so filesToProcess is
+			// empty and every disclosure downstream — collectUnscanned, the stats
+			// denominator, the NOT FULLY EXAMINED block, and both
+			// resolveIncompleteExitCode calls — was skipped. The run reported a clean,
+			// complete result for a file it never opened, and --fail-on-incomplete
+			// returned 0. It also printed strictly LESS than before this branch, because
+			// discovery warnings were deferred to a report that never ran.
+			//
+			// Same shape for a directory or glob whose inputs are ALL refused.
+			if len(discoveryUnexamined) > 0 {
+				entries := make([]unscannedEntry, 0, len(discoveryUnexamined))
+				for _, u := range discoveryUnexamined {
+					entries = append(entries, unscannedEntry{Path: u.Path, Cause: u.Cause, Detail: u.Reason})
+				}
+				var report strings.Builder
+				if writeUnscannedReport(&report, entries, len(entries), *failOnIncomplete, finalConfig.debug) {
+					fmt.Fprint(os.Stderr, report.String())
+				}
+				fmt.Println("No files to process")
+				// Honour the flag: these files were expected to produce results and did
+				// not, which is precisely what code 3 means.
+				os.Exit(resolveIncompleteExitCode(0, finalConfig.failOnIncomplete, len(entries)))
+			}
 			fmt.Println("No files to process")
 			os.Exit(0)
 		}
@@ -2663,7 +2690,6 @@ func getFilesToProcess(inputPath string, recursive bool, excludePatterns []strin
 						Reason: "file too large (max size: 100MB)",
 						Cause:  causeTooLarge,
 					})
-					skippedFiles = append(skippedFiles, cleanWalkPath)
 				} else {
 					// Unprocessable at any size: a genuine skip.
 					result.SkippedFiles = append(result.SkippedFiles, SkippedFile{
@@ -2671,8 +2697,13 @@ func getFilesToProcess(inputPath string, recursive bool, excludePatterns []strin
 						Reason: "file too large (max size: 100MB)",
 						Silent: true,
 					})
-					skippedFiles = append(skippedFiles, cleanWalkPath)
 				}
+				// Deliberately NOT added to the walk's local skippedFiles slice, which
+				// feeds only "Skipped N files or directories due to errors". A size
+				// refusal is not a walk error, and both branches above now record the
+				// file in a real counter — so appending there both mislabelled it and
+				// reported the same file twice, once as an error and once in the
+				// NOT FULLY EXAMINED block.
 			} else if info.Mode()&os.ModeSymlink != 0 {
 				// A symlink. filepath.Walk hands us Lstat info, so a link is never
 				// ModeRegular and used to fall past the branch above with NO else —
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -1502,9 +1502,14 @@ func main() {
 			}
 		}
 
-		result, err := getFilesToProcess(cleanPath, finalConfig.recursive, finalConfig.excludePatterns, ignoreMatcher)
+		result, err := getFilesToProcess(cleanPath, finalConfig.recursive, finalConfig.excludePatterns, ignoreMatcher, finalConfig.enablePreprocessors)
 		if err != nil {
-			fmt.Printf("Error processing %s: %v\n", inputPath, err)
+			// stderr, not stdout: this is on the scan path, and a caller redirecting
+			// stdout to a machine artifact (--format json > report.json) would
+			// otherwise get a diagnostic spliced into the middle of the document,
+			// leaving an unparseable file at exit 0. Same defect class as the
+			// media-error write already moved off stdout.
+			fmt.Fprintf(os.Stderr, "Error processing %s: %v\n", inputPath, err)
 			continue
 		}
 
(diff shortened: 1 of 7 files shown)


sectoramen:
Pushed `86645fc`, fixing three defects found in this branch's own new code by reviewing a real ~1,870-file scan. Two of them silenced exactly the disclosure this PR exists to add.

**#339 — an all-inputs-refused run exited 0 with no disclosure.** A solo oversize processable file leaves `filesToProcess` empty, so `main()` took the "No files to process" early exit — one stdout line and `os.Exit(0)`, *before* `collectUnscanned`, the stats denominator, `writeUnscannedReport` and both `resolveIncompleteExitCode` calls. The one invocation this feature exists for produced a silent, complete-looking, zero-exit result, and printed strictly *less* than main did. Now:

```
NOT FULLY EXAMINED: 1 of 1 file — findings may be missing
  file too large to scan (1)
    .../big_report.txt  file too large (max size: 100MB)
exit 3 with --fail-on-incomplete
```

**#336 defect 1 — the collapsed breakdown dropped every cause added since the original four.** Above `inlineDetailLimit` the per-cause tally named the four legacy causes explicitly while the header printed `len(entries)`, so `causeNotFollowed` (#326) and `causeTooLarge` (this PR) were counted and given no bucket line — the observed `65` header over `23 + 41 = 64` buckets, with the missing entry being the oversize file. Now derived from the causes actually present, ordered by enum value (which is the deliberate presentation order).

**A size refusal was double-reported as a walk error.** Both size branches also appended to the walk's local `skippedFiles` slice, which feeds only "Skipped N files or directories due to errors". Removed.

**Why the tests missed all of this:** the route table's comment said "all three discovery routes" over a **two**-entry table, and both fixtures create a small companion file — so `FilesToProcess` was never empty in any test. Added the solo-file route, plus `TestBreakdownAccountsForEveryCause`, which asserts the *invariant* (buckets account for every entry the header counts) rather than a list of labels. Restoring the hardcoded list fails it naming both missing causes.

**Correcting my own commit message on `a625462`:** it claimed an unprocessable oversize file "stays in the denominator". It does not — that entry reaches only `files_skipped`, which #336 defect 2 shows `total_files` excludes. I have not fixed that here, to avoid widening the PR; it and the walk-access-error gap (#336 defect 3) remain tracked on #336.

rustic (approved):

Issue #321, opened 2026-08-15
Two genuine occurrences of one value on the same line render as byte-identical findings, so they read as duplicates and cannot be told apart by any consumer


sectoramen:
When a value appears more than once on a single line, the report emits one finding per occurrence — correct — but **nothing in the output distinguishes them**. A human reads the identical rows as a double-emit bug, and a machine consumer deduping on the available fields collapses them and under-counts.

## Reproduction

```console
$ printf 'From: jane.doe@example.com Sent: Monday To: bob@example.com; jane.doe@example.com; carol@example.com\n' > hdr.txt
$ ferret-scan --file hdr.txt --checks EMAIL --show-match

[HIGH] email BUSINESS 100.00% line 1  jane.doe@example.com  hdr.txt
[HIGH] email BUSINESS 100.00% line 1  jane.doe@example.com  hdr.txt
[HIGH] email BUSINESS 100.00% line 1  bob@example.com       hdr.txt
[HIGH] email BUSINESS 100.00% line 1  carol@example.com     hdr.txt
```

4 findings, 3 distinct values. The two rows for the repeated address are identical in every visible column.

## Not a detection bug, and not a leak

Both are true occurrences at different byte offsets, which is exactly what the byte-span dedup key from #202 is designed to preserve — and preserving them is what makes redaction remove **both**:

```
From: j*******@example.com Sent: Monday To: b**@example.com; j*******@example.com; c****@example.com
```

Verified: zero occurrences of the raw value survive. So detection and redaction are correct; only the report is ambiguous.

## The actual defect: the two findings are indistinguishable

Comparing the two JSON findings field by field:

```
top-level fields that DIFFER:  NONE
metadata keys that DIFFER:     NONE
any offset/column field:       NONE
```

`JSONMatch` carries `line_number` but no column, byte offset, or occurrence index. Consequences:

- **Humans** reasonably report it as a duplicate-emit bug. It was reported that way, against a real document, which is how this issue exists.
- **Machine consumers** that dedupe defensively — a reasonable thing to do given #308 shipped a real double-emit — will key on `(filename, line_number, type, text, confidence)` and drop one of two genuine findings. The count they report is then wrong in the safe-looking direction.

## Why it shows up in practice rather than in fixtures

[...]

Issue #289, opened 2026-08-12
SOCIAL_MEDIA clustering replaces real matches with a synthetic match that is not in the file, so clustered handles survive redaction in cleartext


sectoramen:
Found while reviewing #285 (which exposes `Match.Context` on the public `Finding`). The context question turned out to be the cosmetic half of this; the redaction half is a cleartext leak.

## The leak

When social-media clustering fires, the individual matches it grouped are **replaced** by a single synthesized match whose `Text` does not occur anywhere in the file. Every redactor locates a match by its text, so the synthesized match redacts nothing — and the real spans that were correctly detected have already been discarded.

Reproduction on `main` @ 1824563 (see the two prerequisites below — they matter):

```console
$ cat > cluster.txt <<'EOF'
Profile: https://twitter.com/janedoe
Also https://instagram.com/janedoe
And https://linkedin.com/in/janedoe
EOF

$ ferret-scan --config examples/ferret.yaml --file cluster.txt --checks SOCIAL_MEDIA \
    --enable-redaction --redaction-output-dir ./red
[HIGH  ] SOCIAL_MEDIA SOCIAL_MEDIA_CLUSTER   95.00% line     1 [HIDDEN]   cluster.txt
Files: 1 scanned | Findings: 1 (1 high, 0 medium, 0 low)

$ diff cluster.txt red/**/cluster.txt && echo IDENTICAL
IDENTICAL
```

One HIGH finding at 95%, a "redacted" file byte-identical to the input, both handles in the clear.

**Prerequisites, so this doesn't look unreproducible:**

1. **A config with social-media platform patterns.** SOCIAL_MEDIA ships none built-in, so with no config file the scan reports `No matches found` and there is nothing to leak. The shipped `examples/ferret.yaml` (what `make install-config` installs) reproduces it; so does the repo's tracked `config.yaml`.
2. **Do not put the fixture under a path containing `tmp`.** `exclude_patterns` in the shipped config includes `tmp`, so a fixture in `/tmp` is skipped with `excluded by --exclude pattern` and never scanned.

**Control — the same handles, not clustered, redact correctly:**

```console
$ echo 'Profile: https://twitter.com/janedoe' > one.txt   →  Profile: ***************************
$ echo 'And https://linkedin.com/in/janedoe' > one.txt    →  And *******************************
```

[...]

PR #345, opened 2026-08-18
fix(config): drop the mandatory trailing slash from social profile patterns, so the canonical URL is detected

### LINKED ISSUE #343: Shipped config's instagram pattern requires a trailing slash, so the canonical profile URL is never detected
Found while fixing #289. Separate from that leak — this is a detection gap in the **shipped** config, so every user who runs `make install-config` has it.

## The pattern

`examples/ferret.yaml`, `social_media.platform_patterns`:

```yaml
instagram:
  - "(?i)https?://(?:www\\.)?instagram\\.com/[a-zA-Z0-9_.]+/"
```

The trailing `/` is mandatory in the regex. Every other platform omits it:

```yaml
linkedin:  ".../linkedin\\.com/in/[a-zA-Z0-9_-]+"      # no trailing slash
twitter:   ".../(twitter|x)\\.com/[a-zA-Z0-9_]+"        # no trailing slash
facebook:  ".../(facebook|fb)\\.com/[a-zA-Z0-9._-]+"    # no trailing slash
youtube:   ".../youtube\\.com/(?:user|c|channel)/[a-zA-Z0-9_-]+"
```

## Measured

```console
$ echo 'Also https://instagram.com/janedoe/' > a.txt   # trailing slash
$ ferret-scan --config examples/ferret.yaml --file a.txt --checks SOCIAL_MEDIA
  -> 1 finding

$ echo 'Also https://instagram.com/janedoe'  > b.txt   # canonical form
$ ferret-scan --config examples/ferret.yaml --file b.txt --checks SOCIAL_MEDIA
  -> 0 findings, "No matches found."
```

The form without the trailing slash is what a browser address bar shows, what people paste, and what appears in documents. So the common case is the undetected one, and — per the sink rule — an undetected value is also an unredacted one: it passes straight through `--enable-redaction` in cleartext with no disclosure.

This is why the 3-line fixture in #289 still shows `Also https://instagram.com/janedoe` in the output after that leak is fixed. Twitter and LinkedIn redact; instagram was never detected, so there was nothing to redact.

## Fix

Drop the trailing slash so it matches the other platforms:

```yaml
instagram:
  - "(?i)https?://(?:www\\.)?instagram\\.com/[a-zA-Z0-9_.]+"
```

Both forms then match, since the character class does not consume `/`.

Worth checking the same file for other patterns whose anchoring differs from their siblings — the inconsistency here is the tell, and nothing in the test suite compares platform patterns against each other.

Also worth a test that asserts each shipped platform pattern matches BOTH the slashed and unslashed profile URL, since a config regression is invisible to the Go tests today: the patterns live in YAML, and no test loads the shipped config and exercises them.

## Note on scope

`--checks SOCIAL_MEDIA` reports nothing at all with no config file (the validator ships no built-in platform patterns), so this only affects users who install the example config or write their own from it. That is the documented setup path, so it is the shipped behaviour.

sectoramen:
Closes #343. Found while fixing #289.

## The bug

Several shipped social-media profile patterns required a **trailing slash**:

```yaml
instagram:
  - "(?i)https?://(?:www\\.)?instagram\\.com/[a-zA-Z0-9_.]+/"
```

so the form a browser address bar shows — and the form people paste — never matched:

```
https://instagram.com/janedoe/   -> 1 finding
https://instagram.com/janedoe    -> 0 findings, "No matches found."
```

By the sink rule an undetected value is also an **unredacted** one: it passed straight through `--enable-redaction` in cleartext with no disclosure. This is why the #289 fixture still showed `Also https://instagram.com/janedoe` after that leak was fixed — instagram was never detected, so there was nothing to redact.

## Scope — bigger than I first thought

I found instagram by hand and assumed it was the only one. **Writing the general test found three more platforms with the identical mistake:**

| platform | patterns affected |
|---|---|
| instagram | 3 (profile, `instagr.am` short domain, without protocol) |
| tiktok | 2 |
| snapchat | 2 |
| pinterest | 2 |
| `examples/ferret.yaml`, `examples/ferret-windows.yaml` | 1 each |

**11 patterns across three shipped files.** No other platform required a trailing slash, which is what made it a typo rather than a deliberate choice.

Dropping the slash matches both forms, because the character class stops at `/`. It adds no false-positive class — a post URL like `instagram.com/p/ABC123` matched the slashed pattern too (as `p/`), so that's unchanged.

## Why no gate caught it

Nothing in the test suite loaded a shipped config and exercised its regexes — the patterns live in YAML and the Go tests never read them. Two tests close that:

- each shipped config's platform patterns must match **both** the slashed and unslashed profile URL, and must **compile** (an RE2-invalid pattern is silently dropped at runtime, so a platform stops detecting with no warning);
- no shipped pattern may end with a character class followed by `/`, which names the specific mistake so a future failure is self-explanatory.

The first also fails if none of the expected platforms are present, so a config rename can't leave it asserting nothing.

Verified on the binary: instagram, tiktok, snapchat and pinterest all now detect the canonical profile URL, and redaction removes it.

## Pre-existing, not addressed here

A single URL reports **twice**, because the with-protocol and without-protocol patterns both match (nested spans). `ResolveOverlaps` collapses those on the redaction path but not on the reporting path, so the finding count is inflated while redaction is correct. Same family as #321; out of scope for a config fix.

---

**Not self-merging** — for @rustic.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -367,9 +367,9 @@ validators:
 
       # Instagram - Photo/video sharing platform
       instagram:
-        - "(?i)https?://(?:www\\.)?instagram\\.com/[a-zA-Z0-9_.]+/" # Profile URLs
-        - "(?i)https?://(?:www\\.)?instagr\\.am/[a-zA-Z0-9_.]+/" # Short domain
-        - "(?i)instagram\\.com/[a-zA-Z0-9_.]+/" # Without protocol
+        - "(?i)https?://(?:www\\.)?instagram\\.com/[a-zA-Z0-9_.]+" # Profile URLs
+        - "(?i)https?://(?:www\\.)?instagr\\.am/[a-zA-Z0-9_.]+" # Short domain
+        - "(?i)instagram\\.com/[a-zA-Z0-9_.]+" # Without protocol
 
       # YouTube - Video sharing platform
       youtube:
@@ -380,9 +380,9 @@ validators:
 
       # TikTok - Short video platform
       tiktok:
-        - "(?i)https?://(?:www\\.)?tiktok\\.com/@[a-zA-Z0-9_.]+/" # Profile URLs
+        - "(?i)https?://(?:www\\.)?tiktok\\.com/@[a-zA-Z0-9_.]+" # Profile URLs
         - "(?i)https?://(?:www\\.)?tiktok\\.com/t/[a-zA-Z0-9]+" # Short URLs
-        - "(?i)tiktok\\.com/@[a-zA-Z0-9_.]+/" # Without protocol
+        - "(?i)tiktok\\.com/@[a-zA-Z0-9_.]+" # Without protocol
 
       # Discord - Gaming/community chat platform
       discord:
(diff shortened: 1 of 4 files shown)


rustic (approved):

sectoramen:
Correcting my own claim on this PR, post-merge: **"It adds no false-positive class" was wrong.** Filed as #349 with the measurement.

7 lines of ordinary documentation URLs (`instagram.com/explore`, `/reels`, `/tv`, `/legal`, `pinterest.com/today`, `/business`, `/categories`): **0 findings before, 14 after, all HIGH 100** — and `--enable-redaction` overwrites all of them, so the redactor corrupts legitimate content.

I missed it because I checked URLs with a **second** path segment (`instagram.com/p/ABC123`, `pinterest.com/pin/123`) and measured 2 before / 2 after. Those already matched via the `<segment>/` form, so they were precisely the cases that could not change — the wrong control. The class that changed is a **single** path segment with no trailing slash.

instagram and pinterest are the two platforms whose patterns have no path prefix, so any single segment reads as a handle; tiktok (`/@`) and snapchat (`/add/`) are unaffected, and the trailing-slash fix for those stands.

#349 carries the full diagnosis, including that `validateNotFalsePositive` is a −30 confidence penalty rather than a veto (100 → 70 still reports) and that `isOverlyBroadMatch`'s platform switch uses bare-only keys where every other switch in the file matches both forms.

Issue #337, opened 2026-08-18
One finding pins its file's entire extracted buffer: match/context strings are substrings of the full content, and all findings are retained twice until exit (1.74 GB RSS on a ~1,900-file scan)


sectoramen:
# Summary

A directory scan of ~1,870 files (168,645 findings) held **1.74 GB resident** (1.33 GB private) at ~3 min wall time. Back-of-envelope against the code brackets exactly that (1.2–1.9 GB), and the two dominant terms are defects, not workload:

1. **Substring retention:** every match/context string is a Go substring of the file's *entire* extracted content, so one 16-byte finding pins the whole buffer of the file it came from.
2. **Full-run retention:** all findings — plus a second struct copy — stay live until process exit; `--limit` bounds display only, and the one memory guard (`--max-live-bytes`) releases its budget while the findings still pin the bytes.

On a larger corpus (or larger finding-bearing files) this OOMs with no backpressure anywhere.

## Mechanism 1 — one finding pins its file's entire extracted buffer

Validators split the whole extracted content and slice lines directly into the match:

- `internal/validators/email/validator.go:233` — `lines := strings.Split(content, "\n")` (Go substrings share the parent's backing array)
- `:263` — `match := line[matchIndex:matchEnd]` → `Match.Text`
- `:273` — `FullLine: line`; `:285-286` — `BeforeText`/`AfterText` sliced from the same line
- Same pattern in `internal/validators/ssn/validator.go:337/374/478-480/598` and 20+ other validators; the shared helper `internal/detector/line_context.go` (`LineContext`) slices identically and stores `FullLine: line`.

`content` is `ProcessedContent.Text` — the full extracted concatenation for the file (`internal/parallel/validator_runner.go` passes it whole; the dual-path route hands validators `FullText` = the same string). `git grep strings.Clone` over non-test code: **zero hits**. Consequence: a 30 MB minified JSON with a single email retains 30 MB for one finding; on a single-line file, `FullLine` *is* the whole file (and is emitted per-finding into JSON/YAML output under `--verbose --show-match`).

The plaintext preprocessor reads whole files up to the 100 MB cap into one string with no chunking, so the pinnable unit really is the full file.

## Mechanism 2 — everything is retained until exit, twice

- Per-file matches are appended into a process-lifetime slice (`internal/parallel/parallel_processor.go:143`, `:215`) and again into the CLI's `allMatches`.
[...]

PR #347, opened 2026-08-18
perf(preprocessors): assemble embedded-media text with a Builder instead of re-copying it per part

### LINKED ISSUE #338: ProcessEmbeddedMedia builds combined embedded text with result += block: O(n²) bytes copied on media-rich Office files
# Summary

`ProcessEmbeddedMedia` assembles the combined embedded-media text with string concatenation inside the per-part loop:

- `internal/preprocessors/base_metadata_preprocessor.go:247` — `var result string`
- `:255` — `for i, media := range embeddedMedia {`
- `:327` — `result += block`

Go's `string +=` reallocates and re-copies the accumulated prefix on every iteration, so building the combined text is **O(n²) bytes copied** across n successfully processed embedded parts. For an Office file with many embedded parts this is pure transient allocation and GC churn during extraction.

## Scope of "n"

The effective n is the count of parts that survive the pipeline to an append:

- SVG/EMF/WMF/WDP are excluded up front by `embedded.SkipTextPipeline`, so vector-heavy decks are unaffected.
- Raster-image-heavy documents (each image round-trips through the router for metadata text) and files with many embedded documents/PDFs — whose per-part *text blocks are large* — are the worst cases: the quadratic is in bytes, so a few dozen multi-hundred-KB blocks already costs tens of MB of copying.

Memory-transient only: it does not affect scan results or any counter, and the garbage is collectable. But it is the same defect shape the repo has fixed before — `internal/router/file_router.go` (~`:416`) already assembles multi-part text with `strings.Builder` and even special-cases the single-part path to avoid the Builder's one extra copy.

## Fix direction

Replace the accumulator with a `strings.Builder` (pre-size with the summed block lengths if convenient). The surrounding per-block `strings.Count`/section re-anchoring is linear-total and can stay.

Worth adding a complexity-guard target for extraction over a fixture with many embedded parts, so the growth curve is pinned the same way the detector/redaction hot paths are — with a non-vacuity floor so the target can't silently no-op.

sectoramen:
Closes #338.

`ProcessEmbeddedMedia` accumulated the combined embedded-media text with `var result string` and `result += block` inside the per-part loop. Go's `string +=` reallocates and re-copies the whole accumulated prefix every iteration, so assembly cost **O(n²) bytes** across n embedded parts.

The quadratic is in *bytes*, not parts, so large blocks dominate — embedded documents and PDFs. A vector-heavy deck is unaffected, since `embedded.SkipTextPipeline` excludes SVG/EMF/WMF/WDP up front.

## Memory vs speed — no trade, it wins on all three axes

Both strategies side by side on 64KB blocks (an embedded document, not a thumbnail):

| n | time: accum → builder | allocated: accum → builder | allocs |
|---|---|---|---|
| 60 | 7.0ms → 1.2ms (**5.8x**) | 120MB → 20MB (**5.9x**) | 62 → 16 |
| 240 | 56.9ms → 2.9ms (**19.9x**) | 1895MB → 80MB (**23.8x**) | 243 → 22 |
| 480 | 200.8ms → 7.0ms (**28.7x**) | 7565MB → 156MB (**48.4x**) | 481 → 25 |

Growth per doubling — time: accum 3.5x, builder 2.4x. Bytes: **accum 4.0x (quadratic), builder 2.0x (linear)**.

At n=480 the accumulator allocated **~240x the size of its own 31MB output**; the Builder allocates ~5x.

**Peak memory is roughly unchanged** — both hold an old and a new buffer at the moment they grow, ~2x the final size. What collapses is **churn**, and 7.6 GB of garbage per document is real collector CPU. Worth contrasting with the attempt at #337, which traded churn for peak and delivered nothing measurable, so it was dropped.

## End-to-end — checked because a fixed quadratic can just move

Real binary, `.docx` files carrying 40/80/160 embedded EXIF-bearing JPEGs:

| binary | n=40 | n=80 | n=160 | growth |
|---|---|---|---|---|
| baseline | 0.135s | 0.261s | 0.611s | 1.93x then **2.34x** |
| this | 0.132s | 0.262s | 0.545s | 1.98x then **2.08x** |

Findings **identical** (1200 / 4800) on both, so the comparison isn't vacuous. The whole scan stays linear and is 11% faster at the largest size — **it did not relocate**. The delta is modest here because an EXIF caption block is ~1.3KB, where the quadratic term is small; the 28.7x case is the large-block one the issue names.

## The guard was vacuous on my first attempt

Asserting only an absolute wall-clock ceiling let the restored accumulator **pass** at 9.5x growth — 70ms is nowhere near any sane ceiling, so the exact defect the test exists for was invisible to it.

It now asserts the **ratio** (threshold 6.0, sitting between the Builder's ~2.9 and the accumulator's ~9.5, with linear at 4.0 and quadratic at 16.0), keeping the ceiling only as a backstop for a proportional regression a ratio cannot see. A ratio is sound here because the base sample is *milliseconds* — unlike the redaction guard, whose base is small enough that only a ceiling is meaningful.

Non-vacuity floors on the guard itself: every part must have been routed, the combined text must be non-empty, and it must grow ≥3.5x for 4x the parts — so an assembly that silently stopped appending fails rather than looking fast.

Correctness is asserted separately from timing, since a fast wrong assembly is what a ceiling cannot catch: every part appears in order, sections are still produced, and each section's `LineOffset` falls inside the combined text. The line cursor is maintained independently of the accumulator (`strings.Count` per block), so a Builder can't shift it — asserted, not assumed.

Not pre-sized: blocks are produced inside the loop, so summing their lengths first would generate each twice. Builder's amortized growth is what removes the quadratic.

---

**Not self-merging** — for @rustic.

### DIFF
--- a/internal/preprocessors/base_metadata_preprocessor.go
+++ b/internal/preprocessors/base_metadata_preprocessor.go
@@ -244,7 +244,24 @@ func (bmp *BaseMetadataPreprocessor) ProcessEmbeddedMedia(originalFilePath strin
 		return "", nil, nil
 	}
 
-	var result string
+	// A Builder, not a string accumulator.
+	//
+	// This was `var result string` with `result += block` inside the per-part loop below.
+	// Go's string += reallocates and re-copies the whole accumulated prefix on every
+	// iteration, so assembling the combined text cost O(n²) BYTES copied across n embedded
+	// parts. The quadratic is in bytes, not parts, so a few dozen multi-hundred-KB blocks
+	// (embedded documents and PDFs, or a raster-image-heavy deck where each image
+	// round-trips the router for metadata text) already costs tens of MB of pure transient
+	// copying and GC churn.
+	//
+	// Transient only — it never affected results or any counter — but it is the same shape
+	// internal/router/file_router.go already fixed with a Builder when it assembles
+	// multi-part text. See #338.
+	//
+	// Not pre-sized: the blocks are produced inside the loop, so summing their lengths up
+	// front would mean generating each one twice. Builder's amortized growth is what
+	// removes the quadratic; a Grow hint would only shave a constant.
+	var result strings.Builder
 	var sections []ContentSection
 	// warnings records embedded items we declined to descend into, so the caller can
 	// surface them through ExtractionWarning.
@@ -324,7 +341,7 @@ func (bmp *BaseMetadataPreprocessor) ProcessEmbeddedMedia(originalFilePath strin
 
 			// Format and append embedded media section
 			block := bmp.utilities.RouterHelper.FormatEmbeddedMediaSection(i, media.OriginalName, processed.Text)
-			result += block
+			result.WriteString(block)
 
 			// FormatEmbeddedMediaSection prefixes "\n--- ... ---\n", so the
 			// item's own text starts two lines further on.
@@ -365,7 +382,7 @@ func (bmp *BaseMetadataPreprocessor) ProcessEmbeddedMedia(originalFilePath strin
 		}
 	}
 
-	return result, sections, warnings
+	return result.String(), sections, warnings
 }
 
 // EmbeddedMedia represents embedded media extracted from documents
--- a/internal/preprocessors/embedded_media_complexity_test.go
+++ b/internal/preprocessors/embedded_media_complexity_test.go
@@ -103,9 +103,44 @@ func timeAssemble(t *testing.T, n, blockSize int) (time.Duration, int, int) {
 
 // TestProcessEmbeddedMediaAssemblyIsNotQuadratic is the guard.
 //
-// The ratio is LOGGED rather than asserted and an absolute ceiling is asserted instead —
-// the same idiom the redaction guard in internal/goldencorpus settled on, because the base
-// sample is small enough that scheduler noise dominates a ratio.
+// It asserts the GROWTH RATIO (see maxGrowth below), with an absolute ceiling kept only as a
+// backstop. An earlier revision of this comment claimed the opposite — that the ratio was
+// merely logged, in the idiom the redaction guard in internal/goldencorpus uses — which was
+// a leftover from a first attempt that was itself vacuous.
+//
+// # Why this drives ProcessEmbeddedMedia directly instead of scanning a real document
+//
+// #338 asks for the growth curve to be pinned "the same way the detector/redaction hot paths
+// are", i.e. over a real fixture. That target was built and measured, twice, and neither
+// version can gate on the defect.
+//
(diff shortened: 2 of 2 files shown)


sectoramen:
## Keeping `Closes #338`, with the guard question settled by measurement

A review pass argued this should be downgraded to a reference, on the grounds that #338's closing ask — *"a complexity-guard target for extraction over a fixture with many embedded parts, so the growth curve is pinned the same way the detector/redaction hot paths are"* — is not met, because the committed guard stubs the router instead of driving a real document.

That was a fair challenge, so I built the target it asks for. Twice. Neither version can gate on the defect.

**Timed end-to-end**, over a real `.docx` carrying n embedded EXIF JPEGs through `core.ScanFile`, the fix is invisible:

| parts | `string +=` | `strings.Builder` |
|---|---|---|
| 40 | 0.26s | 0.25s |
| 160 | 0.72s | 0.71s |
| 320 | 1.38s | 1.31s |
| 640 | 2.56s | 2.49s |

Both linear. The copying is swamped by zip inflate, per-part EXIF extraction, and validating the combined text.

Since #338 describes the defect as *"O(n²) **bytes copied**"*, allocation was the next candidate — `runtime.MemStats.TotalAlloc` across the same real scan. The quadratic does show there, but only just:

| per-part text | with fix | reverted |
|---|---|---|
| 4 KB | 3.9x | 4.4x (299MB → 347MB at 160 parts) |
| 32 KB | 4.0x | 5.0x (1081MB → 1450MB at 160 parts) |

Separating 4.0x from 5.0x needs a threshold with ~12% headroom, on a figure that moves with `GOMAXPROCS` and GC timing across the three CI runners, at ~1.4 GB of churn per run. I wrote that guard, mutation-tested it, watched the mutation **pass**, and deleted it. A gate that flaky is worse than no gate — and shipping one would have been the vacuous-guard failure mode this repo has been bitten by before.

The isolated loop gives a **3.6x margin** instead, re-measured on this branch: **2.7x** with the Builder, **8.6x** with the accumulator restored, which fails the 6.0 assertion. That is why the guard is targeted.

So: the defect #338 reports is fully fixed, and the guard is real rather than decorative. The reviewer's underlying point still stands and is now written into the test's doc comment rather than left implicit — a stub router pins only the append loop, so a quadratic introduced in part enumeration, in `ProcessEmbedded`, or in container-side stitching would need its own target. That is a different defect from the one #338 reports, so it does not belong in the way of closing it.

Also fixed in `2dadebe`: the doc comment claimed *"The ratio is LOGGED rather than asserted and an absolute ceiling is asserted instead"*, contradicting the body twelve lines below it. Leftover from the first, ceiling-only attempt. No production code changed.

rustic (approved):

Issue #336, opened 2026-08-18
Scan accounting doesn't reconcile: the NOT-EXAMINED breakdown drops newer causes, files_skipped counts files total_files excludes, and walk access errors reach no counter at all


sectoramen:
# Summary

A directory scan of ~1,870 files surfaced four related defects in the file-accounting ledger behind the scan summary and the `NOT FULLY EXAMINED` block. Individually each is small; together they mean the summary's numbers cannot be reconciled with each other, one class of refused file is invisible in the per-cause breakdown, and another class (walk access errors) is invisible *everywhere* — including `--fail-on-incomplete`.

Observed output (aggregates only):

```
Skipped 2 files or directories due to errors
Starting scan of 1866 files...
Filtered out 1 unsupported files
Filtered out 168 unsupported file types
[progress] 1698/1698 files - 0 skipped
Scan complete: 1675 files scanned, 169 skipped (unsupported type)
Files: 1649 scanned, 65 NOT examined, 169 skipped
NOT FULLY EXAMINED: 65 of 1867 files — findings may be missing
  cannot parse: 23  ·  no body text (metadata still scanned): 41      <-- 23+41 = 64, header says 65
```

Most of the arithmetic is actually correct by design once decoded (1698 = 1866 − 168; 1675 = 1698 − 23 parse failures; 1649 = 1675 − 26 finding-less empty extractions; the 1866→1867 bump is the documented discovery-unexamined denominator adjustment). The four things below are the parts that are wrong.

## Defect 1 — the collapsed breakdown omits `causeNotFollowed` (and, on PR #332, `causeTooLarge`)

When more than `inlineDetailLimit` (8) files are unexamined, the per-cause tally iterates a **hardcoded list of the four legacy causes**:

- `cmd/unscanned_report.go:374` (main): `for _, c := range []unscannedCause{causeUnreadable, causeUnparseable, causeNoText, causeCutShort}`

The header two lines above prints `len(entries)`, which counts **all** causes. `causeNotFollowed` (refused symlinks, added in #326, appended at `cmd/unscanned_report.go:254`) is tallied in the header but gets no bucket line. PR #332 adds `causeTooLarge` entries into the same slice without widening the loop — so the file silently missing from the breakdown is *exactly the oversize file that #324/#332 exist to disclose*. That is what produced the observed `65` header over `23 + 41 = 64` buckets.

[...]

PR #351, opened 2026-08-18
fix(cli): scope discovery disclosure to what was actually in scope, and stop dropping glob-matched directories

### LINKED ISSUE #342: files_skipped counts files that total_files excludes, so the scan summary can never reconcile
Split out of #336 (defect 2) so the rest of that issue can close with #341. This is the last of its four defects still open.

**No coverage loss** — every file involved is disclosed somewhere. The harm is that the printed numbers cannot be reconciled against each other, so an operator cannot use them to confirm a scan was complete.

## Mechanism

- `cmd/main.go:2119` — `consideredFiles := len(filesToProcess) + len(discoveryUnexamined)` feeds `TotalFiles` (`:2122`) and the `"N of M"` report denominator.
- `cmd/main.go:1922` — `finalSkippedCount := totalSkipped + skippedFiles` feeds `FilesSkipped` (`:2124`).

`totalSkipped` increments per `result.SkippedFiles` entry (`:1524`) and for path-traversal inputs (`:1486`). Those files are **never appended to `FilesToProcess`**, so they never enter `consideredFiles`. The identity breaks by exactly 1 per discovery-time skip.

Minimal repro — a directory with one small `.txt` plus one 200 MB unprocessable file:

```
total_files=1, files_processed=1, files_skipped=1
```

Two files accounted against a total of one.

On the ~1,870-file run that surfaced this: `1649 + 65 + 169 = 1883` against `total_files = 1867`. The gap is 15 (deliberate, documented overlap — no-body-text files that produced metadata findings count as both scanned and not-examined) **plus 1** from this defect.

## Why the existing test cannot catch it

`TestSummaryCountsReconcile` asserts reconciliation on hand-built, non-overlapping fixtures. The real caller hands the formatter categories that overlap (no-body-text) and categories that are exogenous to `total_files` (discovery-time skips), so the fixture can be internally consistent while the pipeline is not.

## Fix direction

Pick one ledger and state it:

1. Either discovery-time skips enter `consideredFiles`, so `total_files` means "every file this run took responsibility for";
2. or they get their own stats slot that the summary line does not add into the same total.

Then add a reconciliation test wired to the **real pipeline counters** — one that scans a fixture directory containing, at minimum: a scannable file, an unsupported-type file, a discovery-time skip, a no-body-text file (the deliberate overlap), and an unexamined file. Assert the documented identity holds including the overlap term, so the term is asserted rather than assumed.

Whichever option is chosen, the identity belongs in a comment next to the computation, because the current numbers are individually correct and only collectively wrong — which is exactly the shape that survives review.

Related: #336, #341, #324.

sectoramen:
## What

Five defects in discovery-time coverage accounting, in **both** directions. Everything refused must be disclosed, and nothing that was never in scope may be reported as a loss. The second half was broken *by* the fixes for the first (`#324`, `#336`/`#341`), so they are corrected together — they need opposite corrections, and testing them apart is how one gets traded for the other.

## Under-disclosure — a real detection miss

A **directory matched by a glob pattern was dropped in silence.** The glob branch tested `info.Mode().IsRegular()` and had no `else`, so a matched directory reached neither `FilesToProcess` nor either disclosure slice. `--recursive` was accepted and did nothing, because recursion lives in the directory branch a glob never reaches.

Measured against a readable tree holding an SSN and a card number:

| | findings | total_files | exit | disclosure |
|---|---|---|---|---|
| before | **0** | 1 | 0 | none |
| after | **2** | 3 | 0 | — |

A shell that expands the pattern hides this, since each match arrives as its own argument. It bites when the pattern is **quoted**, which is what the documentation asks for.

Directories are now walked when `--recursive` is set — by recursing through `getFilesToProcess` itself, so the symlink policy, exclude/`.gitignore` checks and size refusals cannot drift out of step — and reported as skipped, naming `--recursive` as the remedy, when it is not.

A **path named directly on the command line that exists but cannot be read** reached no counter: one stderr line, `No files to process`, `total_files 0`, no `files_not_examined` key, and `--fail-on-incomplete` exited 0. A complete scan of nothing, with an unread SSN at the named path.

A path that does **not** exist is deliberately still not a coverage loss — a typo is a usage error, and failing builds over misspelled arguments is how a flag ends up permanently disabled. `isAccessDenied` is the discriminator, using `Lstat` on the path so the case that matters most (a readable filename inside an unreadable **parent**) classifies correctly.

## Over-disclosure — introduced by `#341`

A permission-denied subdirectory made a **non-recursive** scan exit 3, and made **`--exclude` on that very directory** exit 3 — both reporting lost coverage for something the user never asked to be scanned.

The cause is `filepath.Walk`: it calls `readDirNames` **first** and hands the callback that error in the *same* invocation, so an unreadable directory arrives at the error branch and never reaches the `!recursive` and exclude checks below it. The branch now decides scope before disclosure.

A flag that fires outside the requested scope is one people stop passing, which costs exactly the disclosures it exists to deliver.

## Malformed artifact

`No files to process` went to stdout regardless of `--format`, so `--format json > report.json` produced a document whose first byte is `N` — failing at the exact moment the scan had something to report, since that path is reached when every input was **refused**.

Structured formats now fall through to the ordinary output path, which already emits an empty report carrying `total_files`, `files_not_examined` and the unscanned entries. The pipeline needed no changes to tolerate zero files. The test is `!= "text"` rather than a list of structured formats, so a format added later is parseable by default rather than by remembering to add it here.

## Verification

- All five **mutation-tested**; each mutation compiles and fails a test.
- 66 packages pass; `-race` and `vet` clean on `./cmd`; golden corpus passes forced; `gofmt` clean repo-wide (the exact CI command).
- Old vs new on the repo source tree: **identical 1725 findings**. Text format, pre-commit mode (`rc` 0/1) and `--preprocess-only` (`rc` 2) unchanged. A nonexistent path still exits 0 under `--fail-on-incomplete`.
- An empty directory with `--format json --output` now writes a valid report where it previously wrote **no file at all**.

No file overlap with open PRs `#347`, `#348` or `#350`.

## Issue references

**Refs #353** — partially addresses it. This PR fixes the *forced text* symptom (an explicit `--format` now outranks auto-detected pre-commit defaults). Measured on binaries from both trees, `PRE_COMMIT=1 --format json` on a directory holding one SSN: `main` → 472 bytes of prose, this PR → 837 bytes starting `{"stats":{...`. The rest of #353 is untouched and stays open: the detection heuristic itself (`GIT_EXEC_PATH`/Git Bash), the git-repo-with-hook condition, the missing opt-out, the empty `--output` file, and the `QuietMode`/`NoColor` analogues.

**Refs #342** — this PR *widens* that gap, and it should be fixed there rather than here. #342's mechanism is that every `result.SkippedFiles` entry breaks `files_processed + files_skipped + files_not_examined == total_files` by +1, because those files never enter `consideredFiles = len(filesToProcess) + len(discoveryUnexamined)`. The new glob-matched-directory disclosure lands in that same exogenous ledger. Measured A/B:

| fixture | main | this PR |
|---|---|---|
| glob, 1 matched dir, no `--recursive` | gap +0 | **gap +1** |
| glob, 3 matched dirs | gap +0 | **gap +3** |
| #342's own repro (oversize unprocessable file) | gap +1 | gap +1 (unchanged) |

So the gap now scales with the number of matched directories: the text summary reads `Files: 1 scanned, 3 skipped | Findings: 1` — four files against a total of one. A second new route does the same, and buys real coverage for it: the `--recursive` glob branch propagates `sub.SkippedFiles`, so an oversize file inside a matched subtree also lands in `files_skipped` without entering `total_files` (findings 1 → 2).

This PR's other new counters are identity-**neutral**: the unreadable named-input path and the permission-denied glob match go to `UnexaminedFiles`, which feeds both `total_files` and `files_not_examined` (measured gap +0 before and after), and the new out-of-scope silent return replaces a neutral `UnexaminedFiles` entry with nothing.

Deliberately not fixed here: #342 requires *picking a ledger* and stating it, which changes what `total_files` means for anyone parsing the output, plus a reconciliation test wired to the real pipeline counters including the documented no-body-text overlap term. That belongs in its own change, not bundled into a disclosure fix. A follow-up PR closes #342 generically, which covers these new routes regardless of merge order.

### DIFF
--- a/cmd/discovery_scope_test.go
+++ b/cmd/discovery_scope_test.go
@@ -1,0 +1,315 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// Discovery-time coverage accounting has to get BOTH directions right: everything that was
+// refused must be disclosed, and nothing that was never in scope may be reported as a loss.
+//
+// The first direction was fixed for #324 and #336. The second was broken BY those fixes, and
+// the two failures need opposite corrections, so they are tested together to keep a later
+// change from trading one for the other.
+
+// pii is a value that must be found when a path is genuinely scanned, and must never appear
+// in a disclosure string.
+const pii = "SSN: 452-11-9384"
+
+// A DIRECTORY matched by a glob pattern used to be dropped in silence.
+//
+// The glob branch tested info.Mode().IsRegular() and had no else, so a matched directory
+// reached neither FilesToProcess nor either disclosure slice. --recursive was accepted and
+// did nothing, because recursion lives in the directory branch that a glob never reaches.
+//
+// Measured on the shipped binary against a readable tree holding two values:
+//
+//	--file '<dir>/*' --recursive   ->  0 findings, total_files 1, exit 0, no disclosure
+//	after this change              ->  2 findings, total_files 3
+//
+// A shell that expands the pattern itself hides this, because each match arrives as its own
+// argument. It bites when the pattern is QUOTED, which is what the documentation asks for.
+func TestGlobMatchedDirectoryIsScannedWhenRecursive(t *testing.T) {
+	root := t.TempDir()
+	deep := filepath.Join(root, "deep", "deeper")
+	if err := os.MkdirAll(deep, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(root, "top.txt"), []byte("nothing here\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(root, "deep", "pii.txt"), []byte(pii+"\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(deep, "more.txt"), []byte("Card: 4111111111111111\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	res, err := getFilesToProcess(filepath.Join(root, "*"), true, nil, nil, true)
+	if err != nil {
+		t.Fatalf("getFilesToProcess: %v", err)
+	}
+
+	var sawNested, sawDeepest bool
+	for _, p := range res.FilesToProcess {
+		switch filepath.Base(p) {
+		case "pii.txt":
+			sawNested = true
+		case "more.txt":
+			sawDeepest = true
+		}
+	}
+	if !sawNested || !sawDeepest {
+		t.Errorf("FilesToProcess = %v, want the files inside the matched directory. A glob that "+
+			"matches a directory dropped it entirely, so --recursive was silently a no-op and "+
+			"the PII inside was never scanned — which also means it is never redacted.",
+			res.FilesToProcess)
+	}
+}
+
+// Without --recursive the directory is genuinely out of scope, so it must be REPORTED as
+// skipped rather than either scanned or dropped.
+//
+// Dropped is what it used to be: no counter, no message, nothing for the operator to notice
+// that a matched path contributed nothing.
+func TestGlobMatchedDirectoryIsDisclosedWhenNotRecursive(t *testing.T) {
+	root := t.TempDir()
+	sub := filepath.Join(root, "sub")
+	if err := os.MkdirAll(sub, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(sub, "pii.txt"), []byte(pii+"\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(root, "top.txt"), []byte("nothing\n"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	res, err := getFilesToProcess(filepath.Join(root, "*"), false, nil, nil, true)
+	if err != nil {
+		t.Fatalf("getFilesToProcess: %v", err)
+	}
+
+	for _, p := range res.FilesToProcess {
+		if strings.Contains(p, "sub") {
+			t.Errorf("%q was scanned without --recursive", p)
+		}
+	}
+
+	var found *SkippedFile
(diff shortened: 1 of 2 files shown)


sectoramen:
## Windows CI failure diagnosed and fixed (commit `3ebaef6`)

The `windows-latest` job failed the new "`--format json` must be parseable when every input is refused" test, with stdout `"No files to process\n"`. The cause was **not** the change under test.

On Windows, pre-commit mode is **auto-detected**, and it then overwrote the requested format with its own default. `detectWindowsGitEnvironment()` treats **`GIT_EXEC_PATH`** as a pre-commit signal, and Git Bash always sets it — the CI step runs under `C:\Program Files\Git\bin\bash.EXE`. It also fires whenever the working directory is a git repo carrying a pre-commit hook, which the test's own package directory is. So nothing had to opt in.

This is a real user-facing bug, not a test artifact. Measured with `PRE_COMMIT=1` on a directory holding one SSN, before the fix:

| invocation | stdout |
|---|---|
| `--format json` | 1258 bytes of valid JSON |
| `--format json`, mode detected | `pii.txt: 1 high confidence issues found` |

An explicit request for a machine artifact returned prose, at a success-shaped exit. Every Windows user running from Git Bash got that.

**Fix:** an explicitly-set `--format` outranks the auto-detected default. The guard is narrow and costs nothing — pre-commit's text default is already supplied upstream by the `precommit` profile in `resolveConfiguration`, so the assignment was a *second* application of an already-correct value. I verified by mutation that disabling it entirely is unobservable; I kept it guarded rather than deleting it, because "unobservable in the cases I could construct" is not "provably redundant".

**Tests:** the new case reproduces the defect on **every** platform via `PRE_COMMIT=1` rather than depending on Windows, and pins the other half (pre-commit still overrides a format set in *config* — which needs a config file to observe at all, since the ordinary default is `text` too). The refused-inputs test now runs through a helper that defeats detection — a temp working directory outside any repository plus emptied trigger variables — because it asserts ordinary output and would otherwise measure pre-commit behaviour on one platform only.

**Filed separately as #353**, not fixed here: the detection heuristic itself (Git Bash is not a hook invocation), and that in pre-commit mode an explicitly requested `--output` receives an **empty** file when there are no findings — the same "asked for a machine artifact, got something unparseable" shape.

All six changes in this PR are mutation-tested. `gofmt`/`vet` clean, 66 packages pass, `-race` clean on `./cmd`, golden corpus passes forced.

rustic (approved):

PR #354, opened 2026-08-18
fix(cli): count discovery-time skips in total_files so the summary reconciles

### LINKED ISSUE #342: files_skipped counts files that total_files excludes, so the scan summary can never reconcile

sectoramen:
Closes #342.

## The defect

Every `result.SkippedFiles` entry increments `totalSkipped` and feeds `files_skipped`, but none of those paths is ever appended to `filesToProcess` — so they were absent from `consideredFiles = len(filesToProcess) + len(discoveryUnexamined)`, which is `total_files`.

#342's own repro, one small `.txt` plus one oversize unprocessable file:

| | total_files | files_processed | files_skipped | gap |
|---|---|---|---|---|
| before | 1 | 1 | 1 | **+1** |
| after | 2 | 1 | 1 | **0** |

The gap scaled — three such files gave +3. No coverage is lost; the harm is that an operator cannot use the summary to confirm a scan was complete.

## The identity is now written down next to the computation, and asserted

```
files_processed + files_skipped + files_not_examined
    == total_files + overlapMetaOnly + overlapCutShort
```

#342 asked for one ledger to be picked and stated. Both correction terms are **deliberate** double counts, where one file legitimately occupies two categories because it was both examined and incompletely examined:

- **`overlapMetaOnly`** — a body-empty file whose *metadata* produced a surviving finding. Counted as scanned (it produced findings) and as not-examined (its body was unread). This is the one the `silentEmpty` comment already documented.
- **`overlapCutShort`** — a file whose validator coverage was cut short by a budget, cancellation, or match-budget stop. The worker pool leaves `Result.Error` nil for these, so the file lands in `ProcessedFiles` **and** in `IncompleteFiles`.

### The second term is not in #342, and it was documented nowhere

It breaks the naive sum with **no discovery-time skip involved at all**. One file under `--validator-budget all=1ns` measures `total_files=1, files_processed=1, files_not_examined=1`. It is not a defect and is not "fixed" here — it is stated, so a future reader doesn't delete it as an off-by-one.

Two things deliberately **absent** from the identity, both verified rather than assumed:

- unreadable and failed-processing files sit in `total_files` and `files_not_examined` but in **neither** `files_processed` nor `files_skipped`. They need no correction term — they are what makes the sum close. So an assertion of the form `files_processed + files_skipped == total_files` would be wrong.
- the `scannedFiles` clamp cannot fire: `silentEmpty` counts a subset of `emptyExtractionFiles`, every one of which is already counted in `processedFiles`.

The identity is over **entries**, not distinct paths — `collectUnscanned` does not deduplicate, so a file that is both body-empty and cut short contributes two entries.

## Why the existing test couldn't catch it

`internal/formatters/text.TestSummaryCountsReconcile` builds `ScanStats` **by hand** and asserts `processed + notExamined + skipped <= total`, failing with *"the caller must not hand the formatter overlapping categories"* — while the real caller hands it overlapping categories deliberately. The fixture was internally consistent while contradicting the pipeline it stood in for. It still passes and is untouched.

The counters only interact in `cmd/main.go`, so the new test drives the **real binary** and states each fixture's deliberate overlap count *by construction*, which asserts the term rather than assuming it. Six shapes plus the cut-short case: plain files, one skip, three skips, one and two metadata-only files, and both sources at once.

## Verification

- Three mutations tested, each compiling and failing a test: dropping the new term, double-counting it, and adding the **wrong** counter (`skippedFiles`, which is already inside `filesToProcess`).
- 66 packages pass; `-race` clean on `./cmd`; `gofmt` clean repo-wide; the pre-existing formatter test still passes.
- **Cross-checked against open PR #351**, which adds two more `result.SkippedFiles` routes: on its branch a glob-matched directory without `--recursive` measures gap **+1**; with this change merged it is **0**. The ledger fix is generic, so it covers those routes whichever merges first. Union of both branches: 66 packages pass, `gofmt` clean.
- The test defeats pre-commit auto-detection explicitly (temp cwd outside any repo, trigger variables emptied). Without that it would measure pre-commit's forced text format on Windows only — the failure mode #351 hit.

## Noted, not fixed here

- `cmd/main.go` has an unreachable `scanMalfunction` branch: `ProcessFilesWithProgress` has a single return and it is always `nil`.
- The directory walk records `--exclude`/`.gitignore` refusals nowhere at all (plain `return nil`), so in a recursive scan they reach no counter — distinct from the skip ledger this PR corrects.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -2108,15 +2108,57 @@ func main() {
 		scannedFiles = 0
 	}
 
-	// consideredFiles is every entry this run took responsibility for: the files it
-	// queued to scan PLUS the ones discovery refused. The denominator has to include
-	// both, or the report reads "2 of 1 file" — the refused entries appear in the
-	// numerator while being absent from the total they are counted against.
+	// consideredFiles is every entry this run took responsibility for: the files it queued
+	// to scan, PLUS the ones discovery refused as a coverage loss, PLUS the ones discovery
+	// SKIPPED. The denominator has to include all three, or the report reads "2 of 1 file" —
+	// entries appear in the numerator while being absent from the total they are counted
+	// against.
 	//
-	// Derived once and used by the stats block AND the not-examined report below, for
-	// the same reason the entries themselves are collected once: two independent counts
-	// of the same thing eventually disagree.
-	consideredFiles := len(filesToProcess) + len(discoveryUnexamined)
+	// totalSkipped was the missing term (#342). Every result.SkippedFiles entry increments it
+	// and none of them are ever appended to filesToProcess, so a run with one small .txt plus
+	// one oversize unprocessable file printed total_files=1, files_processed=1,
+	// files_skipped=1 — two files accounted against a total of one, off by exactly one per
+	// discovery-time skip. Measured before this change: gap +1 for one such file, +3 for three.
+	//
+	// # The identity, which is asserted rather than assumed
+	//
+	//	files_processed + files_skipped + files_not_examined
+	//	    == total_files + overlapMetaOnly + overlapCutShort
+	//
+	// The two correction terms are DELIBERATE double counts, not defects: one file legitimately
+	// occupies two categories, because it was both examined and incompletely examined.
+	//
+	//   - overlapMetaOnly: a file whose body yielded no text but whose METADATA produced a
+	//     surviving finding. Counted as scanned (it produced findings) and as not-examined (its
+	//     body was not read). See the silentEmpty comment above: only the genuinely silent ones
+	//     are removed from the scanned count.
+	//
+	//   - overlapCutShort: a file whose validator coverage was cut short — a per-validator
+	//     budget, a cancellation, or a match-budget stop. The pool leaves Result.Error nil for
+	//     these, so the file is counted in ProcessedFiles AND listed in IncompleteFiles.
+	//     Measured: one file under --validator-budget all=1ns gives total_files=1,
+	//     files_processed=1, files_not_examined=1.
+	//
+	// This term is NOT mentioned in #342 and was not documented anywhere; it breaks the naive
+	// sum with no discovery-time skip involved at all, which is why the reconciliation test
+	// asserts the identity WITH its overlap terms rather than asserting a plain sum.
+	//
+	// Two things deliberately absent from the identity:
+	//   - unreadable and failed-processing files are in total_files and in files_not_examined
+	//     but in NEITHER files_processed nor files_skipped. They need no correction term; they
+	//     are what makes the sum close. An assertion of the form
+	//     "files_processed + files_skipped == total_files" is therefore wrong.
+	//   - the scannedFiles clamp above cannot fire: silentEmpty counts a subset of
+	//     emptyExtractionFiles, every one of which is already counted in processedFiles.
+	//
+	// The identity is over ENTRIES, not distinct paths. collectUnscanned does not deduplicate,
+	// so a file that is both body-empty and cut short contributes two entries — which is why
+	// both terms are counted independently rather than as a set union.
+	//
+	// Derived once and used by the stats block AND the not-examined report below, for the same
+	// reason the entries themselves are collected once: two independent counts of the same
+	// thing eventually disagree.
+	consideredFiles := len(filesToProcess) + len(discoveryUnexamined) + totalSkipped
 
 	formatterOptions.Stats = &formatters.ScanStats{
 		TotalFiles:       consideredFiles,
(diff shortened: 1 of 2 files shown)


sectoramen:
## Windows failure fixed (not a counting bug)

`TestCutShortCoverageIsTheOtherOverlapTerm` failed on `windows-latest` with `stats artifact is not parseable JSON: unexpected end of JSON input`. The cause is environmental, not arithmetic.

Git Bash sets **`MSYSTEM`**, `MINGW_PREFIX` and `GIT_EXEC_PATH`; the Windows test step runs under `C:\Program Files\Git\bin\bash.EXE`; and this test's isolation list scrubbed only the last of the three. So pre-commit mode stayed live, and the json formatter *deliberately* returns an empty document in that mode when there are no findings (`internal/formatters/json/formatter.go:63`) — which is exactly this fixture, since the validator budget stops every validator before a finding is produced. `--output` therefore received **zero bytes**.

Reproduced locally by building a binary with the platform branch forced on:

| environment | `--output` file |
|---|---|
| `MSYSTEM` set (Git Bash) | **0 bytes** |
| `MSYSTEM` emptied | 266 bytes |

The counters and the identity are unaffected — no production code changed here, and the reconciliation assertions are untouched.

The list is deliberately **not** a copy of the shared `precommitFreeEnv()` helper in #356: defining that function in two files of one package would compile fine on each branch and fail on the union. Once #356 lands this becomes a call to it, and the list is then covered by that PR's drift guard, which derives the trigger set from `internal/precommit/detector.go` — a guard that immediately caught a third variable (`PSModulePath`) my hand-written grep had missed.

sectoramen:
## Windows: this PR's own failure is fixed; the two that remain are main's

Latest run on `e6677a1`:

- ✅ `TestCutShortCoverageIsTheOtherOverlapTerm` — **now passes** (the `MSYSTEM`/`MINGW_PREFIX` scrub added in this PR).
- ❌ `TestExplicitFormatOutranksPrecommitDetection` and `TestMachineFormatEmitsValidJSONWhenEveryInputIsRefused` — these live in `cmd/discovery_scope_test.go`, which **this PR does not touch**. They are red on `main` itself, having been merged with #351 before the missing trigger was known.

PR CI builds the *merge* of this branch with `main`, which is why main's file appears in this run at all: this branch's merge-base is `fefc0ee`, main is `efc4884`, and `cmd/discovery_scope_test.go` does not exist on this branch.

**#356 fixes those two**, and its `windows-latest` job is **green**. So the merge order is:

1. **#356 first** — it unreds `main`.
2. then this PR, whose Windows job goes green on its next run with **no change needed here**, since CI re-tests the merge.

Verified locally: merging #356 into `main` makes all three tests pass together, and merging both this PR and #356 into `main` compiles, passes `./cmd`, and is `gofmt` clean — there is no duplicate-helper collision between them (that is why the scrub list here is deliberately inline rather than a copy of `precommitFreeEnv`).

No production code in this PR changed as part of the Windows work, and none of the reconciliation assertions were touched.

rustic (approved):

Issue #315, opened 2026-08-14
Score-gate sink coverage still excludes four corpus cases that now redact fine, and a file with findings but no redacted copy still exits 0 with no recorded decision


sectoramen:
The scanning side and the redaction side admit different sets of file types, and the gap is widening. A type in that gap produces findings and **no redacted copy**, at exit 0.

## The asymmetry

The plaintext **redactor** claims 11 extensions (`internal/redactors/plaintext/redactor.go:82`):

```go
return []string{"text", ".txt", ".log", ".csv", ".json", ".xml", ".yaml", ".yml", ".md", ".conf", ".ini"}
```

The plaintext **preprocessor**, since #310, claims essentially any file whose bytes look like text — deliberately, and correctly, because an extension allowlist will always lag the set of text formats that exist.

So #310 widened detection without widening redaction. Verified end to end on the current `main`:

```
$ ferret-scan --file diagram.svg --checks SSN,EMAIL
[HIGH] SSN 90%   ...

$ ferret-scan --file diagram.svg --enable-redaction --redaction-output-dir out
rc=0
WARNING: redaction incomplete — 1 of 1 file(s) have findings but no redacted copy was
written; the original values remain in cleartext:
  diagram.svg: failed to get redactor: no redactor registered for file type: .svg
files written: 0
```

This is not a new class of problem — the score corpus already names four such cases:

```
no redactor registered for these file types (a disclosed CLI limitation, not a leak):
  [c04_tsv c45_tsv_no_header s09_html_table s11_sql_insert]
```

so `.tsv`, `.html` and `.sql` were already in the gap. #310 added `.svg`, `.swift`, `.kt`, `.tf`, `.tfvars`, `.properties`, `.jsx`, `.tsx`, `.dart`, `.proto`, `.gradle` and more — each confirmed to produce findings and no redacted copy.

`.tf`/`.tfvars` and `.properties` are the costly ones: they are among the likeliest places on disk to hold a live credential, which is precisely why #310 added them.

## Why this is a bug and not just a limitation

The behaviour is **disclosed** — the warning above is real, tested (`cmd/unredacted_warning_test.go`) and golden-covered — so this is not a silent leak, and that is why the severity is moderate rather than high. Two things still make it a defect:

[...]

PR #357, opened 2026-08-18
feat(redactors): add an audio metadata redactor so reported audio findings can be removed

### LINKED ISSUE #306: Audio metadata findings are reported but no redactor handles audio, so --enable-redaction exits 0 with the value still in the file
## Summary

Audio files are scanned and their metadata findings are reported, but **no redactor handles audio**, so those findings can never be removed. `--enable-redaction` produces an output file that still contains the value, at exit 0, with nothing saying otherwise.

A reported-but-unredactable value is the same leak as an undetected one, dressed as a success.

## Evidence

The read side is fully built out — there is a dedicated preprocessor:

```
internal/preprocessors/audio_metadata_preprocessor.go:239   case ".mp3":
internal/preprocessors/audio_metadata_preprocessor.go:241   case ".flac":
internal/preprocessors/audio_metadata_preprocessor.go:245   case ".m4a":
internal/preprocessors/shared_utilities.go:132              ".mp3": true,
```

Audio is also admitted as an **embedded** part, so it is reached inside containers too (`embeddedMediaType` returns `"audio"` for `.mp3`, `.wav`, `.m4a`, `.flac`).

The write side has nothing. Every registered redactor's supported types, checked:

| redactor | claims |
|---|---|
| `office` | `.docx` `.xlsx` `.pptx` |
| `legacyole` | `.doc` `.xls` `.ppt` |
| `image` | `.bmp` `.gif` `.jpeg` `.jpg` `.png` `.tif` `.tiff` `.webp` |
| `pdf` | `.pdf` |
| `plaintext` | `.txt` `.log` `.csv` `.json` `.xml` `.yaml` `.yml` `.md` `.conf` `.ini` |

```console
$ grep -o '"\.\(mp3\|wav\|m4a\|flac\)"' internal/redactors/*/*.go
# no matches
```

So `RedactionManager.GetRedactorForFile` cannot resolve a redactor for any audio extension.

## Why it matters

ID3 and MP4/M4A metadata carry author, artist, comment, copyright and free-text description fields — exactly the shape the engine reports as `AUTHOR_INFO` and friends, and exactly where a name, email or phone number ends up in a voice memo or recorded call.

The failure is **silent**: the finding is named in the report, redaction is requested, the run exits 0. `RedactionCount` counts *findings*, not replacements (`internal/core/redact.go:148`), so it is not usable as evidence either way.

## Scope

Independent of container nesting. It applies to a standalone `.mp3` on the command line, and #305's nested-redaction loop will hit the same wall from the other direction — when that loop asks the manager for an audio redactor it will get none, and that outcome has to be **disclosed** rather than silently skipped.

## Two ways to close it

1. **Disclose it** (small, and needed regardless). When no redactor can handle a file that produced findings, surface it through the existing `stats.UnredactedFiles` path so the run cannot report success. `internal/core/redact.go:118-130` already carries the guard and the reasoning for exactly this case; it just is not reached here.
2. **Write an audio redactor** (complete). Strip or overwrite the offending ID3 / MP4 atom fields, mirroring what the `image` redactor does for EXIF. That redactor is the closest existing model — same shape of problem, metadata-only, simple strategy.

(1) should probably land first and on its own, since it stops the false success for *any* unhandled type, not just audio.

## Note

Found while auditing embedded-part parity for #305. The parity table there records this row; this issue exists because the gap is not about nesting.

sectoramen:
Closes #306. Refs #315.

## The gap

Audio was scanned but had no redactor. The read side is complete — a dedicated preprocessor, four extractors, and audio admitted as an embedded part — so a voice memo's ID3 comment holding a phone number was found, scored and printed. Then `--enable-redaction` wrote **no output file and exited 0**.

Reproduced across all four scanned formats before this change:

| | findings | redacted copy | exit |
|---|---|---|---|
| `.wav` / `.mp3` / `.flac` / `.m4a` | 5 each (SSN, PHONE, AUTHOR_INFO, …) | **none** | **0** |

The *disclosure* half of #306 turned out to be already wired and firing (`redaction incomplete — 1 of 1 file(s) have findings but no redacted copy was written`), so this PR adds the missing half.

## Design: same-length in-place overwrite, confined to tag regions

Every one of these containers stores tag text behind a length — a RIFF chunk size, an ID3v2 synchsafe tag size, a FLAC 24-bit block length, an MP4 atom size. Changing a value's length means rewriting every enclosing size, and in MP4 also every sample offset in `stco`, because moving bytes moves the audio. A corrupt file that *looks* redacted is worse than an honest refusal, so the replacement is always exactly `len(original)` bytes. Same technique as `legacyole`. `Synthetic` is declined — it cannot preserve length.

Overwrites are confined to **located metadata regions** rather than applied file-wide. A whole-file search would also hit coincidental occurrences in the audio samples: an 11-byte SSN is unlikely to appear in PCM, but `"Jane Doe"` in a 40 MB file is not.

## The bug found while verifying — the interesting part

The first working version removed the SSN, phone and email from all four formats and **left "Jane Doe" in cleartext at exit 0**.

Reported matches **overlap**. The scan reports both:

```
SSN          "452-11-9384"
AUTHOR_INFO  "Contact Jane Doe SSN 452-11-9384"
```

Replacing them sequentially against the buffer being written loses the second: masking the SSN destroys the tail of the AUTHOR_INFO string, the search for it then finds nothing, and the part reported *only* as AUTHOR_INFO — the name — stays. This is the mechanism of #191.

**The fail-closed residue check was fooled the same way**: it searched the output for each reported value, and the AUTHOR_INFO string was genuinely absent because its own substring had been altered. The verification agreed with the bug.

Sorting longest-first fixes *containment* but not *partial* overlap. So every occurrence is now located in the **original** bytes before anything is written, and overlapping spans are merged and masked — a format-preserving rendering is only meaningful for one value of one type.

## Also: both UTF-16 byte orders

ID3v2 declares UTF-16LE (with a BOM) or UTF-16BE (`0x02`, no BOM) per frame, and a value stored in the order not searched is a value left behind. Found *by* mutation testing: disabling the residue check wasn't detectable, and asking why exposed that neither the planner nor the check knew about big-endian.

## Verification

Corpus built with **ffmpeg** (real containers, real tag structures) across all four formats plus unicode, long-value, title-only and no-metadata controls, verified with **exiftool as an independent parser** rather than by grepping bytes:

| fixture | PII after | plays (ffprobe) | size |
|---|---|---|---|
| `plain.wav` / `.mp3` / `.flac` / `.m4a` | **clean** | yes | identical |
| `unicode.mp3`, `longval.flac` | **clean** | yes | identical |

A real `.m4a` on a dev machine reports 0 findings and correctly receives no redacted copy, and its `udta` atom is located (275 bytes at offset 198304).

### Correction: the "Jane Doe" residue is not a detection gap

An earlier revision of this description (and the commit message, which is already pushed and so cannot be amended) claimed that `"Jane Doe"` surviving in an M4A title was a PERSON_NAME detection gap. **That was wrong**, and the fixture was at fault rather than the validator.

`internal/validators/personname/validator.go:640` carries a deliberate test-data list — `"john doe"`, `"jane doe"`, `"foo bar"`, `"lorem ipsum"`, … — worth **−50 confidence**, documented in `help.go` as the "Not Test Data" criterion. `"Jane Doe"` is the canonical placeholder, so it is intentionally driven below the reporting threshold. Only findings reach the redactor, so it is correctly left alone. Suppressing it is the right call: reporting it would flood documentation, templates and test fixtures with false positives.

Verified with a non-placeholder name, and the suppression is not audio-specific:

| fixture | PERSON_NAME |
|---|---|
| `control_placeholder.txt` (plain text, "Jane Doe") | none |
| `control_real.txt` (plain text, "Marcus Whitfield") | 92 |
| `t_placeholder.m4a` (title) | none |
| `t_real.m4a` (title) | **92** |
| `a_real.m4a` (artist) | **92** + AUTHOR_INFO |
| `t_real.mp3` / `a_real.mp3` | **92** |

Plain text and audio metadata behave identically, so there is nothing audio-specific to fix and no issue to file.

End to end, a real name **is** removed:

```
t_real.m4a   before: Recording for Marcus Whitfield, ref 452-11-9384
             after : Recording for ****************, ref ***-**-9384
a_real.m4a   after : ***********************************************
```

The artist case masks the whole field because AUTHOR_INFO covers the entire value and overlaps both PERSON_NAME and the SSN — the span merge described above, working as intended.

Tests use **hand-built containers**, not ffmpeg output, so they don't depend on ffmpeg being on a CI runner.

**12 mutations tested, 11 fail a test.** The 12th — removing the MP4 depth bound — is deliberately uncovered and the comment says why: termination is already guaranteed because every level consumes at least 8 bytes of a finite buffer, so a fixture that could distinguish it would be measuring Go's stack growth. Range boundaries are asserted by **exact offset**, not just "the value was removed", because a range that is too large still removes the value while handing the overwrite permission over audio frames — the synchsafe mutation is caught precisely this way.

67 packages pass (up from 66), `-race` clean on redactors + core, golden corpus forced, `gofmt`/`vet` clean.

## Scope

`#315`'s remaining half — the plaintext redactor's 11-extension allowlist against a byte-sniffing preprocessor — is untouched, so this references rather than closes it.

No file overlap with open PRs `#354` or `#356`.

### DIFF
--- a/internal/core/redact.go
+++ b/internal/core/redact.go
@@ -15,6 +15,7 @@ import (
 	"github.com/awslabs/ferret-scan/v2/internal/observability"
 	"github.com/awslabs/ferret-scan/v2/internal/parallel"
 	"github.com/awslabs/ferret-scan/v2/internal/redactors"
+	"github.com/awslabs/ferret-scan/v2/internal/redactors/audio"
 	"github.com/awslabs/ferret-scan/v2/internal/redactors/image"
 	"github.com/awslabs/ferret-scan/v2/internal/redactors/legacyole"
 	"github.com/awslabs/ferret-scan/v2/internal/redactors/office"
@@ -212,6 +213,7 @@ func NewDefaultRedactionManager(outputDir string, strategy redactors.RedactionSt
 		officeRedactor,
 		legacyole.NewLegacyOLERedactor(outputManager, observer),
 		image.NewImageMetadataRedactor(outputManager, observer),
+		audio.NewAudioRedactor(outputManager, observer),
 	} {
 		if err := manager.RegisterRedactor(r); err != nil {
 			return nil, nil, fmt.Errorf("failed to register redactor %s: %w", r.GetName(), err)
--- a/internal/redactors/audio/overlap.go
+++ b/internal/redactors/audio/overlap.go
@@ -1,0 +1,215 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package audio
+
+import (
+	"bytes"
+	"sort"
+
+	"github.com/awslabs/ferret-scan/v2/internal/detector"
+	"github.com/awslabs/ferret-scan/v2/internal/redactors"
+)
+
+// occurrence is one span of the file to overwrite, with the bytes to write there.
+type occurrence struct {
+	start int
+	end   int
+	repl  []byte
+	wide  bool // the value was found UTF-16LE encoded, so a mask must keep that width
+	count int  // how many matches contributed to this span, after merging
+}
+
+// planOverwrites computes every span to overwrite, resolving overlaps.
+//
+// # Why this cannot be a sequential search-and-replace
+//
+// Reported matches OVERLAP. On a real .wav the scan reports both
+//
+//	SSN          "452-11-9384"
+//	AUTHOR_INFO  "Contact Jane Doe SSN 452-11-9384"
+//
+// Replacing them one at a time against the buffer being modified loses the second one:
+// masking the SSN destroys the tail of the AUTHOR_INFO string, so searching for
+// AUTHOR_INFO next finds nothing, is recorded as "not located", and "Jane Doe" — which is
+// only reported as part of that longer value — stays in the file. Measured before this
+// existed: SSN, phone and email removed from all four formats, "Jane Doe" left in cleartext,
+// exit 0.
+//
+// Worse, a residue check that searches the OUTPUT for each reported value is fooled the same
+// way: the AUTHOR_INFO string is genuinely absent, because its own substring was altered. So
+// the verification agreed with the bug.
+//
+// Sorting longest-first is not sufficient either. It fixes CONTAINMENT but not PARTIAL
+// overlap: for "…SSN 452" and "452-11-9384" the first masks only the head of the second, and
+// the untouched tail is still cleartext. This is the same mechanism as #191 — a destructive
+// sequential replace over partially overlapping matches.
+//
+// So every occurrence is located in the ORIGINAL bytes, before anything is written, and
+// overlapping spans are merged into one. A merged span is masked rather than
+// format-preserved, because it covers parts of two values of different types and no single
+// format-preserving rendering is correct for both.
+// It also reports, per match, how many occurrences of that match were located, so the caller
+// can record a mapping only for values it actually wrote over — never one it merely tried.
+func planOverwrites(orig []byte, ranges []byteRange, matches []detector.Match, strategy redactors.RedactionStrategy) (plan []occurrence, perMatch []int) {
+	var found []occurrence
+	perMatch = make([]int, len(matches))
+
+	for mi, m := range matches {
+		if m.Text == "" {
+			continue
+		}
+		repl := sameLengthReplacement(m.Text, m.Type, strategy)
+
+		narrow := []byte(m.Text)
+		narrowRepl := []byte(repl)
+		// Both UTF-16 byte orders: ID3v2 declares UTF-16LE (with a BOM) or UTF-16BE (0x02,
+		// no BOM) per frame, and a value stored in the order not searched is a value left
+		// behind.
+		wideEncodings := []struct {
+			pat  []byte
+			repl []byte
+		}{}
+		for _, enc := range []func(string) []byte{toUTF16LE, toUTF16BE} {
+			pat := enc(m.Text)
+			if len(pat) == 0 {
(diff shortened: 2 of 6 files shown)


sectoramen:
## Correction to the commit message

The commit message states that `"Jane Doe"` surviving in an M4A title is a PERSON_NAME detection gap. **That is wrong.** The commit is already pushed so the message cannot be amended; the PR description now carries the correct account, and this comment records it.

`internal/validators/personname/validator.go:640` holds a deliberate test-data list (`"john doe"`, `"jane doe"`, `"foo bar"`, `"lorem ipsum"`, …) carrying a **−50 confidence** penalty, documented in `help.go` as the "Not Test Data" criterion. `"Jane Doe"` is the canonical placeholder, so it is intentionally pushed below the reporting threshold and therefore never reaches the redactor. That is correct behaviour — reporting it would flood documentation and test fixtures with false positives.

My corpus was the problem: I used the canonical placeholder as test PII. With a real name the behaviour is right, and identical in plain text and audio metadata (so nothing audio-specific):

```
control_placeholder.txt  PERSON_NAME none      t_placeholder.m4a  none
control_real.txt         PERSON_NAME 92        t_real.m4a         92
                                               a_real.m4a         92 + AUTHOR_INFO
                                               t_real.mp3         92
```

And it is redacted end to end:

```
before: Recording for Marcus Whitfield, ref 452-11-9384
after : Recording for ****************, ref ***-**-9384
```

No issue filed, because there is no defect. The redactor's behaviour in this PR is unchanged by the correction — only my explanation of one fixture's output was wrong.

rustic (approved):

Issue #367, opened 2026-08-19
Failed-extraction log writes the scanned file path to stderr, bypassing ScanConfig.LogWriter


sectoramen:
## Summary

A failed media/EXIF extraction logs a message containing the **full path of the scanned file**, and that write never consults `ScanConfig.LogWriter`. An embedder that sets `LogWriter: io.Discard` specifically to guarantee no payload reaches a log still gets the path emitted on a process fd.

A filename is frequently the most sensitive string in a scan — a real corpus file here is named `c1-SSN-449-87-4100.jpg`, and users routinely name documents after the person or ID they concern.

## Where

`internal/preprocessors/error_handling.go:321` (v2.3.2):

```go
// STDERR, not stdout.
fmt.Fprintf(os.Stderr, "%s\n", message)
```

`message` is built at `:56` via `fmt.Sprintf("media processing failed for %s", mpe.FilePath)`.

The surrounding comment explains the stdout→stderr move (so `--format json` stdout stays valid JSON) — that part is a good fix. The issue is separate: **neither fd is `LogWriter`**.

## Reproducer

Scanning through `pkg/scan` with `LogWriter: io.Discard` and capturing fd 2:

```
[ERROR] media processing failed for /tmp/.../b-SSN-449-87-4100.jpg
  type=image error=extraction_failed message=no EXIF data found: EOF
  cause=no EXIF data found: EOF context=[file_extension=.jpg suggestion=EXIF metadat...]
```

359 bytes on stderr, path included, with `LogWriter` set to discard.

**Trigger:** an image *extension* over non-image *content* (e.g. a `.jpg` holding text). A merely corrupt EXIF/APP1 segment does **not** trigger it on v2.3.2 — worth knowing, because it means a fixture built the obvious way tests nothing.

## Why this matters beyond cosmetics

The write moved fds between releases: it was `fmt.Printf` (stdout) up to `v2.3.1-0.20260812030445`, and stderr from v2.3.2. Downstream mitigations that redirect a single fd silently stop working across an engine bump — which is exactly what happened to us, and we only noticed because a test asserted *which* fd leaks.

For embedders — mobile apps, desktop agents — fd 1/2 can be a terminal, a CI log, a parent process, or a platform log collector. `LogWriter` is the documented control for this and it is bypassed on this path.

## Suggested fix

[...]

PR #393, opened 2026-08-20
fix(office): close three ways a value hides from Office redaction — escaped, numeric-referenced, and numeric-typed

### LINKED ISSUE #368: Office redaction matches entity-decoded text against still-escaped part bytes, so a value containing & or written as &#48; survives --enable-redaction while the audit log records 0 failed
## What happens

extractTextFromXML (internal/redactors/office/redactor.go:623) reads part text through `encoding/xml`, so the text handed to the validators is entity-DECODED. The rewrite at redactor.go:848 then runs `strings.NewReplacer(args...).Replace(string(xmlContent))` over the RAW part bytes. Any value whose stored form differs from its decoded form is therefore never found. The parent document has no residue check at all: `embedded.XMLEscapeVariants` has exactly one caller, internal/redactors/office/embedded.go:196, which covers embedded children only.

## Measured evidence (HEAD 61a85bd; `git diff --stat 2a0e96c 61a85bd -- internal/redactors/office/` is empty)

Three cases, all rc=0, all output dirs `rm -rf`'d first.

(1) Apostrophe, default flags. `word/document.xml` holds `Patient name: Patrick O&apos;Connor` plus a control `SSN: 449-87-4100`.
```
ferret --file /tmp/a1/escaped.docx --show-match --confidence all
ferret --file /tmp/a1/escaped.docx --enable-redaction --redaction-output-dir /tmp/a1/out6
unzip -p /tmp/a1/out6/tmp/a1/escaped.docx word/document.xml
```
Scan: `[HIGH] PERSON_NAME 100.00% Patrick O'Connor` and `[HIGH] ssn SSN 100.00% 449-87-4100`. Redacted part: `SSN: ***-**-4100` but `Patient name: Patrick O&apos;Connor` VERBATIM. Word renders the full name.

(2) Ampersand, in the SAME part as a redacted sibling — this isolates the cause. `docProps/app.xml` holds `<Application>Microsoft Office Word</Application><Company>Fairbanks &amp; Kettleworth Holdings</Company>`; the scan reports `[LOW] metadata COMPANY_INFO 55.00%`. After redaction the part reads `<Application>*********************</Application><Company>Fairbanks &amp; Kettleworth Holdings</Company>`. Independent oracle, exiftool 13.55 on the redacted copy: `Company : Fairbanks & Kettleworth Holdings` / `Application : *********************`.

(3) Universal numeric-character-reference bypass (xlsx). Cell B1 inline string `449-87-41&#48;0`, B2 `512-44-8321`, B3 `4532-0151-1283-0366`. Scan reports three HIGH 100% findings including `SSN 449-87-4100`. After redaction `sheet1.xml` still holds `449-87-41&#48;0` untouched while B2/B3 mask correctly, and parsing the redacted sheet with Python ElementTree (what Excel sees) yields cell values `['SSN','449-87-4100','Backup SSN','***-**-8321','Card','****-****-****-0366']`.

False attestation, verbatim: `--redaction-audit-log` on case (3) wrote `"successful_redactions": 3, "failed_redactions": 0` and a content_redactions entry `{"data_type":"SSN","redacted_text":"***-**-4100"}` for the value that was never touched.

## Why it matters (sink rule)

A reported value survives --enable-redaction at rc=0 with Success:true and zero failed redactions. This needs no attacker: XML *requires* `&` in character data to be written `&amp;`, so every Word document whose body text or Company/Title property contains an ampersand is affected. The numeric-reference form generalises it — a reference can encode any character at any offset, so any value can be hidden from the replacer. The fail-closed refusal that exists for embedded parts never fires, because the parent document is never checked for residue.

## Fix sketch

1. Stop string-matching escaped bytes: rewrite the part through `encoding/xml` (walk tokens, replace CharData, re-emit) so the codec owns escaping. Offering `embedded.XMLEscapeVariants(v)` as extra Replacer keys is a cheaper interim that covers the `&amp;`/`&apos;` half only — it cannot cover `&#NN;`, which needs canonicalisation.
2. Add a parent-document residue check with the same fail-closed policy the embedded path already has (internal/redactors/office/embedded.go:196), comparing after entity decoding rather than on raw bytes, so a missed rewrite is refused instead of attested as success.

## Failing test

internal/redactors/office: build a .docx whose `w:t` holds `Fairbanks &amp; Kettleworth Holdings` and an .xlsx whose inline string holds `449-87-41&#48;0`; call RedactDocument with matches carrying the DECODED text; re-parse the rewritten part with `encoding/xml` and assert the decoded CharData no longer contains the match text. Both assertions fail at HEAD. A second test should assert RedactDocument returns an error (fail closed) when any parent part still holds a reported value, mirroring the embedded-part refusal. Note that `grep -n 'EscapeText|amp;|apos;|xml.Escape' internal/goldencorpus/ooxml_builder_test.go internal/redactors/office/*_test.go` returns nothing today, so no existing fixture can contain an escaped value.

Related but distinct: the extraction-side mirror of this problem (numeric character references never decoded on the docx/pptx/odf path) is filed separately as defect E18.

### LINKED ISSUE #371: Numeric character references are decoded on the xlsx path only, so the same value written as &#53; in a .docx is reported nowhere and survives --enable-redaction
## What happens

The correct decoder already exists: `decodeXMLEntities` (internal/preprocessors/text-extractors/text-extract-officetextlib/office-text-extractor.go:831) handles `&lt; &gt; &quot; &apos; &amp;` AND `&#NNN;` / `&#xHHH;` in one single pass, and has its own unit test (xlsx_extraction_test.go:202). Its only three call sites are on the XLSX path. Three other paths use a hand-rolled five-call `strings.Replace` chain with no numeric-reference handling: `extractDocxText` (:238-242), `extractTextFromXML` (:690-694, which serves pptx/odt/ods/odp) and `extractWordXMLText` (:1165-1169, docx headers/footers).

## Measured evidence (HEAD)

/tmp/e-batch5/docx/ncr.docx, four paragraphs each `Employee SSN: <value>`:
```
ferret --file /tmp/e-batch5/docx/ncr.docx --preprocess-only
ferret --file /tmp/e-batch5/docx/ncr.docx --confidence all --show-match
```
`--preprocess-only` shows the raw text reaching the validators verbatim: `Employee SSN: 51&#57;-42-8836`. The scan reports 1 finding — only the plain `449-87-4100`. `519-42-8836` (one decimal reference), `563-18-7249` (every digit a reference) and `607-31-9284` (hex reference) produce ZERO findings from every validator at `--confidence all`. That is a 75% miss.

Control, byte-identical references on the other path: /tmp/e-batch5/docx/ncr.xlsx inline strings `--preprocess-only` print `Employee SSN: 519-42-8836` already decoded, and the scan reports SSN 100. The identical evasion is caught in xlsx and invisible in docx.

## Why it matters (sink rule)

After `--enable-redaction`, rc=0, verified by `unzip -p ... word/document.xml` (not by grepping the container), the `w:t` runs are:
```
Employee SSN: ***-**-4100
Employee SSN: 51&#57;-42-8836
Employee SSN: &#53;&#54;&#51;-&#49;&#56;-&#55;&#50;&#52;&#57;
Employee SSN: 60&#x37;-31-9284
```
Three of four SSNs survive in a file the tool reports as successfully redacted, and Word renders every one of them as the real digits, so the document a human opens shows the SSN.

## Fix sketch

Replace the three hand-rolled `strings.Replace` chains (office-text-extractor.go:238-242, :690-694, :1165-1169) with a single call to the existing `decodeXMLEntities`. That also fixes a second latent bug those chains have and which decodeXMLEntities's own doc comment warns about: sequential replaces re-read their own output, so `&amp;lt;` becomes `<` instead of the literal `&lt;`.

## Failing test

A docx analogue of the existing xlsx entity test: build a .docx whose `w:t` holds `Employee SSN: 51&#57;-42-8836`, assert ExtractText yields `519-42-8836`, and assert a scan reports SSN. Fails today. Add a golden corpus case `docx_numeric_character_reference` with its `.redact_simple` snapshot so the sink is locked the way `passport_label_above_value` locks its own.

Related but distinct and in a different file: the redaction-side mirror (matching decoded text against still-escaped part bytes, so `&#48;` also defeats the rewriter) is filed separately as defect A1. Fixing only one direction leaves the other leak open.

### LINKED ISSUE #373: A custom property whose value has no letter is dropped as bookkeeping, and a vt:i4 property is never redacted, so a numeric member or account id in docProps/custom.xml stays cleartext
## Part 1: letterless values are never scored

`isValuelessProperty` (internal/validators/metadata/metadata_validator.go:2940-2990) returns true for ANY value containing no ASCII letter (`if !hasLetter { return true }` at :2961), and the CUSTOM_PROPERTY branch skips such values before scoring (:1229). The documented rationale — "A value made only of digits, hex, and structural punctuation has no prose in it to disclose" — is deliberate, but it swallows identifier-length digit runs.

Measured on /tmp/e-batch5/docx/custprops.docx, 7 custom properties, `--checks METADATA --confidence all`:
- reported: `REF-449871234`, `Marcus Whitfield`
- NOT reported: `449871234` (CaseNumber), `449871234` (RecordId, vt:i4), `4158924471` (ContactPhone), `4532015112830366` (AccountDigits), `449-87-4100` in a property literally named **SubscriberSSN**

`--preprocess-only` shows all seven values ARE extracted, so this is a validator skip, not an extraction gap. Sharpest A/B, /tmp/e-batch5/docx/numonly.docx: `Custom_ControlText` = `Ledger 8291746350284` -> CUSTOM_PROPERTY 60; `Custom_BillingRef` = `8291746350284` -> nothing. Same digits, same part, same mechanism; one English word is the whole difference.

## Part 2: vt:i4 is never rewritten even when a validator DOES report it

internal/redactors/office/redactor.go:560-584 `docPropsValueElements` lists only lpwstr/lpstr/bstr. /tmp/e-batch5/docx/i4only.docx (single property `<vt:i4>729183640</vt:i4>`): the scan reports SSN 50; after `--enable-redaction`, `unzip` shows `<vt:i4>729183640</vt:i4>` byte-identical, rc=0, no warning. The same value in `<vt:lpwstr>` redacts to `*****3640`.

## Why it matters (sink rule)

numonly.docx after `--enable-redaction`, rc=0, verified by `unzip -p ... docProps/custom.xml`:
```
<vt:lpwstr>923456781</vt:lpwstr>
<vt:lpwstr>8291746350284</vt:lpwstr>
<vt:i4>729183640</vt:i4>
<vt:lpwstr>********************</vt:lpwstr>   <- the one letter-bearing value
```
Severity is bounded and worth stating honestly: when the digits happen to be SSN/phone/card-shaped, another validator claims them and format-preserving redaction masks all but the last four. The full escape needs a numeric value that also dodges every other validator, demonstrated above with a 9-digit 9xx value and a 13-digit run. Part 2 is the sharper half, because there the finding IS reported and the rewrite is silently skipped at rc=0.

## Fix sketch

1. In `isValuelessProperty`, replace the blanket `!hasLetter -> valueless` with a shape test: keep bare short integers, integer tuples, GUIDs and ISO timestamps valueless, but admit an unbroken digit run of identifier length (>= 7 digits, optionally with `-` or spaces) as a candidate. That test's own measurement (119 documents, 1,228 findings, 488 Purview MSIP_Label bookkeeping rows) is the false-positive budget to re-measure against.
2. Add i4/i8/ui4/r4/r8/decimal/int/uint to `docPropsValueElements` in internal/redactors/office/redactor.go, or fall back to a raw value-substring replace for any docProps element carrying a reported match, so a reported finding can never be silently skipped.

## Failing test

internal/validators/metadata/value_shape_risk_test.go already owns the guard: extend `TestValuelessPropertiesAreSkipped`'s `meaningful` slice with `"923456781"` and `"8291746350284"` (fails today) while keeping `"0"` and `"50, 3, 0, 1"` in `valueless`. For part 2, add an office-redactor test asserting a `<vt:i4>` property with a reported finding is masked; it fails today.

Distinct from #307 (CUSTOM_PROPERTY scoring HIGH on the substring "confidential") and #267 (legacy-OLE vector properties): both of those concern the letter-bearing/extraction paths, not value admission or docProps element types.

sectoramen:
Closes #368

Two commits: the rewrite that closes the leak, then the residue check that makes a future miss a refusal instead of a false attestation.

## The defect

`extractTextFromXML` reads part text through `encoding/xml`, so the value handed to the validators — and named in the report — is entity **decoded**. The rewrite then ran `strings.NewReplacer` over the **raw** part bytes, so a value whose stored spelling differs from its decoded form never occurred literally, the replacer was a no-op for it, and `RedactDocument` still returned success.

Measured on a `.docx` whose Company property holds an ampersand:

| | |
|---|---|
| scan | `[LOW] metadata COMPANY_INFO  Fairbanks & Kettleworth Holdings` |
| after redaction | `<Company>Fairbanks &amp; Kettleworth Holdings</Company>` |
| **exiftool on the redacted copy** | `Company : Fairbanks & Kettleworth Holdings` |
| audit log | `"successful_redactions": 4, "failed_redactions": 0` |
| exit code | `0` |

A sibling value in the same part (`dc:creator`, no entity) masked correctly, which isolates the cause to the spelling rather than to part selection.

This needs no attacker: XML **requires** `&` in character data to be written `&amp;`. An apostrophe does it too — `Patrick O&apos;Connor`, reported at HIGH 100, came back verbatim.

## Why enumerating escaped spellings cannot fix it

`embedded.XMLEscapeVariants` offers the `&amp;`/`&apos;` forms as extra replacer keys, covering the five predefined entities and nothing else. `&` also introduces character references, so any character at any offset can be respelled. Verified against `encoding/xml` that all of these decode to the same value:

```
449-87-41&#48;0    449-87-41&#x30;0    &#52;49-87-4100    449&#45;87&#45;4100
```

The spellings are combinatorial, so only canonicalising through the codec is complete. On an `.xlsx` inline string holding `449-87-41&#48;0`: before, the cell was reported as `SSN 449-87-4100` at HIGH 100 and left untouched, so **Excel read the real SSN out of the redacted workbook**; after, it reads `***-**-4100`.

## Strict superset, not a swap

`rewritePartText` matches each character-data span on its decoded form and re-emits it escaped. Everything **outside** character data — markup and attribute values — plus any character data that did not change still goes through the **same raw replacer** the old code used, over the byte ranges between rewrites.

So every replacement the old code made is still made. That matters concretely: a value in an attribute is not character data, and a character-data-only rewrite would have stopped redacting it — turning a leak fix into a new leak. A test pins it.

A tokenizer error is not fatal. `encoding/xml` rejects an undefined entity like `&foo;`, and such a part falls through to the raw replacer for its remainder, exactly as before.

`escapeCharData` escapes only `&`, `<`, `>` — not `xml.EscapeText`, which also rewrites quotes and turns newlines into `&#xA;` — so a redaction diff shows the redaction and nothing else. `&` is escaped **unconditionally**: `&amp;lt;` decodes to a literal `&lt;`, so an escaper that skipped `&` before a known entity name would silently corrupt that value.

## Verification

- **67 packages ok, 0 failed.** `gofmt -l ./cmd ./internal ./pkg` clean, `go vet` clean.
- **The golden corpus passes with zero regenerated files**, so the rewrite is byte-identical on every existing case. That is the strict-superset claim measured, not asserted.
- **Mutation-tested, 5 mutations, all caught:** revert to the raw whole-part `Replace` (7 subtests fail); drop the raw pass over the gap regions (attribute + untokenizable-part tests fail); make `escapeCharData` skip `&` when it looks like an entity (round-trip fails); revert the rewrite with the residue check present (the redactor now **refuses and writes 0 files**); remove the residue early return (the anti-quadratic guard fails).
- **Cost**, A/B against the pre-fix binary on a `.docx` where *every* paragraph rewrites through the new path (worst case):

  | matches | before | after |
  |---|---|---|
  | 500 | 0.08 s / 35 MB | 0.08 s / 35 MB |
  | 1000 | 0.13 s / 36 MB | 0.14 s / 38 MB |
  | 2000 | 0.25 s / 40 MB | 0.25 s / 41 MB |
  | 4000 | 0.50 s / 48 MB | 0.51 s / 53 MB |

  Linear in both, 2.0× per doubling. The replacer is still built **once per part**; building it per token would be quadratic in the value count.

## A quadratic caught and removed before it shipped

The residue check's first version did one `strings.Contains` per value per part: +0.00s, +0.01s, +0.02s, **+0.09s** as the value count doubled through 500/1000/2000/4000 — about 4.5× per doubling. It is now two stages, a trie probe that decides presence in one pass and a per-value identification loop reserved for the refusal path. Re-measured flat: +0.00s, +0.00s, +0.01s.

The guard for it is a **counter**, because both obvious alternatives are blind: allocation cannot see it (the quadratic allocates nothing — an allocation-ratio test *passed* with the quadratic restored, i.e. it was vacuous), and wall clock is not portable to the Windows runner.

## Notes

- The process still exits 0 on a refusal. Propagating that to the exit code is the open residual on #315 and is deliberately untouched here.
- #371 is the same entity-handling class in the **opposite** direction (extraction never decodes numeric references on the `.docx` path). Fixing one leaves the other open; they are cross-referenced.

---

# Second half added: #371 — the same evasion on the READ side

Closes #371

`&` introduces every reference form, so the escaped-value problem has two directions. #368 (above) is the write side: match the decoded form against the escaped part bytes. #371 is the read side: **decode the references at all**, on the three Office text paths that never did.

The correct decoder already existed — `decodeXMLEntities`, with its own unit test — and only the xlsx path used it. Three other paths kept a hand-rolled five-call `strings.Replace` chain with no numeric-reference handling: `extractDocxText` (the body), `extractTextFromXML` (pptx slides/notes/masters, odt, ods, odp) and `extractWordXMLText` (docx headers and footers).

```
.docx   1 of 4 SSNs reported      <- 75% miss
.xlsx   4 of 4 at HIGH 100        <- same references, byte-identical values
```

A header is where a template stamps the account or case number on every page, so that path is not exotic.

## Why the two halves are one PR

Measured on the same fixture at four points, residue read from the **decoded** part text:

| | detection | readable in the "redacted" copy |
|---|---|---|
| `main` | 1 of 4 | 3 |
| **#368 alone** (this PR before today) | 1 of 4 | 3 |
| **#371 alone** | **4 of 4** | **3 — file written, no warning** |
| **both** | **4 of 4** | **none** |

Row three is the reason. Fixing detection alone converts a silent miss into a report that says *four SSNs were found and handled* while Word still renders three of them — actively misleading rather than merely incomplete. A unit test on either side passes in that state; only the pipeline shows it, which is what the new integration test drives.

It also removes a golden-snapshot ordering hazard: a `.redact_*` snapshot for this fixture is only correct when both halves are present, so landing them separately would have made whichever merged second go red on `main`.

## The issue's second claim, corrected

#371 also says the chains re-read their own output, with `&amp;lt;` becoming `<`. That **does not reproduce** — the chain runs lt, gt, amp, quot, apos, so `&lt;` is replaced before the `&amp;` step can create one. The bug is real for the entities *after* `&amp;`:

```
&amp;quot;  ->  "     (should be the literal &quot;)
&amp;apos;  ->  '     (should be the literal &apos;)
&amp;#57;   ->  &#57; (correct — and must NOT become "9", which would invent a digit)
```

Both directions are now pinned, including the invent-a-digit direction, which is the same mistake pointing the other way.

## Coverage

- One test per call site, each **mutation-tested by restoring its own chain** — three mutations, three catches.
- A fourth mutation makes `decodeXMLEntities` itself sequential; the corrected-claim test catches it.
- The decoder's own guard: a reference naming no valid code point (`&#0;`, `&#xD800;`, `&#x110000;`) is left as written rather than becoming U+FFFD, which would be unsearchable and unredactable.
- Golden `FileCase file_docx_numeric_character_reference` locks detection at 4 of 4 through `ScanFile` and the real routing. FileCases carry **no** redaction snapshot (0 of the 54 in the corpus belong to a file case), so the sink is locked by the integration test — which is also the only place the two halves compose. Both halves are mutation-tested there: reverting the extractor fails on the count, reverting the redactor fails on the residue.
- `BuildDOCXWithRawRuns` is new because `BuildDOCX` escapes its paragraphs — right for every case about a document's *content*, wrong for the one case whose subject **is** the escaping.

## Cross-PR verification

- No file shared with any other open PR (#394–#397, #403).
- Union of `main` + all six open PRs: **builds clean, `go vet` clean** (which compiles every `_test.go`, so the three independent binary-builder helpers in `tests/integration` do not collide), **69/0**, gofmt clean.
- Behavioural matrix, 8 binaries × 8 fixtures: effects compose **additively**, no PR suppresses or doubles another's. On the union, `b_numrefs.docx` goes 1 → 4 findings with **zero** decoded residue and `w:t` runs reading `***-**-8836`.

---

# Third issue added: #373 — a numeric custom property

Closes #373

`&` was one way an Office value hides from redaction; being a **number** is another. Both halves of #373 are here, and they compose the same way #368 and #371 do — either alone leaves the tool worse off than both.

## Part 1 — a letterless value was never scored

`isValuelessProperty` treated **any** value with no ASCII letter as bookkeeping. The sharpest measured pair:

```
Custom_ControlText = "Ledger 8291746350284"  ->  CUSTOM_PROPERTY 60
Custom_BillingRef  = "8291746350284"         ->  nothing at all
```

Same digits, same part; one English word was the whole difference. A 9-digit case number, a 13-digit billing reference and `449-87-4100` under a property literally named `SubscriberSSN` reached no validator at all.

**The false-positive budget, measured before choosing the rule.** 150 real Office documents from this machine — 117 carrying `docProps/custom.xml`, 908 custom properties, 193 letterless values:

| count | shape |
|---|---|
| 98 | bare integers ≤ 4 digits — Purview `MSIP_Label_*_ContentBits` |
| 89 | integer tuples (`11, 2, 1, 0`) — Purview `MSIP_Label_*_Tag` |
| **2** | unbroken runs of 7-8 digits |
| 4 | one 5-6 digit run, one date-ish, two bracket literals |

The rule admits **2 of 193**, and **0 of the 187 Purview rows**. End to end over the same 150 documents: **1956 → 1958 findings, 2 documents changed**, identical at the default confidence. The two are a 7-digit value under `db_document_id` (a document identifier — the point) and a 7-digit sort key under `Order` (one false positive across 150 documents).

It tests the **value's** shape, not the property's **name**: a name vocabulary would need maintaining and would still miss a digit run under `Ref` or `Field3`. A test asserts `isValuelessProperty`'s signature so nobody later turns it into a keyword list without noticing that invalidates the measurement.

## Part 2 — a `vt:i4` was never rewritten

A property named `MemberId` or `RecordId` is routinely written as `<vt:i4>`. The redactor extracted only the string-typed docProps elements, so no replacement was ever registered for the part:

| build | `<vt:i4>729183640</vt:i4>` after `--enable-redaction` |
|---|---|
| `main` | **byte-identical, rc 0, no warning** — silent leak |
| this PR before today | **NO OUTPUT** — refused, and the refusal disclosed |
| now | `<vt:i4>*****3640</vt:i4>` |

The middle row matters and is worth being explicit about: **the residue check earlier in this PR had already closed the leak** — but it did so by making the document unredactable, which cost the redaction of every *other* value in it. On the issue's own mixed fixture, `main` left three numeric values in cleartext and masked one; this PR then refused the whole file; now all four are masked:

```
main   923456781 | 8291746350284 | 729183640 | ********************
now    ********* | ************* | *****3640 | ********************
```

Every **scalar** value type is listed, not a hand-picked few, because the failure mode is a list missing the next type someone uses. Numeric does not mean harmless — an account, member, patient or case number is a number, a date of birth is a date — but which of them is *sensitive* stays the validators' decision: nothing in that map creates a finding, so a wider list cannot over-redact. `bool` and the binary families are excluded deliberately, and a test pins that a value in a `vt:blob` is **refused** rather than rewritten, because a same-length replacement inside base64 produces invalid base64.

## And a defect in this PR itself, found by measuring it

The residue refusal listed the residual **values** in its error — and that error reaches stderr and every machine format with **no `--show-match`**:

```
refusing to write blobbed.docx: 2 reported value(s) still present in the document's own
parts: 449-87-4100, SSN 449-87-4100 payload
```

A guard added to protect a value was handing it out. It now reports `(types: CUSTOM_PROPERTY, SSN)`. Same rule as the rest of the tool: a matched value stays out of a debug log, a scanned path stays out of a failed-extraction line (#367), and the audio and video redactors name a match's **type** when they cannot locate it. The test asserts the operator-visible error, not just the helper — asserting the helper alone let a mutation restoring the value-joining message survive.

## Verification for this half

- **67 packages ok, 0 failed**; `gofmt`, `go vet`, `-race` clean. Golden corpus and score baseline **unchanged** — no corpus case carries a letterless custom property.
- **Union of `main` + all seven open PRs: builds clean, vet clean, gofmt clean, 69/0.**
- **Mutations, all caught**: revert to blanket letterless-is-valueless; drop the digit floor to 4 (which would admit every Purview `ContentBits`); widen the unbroken pattern to 4+; widen the grouped separator set to `,.:/` (which would admit versions and slash-dates); drop the plain-date exclusion; reject grouped forms (which loses the `SubscriberSSN` case); restore the value-joining refusal message; drop `i4`, `i8`, `int`, `ui4`, `r8` or `decimal` from the element list; drop untyped matches from the type list; leave the type list unsorted.
- One mutation **survived and was right to**: an up-front `ContainsAny(".,:/")` guard turned out to be **unreachable** — no string containing one of those can match either anchored pattern — so it was deleted rather than left claiming to matter. The exclusions are still asserted as behaviour.

### DIFF
--- a/internal/goldencorpus/corpus.go
+++ b/internal/goldencorpus/corpus.go
@@ -642,6 +642,24 @@ var FileCases = []FileCase{
 		EnablePreprocessors: true,
 	},
 	{
+		Name: "file_docx_numeric_character_reference",
+		Description: "Tier 3: a .docx whose w:t runs spell SSN digits as XML numeric character references (&#57; and &#x37;). " +
+			"Word renders them as the digits, so this locks BOTH halves of the escaped-value leak: the extractor must decode " +
+			"them (#371) and the redactor must remove them from the part bytes (#368). Before the extractor fix the scan " +
+			"reported 1 of 4; before the redactor fix 3 of 4 survived --enable-redaction in a file reported as successfully " +
+			"redacted. Diff against file_docx_body_and_metadata, which is the same shape with no references.",
+		Checks:   []string{"SSN", "METADATA"},
+		Filename: "numrefs.docx",
+		Content: BuildDOCXWithRawRuns("Jane Analyst", "Ops Reviewer", []string{
+			"Employee SSN 449-87-4100 on file.",
+			"Employee SSN 51&#57;-42-8836 on file.",
+			"Employee SSN &#53;&#54;&#51;-&#49;&#56;-&#55;&#50;&#52;&#57; on file.",
+			"Employee SSN 60&#x37;-31-9284 on file.",
+		}),
+		Tier1Parity:         false, // container extraction has no content-mode equivalent
+		EnablePreprocessors: true,
+	},
+	{
 		Name:        "file_xlsx_sheet_cells",
 		Description: "Tier 3 control: an .xlsx with PII in worksheet cells at the conventional sheet part name.",
 		Checks:      []string{"SSN", "CREDIT_CARD", "METADATA"},
@@ -904,6 +922,20 @@ func BuildDOCX(creator, lastModifiedBy string, paras []string) []byte {
 	return buildOOXML(docxParts("word/document.xml", creator, lastModifiedBy, paras))
 }
 
+// BuildDOCXWithRawRuns is BuildDOCX with the paragraph text inserted VERBATIM into <w:t>.
+//
+// BuildDOCX escapes its paragraphs, which is right for every case whose subject is the document's
+// content. It is wrong for a case whose subject is the ESCAPING itself: a paragraph written
+// "51&#57;-42-8836" would be stored as "51&amp;#57;-42-8836", i.e. the literal five characters
+// "&#57;", and the case would test nothing. Word renders a numeric character reference as the
+// character it names, so the reference form is what a real evasion looks like on disk.
+//
+// The caller is therefore responsible for the argument being valid XML character data. It exists
+// for exactly one kind of case; anything about a document's content should use BuildDOCX.
+func BuildDOCXWithRawRuns(creator, lastModifiedBy string, rawParas []string) []byte {
+	return buildOOXML(docxPartsRaw("word/document.xml", creator, lastModifiedBy, rawParas))
+}
+
 // BuildDOCXWithMainPart is BuildDOCX with control over the main part's NAME, so a
 // case can lock what happens when the document body is not at the conventional
 // path. It also points the package relationship at that name, as a real producer
@@ -924,6 +956,36 @@ func docxParts(mainPart, creator, lastModifiedBy string, paras []string) []ooxml
 	for _, p := range paras {
 		body.WriteString(`<w:p><w:r><w:t xml:space="preserve">`)
 		body.WriteString(escapeXML(p))
+		body.WriteString(`</w:t></w:r></w:p>`)
+	}
+	doc := xmlDecl + `<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"><w:body>` +
+		body.String() + `</w:body></w:document>`
+
+	contentTypes := xmlDecl + `<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">` +
+		`<Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>` +
+		`<Default Extension="xml" ContentType="application/xml"/>` +
+		`<Override PartName="/` + mainPart + `" ContentType="application/vnd.openxmlformats-officedocument.wordprocessingml.document.main+xml"/>` +
+		`<Override PartName="/docProps/core.xml" ContentType="application/vnd.openxmlformats-package.core-properties+xml"/>` +
+		`</Types>`
+
+	rels := xmlDecl + `<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">` +
+		`<Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="` + mainPart + `"/>` +
+		`<Relationship Id="rId2" Type="http://schemas.openxmlformats.org/package/2006/relationships/metadata/core-properties" Target="docProps/core.xml"/>` +
+		`</Relationships>`
+
+	return []ooxmlPart{
+		{"[Content_Types].xml", contentTypes},
+		{"_rels/.rels", rels},
+		{"docProps/core.xml", coreProps(creator, lastModifiedBy)},
+		{mainPart, doc},
+	}
+}
+
+func docxPartsRaw(mainPart, creator, lastModifiedBy string, paras []string) []ooxmlPart {
+	var body strings.Builder
+	for _, p := range paras {
+		body.WriteString(`<w:p><w:r><w:t xml:space="preserve">`)
+		body.WriteString(p) // verbatim: see BuildDOCXWithRawRuns
 		body.WriteString(`</w:t></w:r></w:p>`)
 	}
 	doc := xmlDecl + `<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"><w:body>` +
--- a/internal/goldencorpus/testdata/golden/file_docx_numeric_character_reference.csv
+++ b/internal/goldencorpus/testdata/golden/file_docx_numeric_character_reference.csv
@@ -1,0 +1,7 @@
+Filename,Type,Confidence Level,Confidence %,Line Number,Text
+<TMPDIR>/numrefs.docx,SSN,HIGH,100.0,1,449-87-4100
+<TMPDIR>/numrefs.docx,SSN,HIGH,100.0,3,519-42-8836
+<TMPDIR>/numrefs.docx,SSN,HIGH,100.0,5,563-18-7249
+<TMPDIR>/numrefs.docx,SSN,HIGH,100.0,7,607-31-9284
+<TMPDIR>/numrefs.docx,AUTHOR_INFO,MEDIUM,60.0,1,Jane Analyst
+<TMPDIR>/numrefs.docx,LAST_MODIFIED_BY,MEDIUM,60.0,2,Ops Reviewer
--- a/internal/goldencorpus/testdata/golden/file_docx_numeric_character_reference.gitlab-sast.json
+++ b/internal/goldencorpus/testdata/golden/file_docx_numeric_character_reference.gitlab-sast.json
@@ -1,0 +1,186 @@
+{
+  "remediations": [],
+  "scan": {
+    "analyzer": {
+      "id": "ferret-scan",
+      "name": "Ferret Scan",
+      "url": "https://github.com/awslabs/ferret-scan",
+      "vendor": {
+        "name": "Amazon Web Services"
+      },
+      "version": "0.0.0-development"
+    },
+    "end_time": "<TIMESTAMP>",
+    "scanner": {
+      "id": "ferret-scan",
+      "name": "Ferret Scan",
+      "url": "https://github.com/awslabs/ferret-scan",
+      "vendor": {
+        "name": "Amazon Web Services"
+      },
+      "version": "0.0.0-development"
+    },
+    "start_time": "<TIMESTAMP>",
+    "status": "success",
+    "type": "sast"
+  },
+  "version": "15.0.4",
+  "vulnerabilities": [
+    {
+      "category": "sast",
+      "confidence": "High",
+      "description": "Ferret Scan detected social security number in this file.\n\n**Location:** <TMPDIR>/numrefs.docx (line 1)\n**Confidence:** HIGH (100.0%)\n**Detected by:** ssn validator\n\n**Context:**\n```\nEmployee SSN 449-87-4100 on file.\n```\n\n**Additional Information:**\n- context_impact: 45\n- source: preprocessed_content\n\n**Remediation:**\nRemove Social Security Numbers from code and documentation. Use test data or anonymized identifiers instead.",
+      "id": "ferret-<HASH>",
+      "identifiers": [
+        {
+          "name": "Ferret Scan Check Type",
+          "type": "ferret_scan_check_type",
+          "value": "SSN"
+        },
+        {
+          "name": "Ferret Scan Validator",
+          "type": "ferret_scan_validator",
+          "value": "ssn"
+        }
+      ],
+      "location": {
+        "end_line": 1,
+        "file": "numrefs.docx",
+        "start_line": 1
+      },
+      "message": "Social Security Number detected: [HIDDEN] (confidence: HIGH)",
+      "name": "Social Security Number Detected",
+      "severity": "Critical"
+    },
+    {
+      "category": "sast",
+      "confidence": "High",
+      "description": "Ferret Scan detected social security number in this file.\n\n**Location:** <TMPDIR>/numrefs.docx (line 3)\n**Confidence:** HIGH (100.0%)\n**Detected by:** ssn validator\n\n**Context:**\n```\nEmployee SSN 519-42-8836 on file.\n```\n\n**Additional Information:**\n- context_impact: 45\n- source: preprocessed_content\n\n**Remediation:**\nRemove Social Security Numbers from code and documentation. Use test data or anonymized identifiers instead.",
+      "id": "ferret-<HASH>",
+      "identifiers": [
+        {
+          "name": "Ferret Scan Check Type",
+          "type": "ferret_scan_check_type",
+          "value": "SSN"
+        },
+        {
+          "name": "Ferret Scan Validator",
+          "type": "ferret_scan_validator",
+          "value": "ssn"
+        }
+      ],
+      "location": {
+        "end_line": 3,
+        "file": "numrefs.docx",
+        "start_line": 3
+      },
+      "message": "Social Security Number detected: [HIDDEN] (confidence: HIGH)",
+      "name": "Social Security Number Detected",
+      "severity": "Critical"
+    },
+    {
+      "category": "sast",
+      "confidence": "High",
+      "description": "Ferret Scan detected social security number in this file.\n\n**Location:** <TMPDIR>/numrefs.docx (line 5)\n**Confidence:** HIGH (100.0%)\n**Detected by:** ssn validator\n\n**Context:**\n```\nEmployee SSN 563-18-7249 on file.\n```\n\n**Additional Information:**\n- context_impact: 45\n- source: preprocessed_content\n\n**Remediation:**\nRemove Social Security Numbers from code and documentation. Use test data or anonymized identifiers instead.",
+      "id": "ferret-<HASH>",
+      "identifiers": [
+        {
+          "name": "Ferret Scan Check Type",
+          "type": "ferret_scan_check_type",
+          "value": "SSN"
+        },
+        {
+          "name": "Ferret Scan Validator",
+          "type": "ferret_scan_validator",
+          "value": "ssn"
+        }
+      ],
+      "location": {
+        "end_line": 5,
+        "file": "numrefs.docx",
+        "start_line": 5
+      },
+      "message": "Social Security Number detected: [HIDDEN] (confidence: HIGH)",
+      "name": "Social Security Number Detected",
+      "severity": "Critical"
+    },
+    {
+      "category": "sast",
+      "confidence": "High",
+      "description": "Ferret Scan detected social security number in this file.\n\n**Location:** <TMPDIR>/numrefs.docx (line 7)\n**Confidence:** HIGH (100.0%)\n**Detected by:** ssn validator\n\n**Context:**\n```\nEmployee SSN 607-31-9284 on file.\n```\n\n**Additional Information:**\n- context_impact: 45\n- source: preprocessed_content\n\n**Remediation:**\nRemove Social Security Numbers from code and documentation. Use test data or anonymized identifiers instead.",
+      "id": "ferret-<HASH>",
+      "identifiers": [
+        {
+          "name": "Ferret Scan Check Type",
+          "type": "ferret_scan_check_type",
+          "value": "SSN"
+        },
+        {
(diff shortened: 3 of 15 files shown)


sectoramen:
### Follow-up measurement on `identifierDigitFloor` (part 1 of #373)

Re-measured the digit floor against real documents after the fact, because 7 was taken from
the issue's wording rather than from evidence. Corpus: 150 real Office documents (842 custom
properties) plus 24 real legacy `.doc`/`.xls`/`.ppt`.

| floor | total findings on the 150 docs | CUSTOM_PROPERTY | vs `main` |
|---|---|---|---|
| `main` (no rule) | 1957 | 119 | — |
| **7 (shipped)** | 1958 | 121 | +2 |
| 9 | 1957 | 119 | **byte-identical to `main`** |

**Floor 9 would make this rule a no-op.** It admits nothing on any of the 174 real documents
available, so "no new findings" here means the change accomplishes nothing rather than that it
is free. Keeping 7.

What the two admitted values actually are, judged individually:

- an id-named custom property (`…_document_id`) holding a 7-digit value — a **true positive**,
  and precisely the case part 1 exists to recover: an identifier stored in a custom property
  that the valueless filter was discarding unscanned
- the standard SharePoint `Order` column, whose value is a list **sort position**, not an
  identifier — a **false positive**, at MEDIUM 60

So the measured cost is 1 false positive per 150 documents (0.05% of findings) to recover the
identifier class the rule targets. Both legacy corpora are unaffected (165 → 165 → 165); every
legacy `!hasLetter` custom property in that set is a single digit.

Nothing in the code changes as a result — recording the evidence so the trade is explicit
rather than implied. If the `Order`-style sort-key shape recurs on other corpora, the
discriminator would have to be the **property name** (nothing intrinsic separates a sort key
from a document ID), which is a precision-policy decision worth its own issue rather than a
quiet addition here.

rustic (approved):

PR #407, opened 2026-08-20
fix(validators): let a keyword space match zero separators so camelCase labels count as context

### LINKED ISSUE #372: A concatenated or camelCase label (memberId, driversLicense, routingNumber) never counts as context because a keyword space demands at least one separator byte
## What happens

internal/validators/kwmatch/kwmatch.go:159, inside `matchSepFlexAt`: `if n == 0 { return -1 }` — a space in a keyword must match at least one separator byte. So the keyword `member id` can never match the text `memberid` or `memberId`. Fourteen validators use kwmatch (`grep -rln kwmatch internal/validators`, excluding tests), so every multi-word context keyword fails in its concatenated form.

## Measured evidence (HEAD)

/tmp/e-batch5/memberforms.txt, `--checks MEDICAL_ID --confidence all`: `member id:` 80, `member_id:` 80, `member-id:` 80, and ZERO for `memberid:` and `memberId:`.

/tmp/e-batch5/camel.json — one file, camelCase vs snake_case pairs of the same shape:
```
driversLicense D1234562        -> 0        drivers_license B7345219      -> DRIVERS_LICENSE 75
medicalRecordNumber 4839271    -> 0        medical_record_number 5729183 -> MRN 90
routingNumber 026009593        -> no ABA   routing_number 091000019      -> ABA_ROUTING 100
```
The camelCase half of 3 of 3 validator pairs scores 0 while its snake_case twin scores 75 / 90 / 100. PASSPORT is immune and the reason matters: passport/validator.go:919-943 uses plain `strings.Contains` rather than kwmatch, so `passport` matches inside `passportnumber` — and pays for it with substring false positives.

## Why it matters (sink rule)

/tmp/e-batch5/member.json is a two-key JSON object. After `--enable-redaction`, rc=0:
```
"memberId": "XQ4839271",
"member_id": "*********"
```
One real insurance member ID cleartext, its twin redacted, same file, same run. camelCase is the default key style of JSON/REST payloads and of most ORM exports, so this is the ordinary case rather than an adversarial one.

## Scope correction

One third of the original observation is INTENTIONAL and must not be changed: `member.id` and `/member/id/` do not match because kwmatch.go:115-120 documents `isSepByte` as "Deliberately narrow: '.' and '/' produced measured false positives across sentence and URL boundaries." That is a recorded decision with a stated measurement. This issue is only about the ZERO-separator (concatenated / camelCase) form.

## Fix sketch

In `matchSepFlexAt`, let a keyword space match ZERO separator bytes (drop the `if n == 0 { return -1 }` at kwmatch.go:159) while keeping the outer word-boundary rule that anchors both ends. Do NOT touch `isSepByte`.

This change is BIDIRECTIONAL and that is the real work: kwmatch also serves negative/suppression keywords, so admitting the concatenated form can newly SUPPRESS findings — e.g. medicalid's negative keyword `ip address` would start matching `ipaddress`. It needs a full-corpus before/after diff, not just a unit test.

## Failing test

internal/validators/kwmatch: a table case asserting `ContainsLower("\"memberid\": \"xq4839271\"", "member id")` is true, plus per-validator cases in medicalid/driverslicense/bankaccount on the camelCase JSON line. All fail today. The negative cases must ship with it: `remembering` and `teammemberid` must NOT match, which the existing outer word-boundary check already gives.

Note the complementary long-form gap (`member id` not matching `member identification number`) is already handled in-tree by explicit long-form entries at internal/validators/medicalid/validator.go:744, :785, :808 — that half needs nothing.

sectoramen:
## The bug

A multi-word keyword's space demanded **at least one** separator byte, so `member id`
matched `member id`, `member_id`, `member-id` and padded alignment — but never `memberid`
or `memberId`. Text is lowercased before matching, so those last two are the same string:
the exclusion dropped the camelCase spelling that is the default key style of JSON, REST
payloads and ORM exports.

14 validator packages match multi-word keywords, so each lost its concatenated form. A
keyword can gate a **shape suppressor** rather than only a confidence boost, so the
finding does not merely score lower — it disappears. And only reported findings reach the
redactor.

On a JSON file holding camelCase/snake_case pairs of the same shape, every camelCase half
scored 0 while its twin scored 75–100:

| key | before | after |
|---|---|---|
| `routingNumber` | — | ABA_ROUTING MEDIUM 85 |
| `memberId` | — | INSURANCE_MEMBER_ID MEDIUM 80 |
| `driversLicense` | — | DRIVERS_LICENSE MEDIUM 65 |
| `routing_number`, `member_id`, `drivers_license` | already detected | unchanged |

In a two-key object that means one member ID sat in **cleartext beside its redacted twin**.
The golden snapshot shows the sink closing: `memberid: W9998887776` →
`memberid: [INSURANCE_MEMBER_ID-REDACTED]`.

## The fix

A keyword space now matches **zero or more** separator bytes.

What is deliberately *kept*:

- **`.` and `/` are still not separators.** That exclusion carries a recorded
  false-positive measurement — they cross sentence and URL boundaries, where the two words
  are unrelated (`see member.id in the docs`, `/member/id/lookup`). The zero-separator
  exclusion carried no measurement; it was asserted.
- **The outer whole-word rule**, which is what keeps zero separators honest: a left
  boundary at the anchor and a right boundary at the end. `member id` still does not match
  inside `remembering`, `teammemberid` or `memberidentification`.

## A concatenation can be a different word

Checked against the system dictionary (234,456 lowercased entries) over all **457**
multi-word lowercase string literals in non-test validator code — a superset of the
keyword lists — exactly **two** concatenate into a dictionary word:

| keyword | concatenation | role | wants the new behaviour? |
|---|---|---|---|
| `one time` | `onetime` | positive OTP keyword | yes — matching `oneTime` is the point |
| `call us` | `callus` | phone-context **suppressor** in bankaccount | no |

So the old behaviour stays available as `kwmatch.ContainsPhrase`, and `call us` uses it.
Without that, a podiatry billing line mentioning a callus would newly **silence** a real
account number sitting beside it — a suppressed finding is never redacted. Proven
load-bearing: swapping `containsPhrase` back to `containsKeyword` loses the
`US_BANK_ACCOUNT` finding on `callus debridement, bank account 4432198765`.

## Measured, bidirectionally — for findings lost as well as gained

| corpus | before | after | lost | gained |
|---|---|---|---|---|
| 150 real Office documents | 1956 | 1956 | 0 | 0 |
| 24 real legacy `.doc`/`.xls`/`.ppt` | 165 | 165 | 0 | 0 |
| 15 real FHIR / OpenAPI payloads | 45 | 45 | 0 | 0 |
| score corpus | unchanged | | | |
| golden corpus | one case moves | | | |

Set-level, not count-level: the finding sets are byte-identical on all 189 real files.

**Performance.** Dropping the fast rejection walks at most the keyword's length further on
a near-miss, so it is bounded by the keyword, not the input. 16,000 unmatched `member`
anchors: no measurable change (0.05s → 0.05s; also checked at 2k/4k/8k).

**Mutation-tested** (`-count=1`), 7 mutations, all caught — including restoring the
zero-separator rejection, inverting `requireSep` on either entry point, dropping either
word-boundary check, widening `isSepByte` to `.`/`/`, and reverting the bankaccount
suppressor to plain keyword matching.

## The golden case changed sides

`keyword_separator_forms`'s own description asserted the rule this overturns. It now
records **which clause changed sides**, that it was asserted rather than measured, and
that the `.`/`/` exclusions beside it are untouched. Line 6 becomes a finding; all four
decoy lines (`member.id`, `/member/id/`, `remember id`, `member idx`) stay silent, which
is what makes the change safe rather than merely wider.

## Interference check against the other open PRs

No textual conflict with any of the 7 other open PRs. Only `internal/goldencorpus/corpus.go`
is shared (with PR #405 and PR #393) and the union of all eight branches builds clean,
`go vet` clean, gofmt clean, **69/0**, with the golden suite passing — no merge-order hazard.

PR #396 is the one behavioural coupling candidate: it adds `kwmatch.LooksLikeFieldLabel`,
which calls `ContainsAny`, so this change widens its input. Measured as a 2x2 (main / #396 /
this / both) — effects are exactly additive, and the four binaries produce byte-identical
output on all 189 real files.

Two residual gaps that measurement turned up, both pre-existing and **not** widened here,
filed separately rather than scope-creeping this PR:

- a camelCase label alone on its own line still opens no cross-line window, because #396's
  per-word label check does not recognise a concatenated word (`member_id:` and
  `member id:` do work)
- `medicalRecordNumber` still does not report, for an unrelated reason: the concatenated
  form scores context impact 10 against the separator form's 30, below the MRN emit floor

Closes #372


sectoramen:
### Correction pushed (`c1266ce`) — the first two commits leaked, and this fixes it

I found a redaction leak in my own change after pushing, and reproduced it before fixing it. **Please review `c1266ce` as part of this PR**; the design has changed from *default* to *opt-in*.

**What was wrong.** Making a keyword space match zero separators by default widened SUPPRESSORS too. `medicalid`'s `ip address` began matching the ubiquitous key `ipAddress`, and `nonInsuranceKeywordPresent` is an unconditional veto:

```
{"member_id": "W1234567801", "ipAddress": "10.11.12.13"}

main             {"member_id": "***********", "ipAddress": "********.13"}
first 2 commits  {"member_id": "W1234567801", "ipAddress": "********.13"}   <- CLEARTEXT
```

The finding vanished entirely and the tool wrote a "redacted" file containing the member ID, reporting success. A suppressed finding is never redacted, so this is a leak, not a precision regression. `ssn` carried the same threat unfired — its suppressor list holds `part number`, `policy number`, `order number`, `employee id` and `tax id`, every one a common camelCase key.

**Why my own screening missed it.** I checked concatenations against the system dictionary, and `ipaddress` is not an English word, so it passed. The property that matters is not *is a word* but *is a token that occurs in real text*, which no word list decides. My 189-real-file corpus check also passed — those files are byte-identical under both builds, because none of them happens to put a member ID and an `ipAddress` on the same line.

**The rule now.** A positive keyword identifies the value it labels, so widening its reach can only add findings. A suppressor withholds one, so widening its reach silences real values. Therefore:

- `kwmatch.Contains` and siblings go back to requiring a separator — exactly `main`'s behaviour, and what every suppressor keeps.
- `kwmatch.ContainsLabel`/`ContainsLabelLower` is the widened form, opt-in.
- Only positive label lists opt in: medicalid's two insurance gates and its +10 boost, bankaccount's `hasBankingKeywords`, driverslicense's positive gate and keyword count. `ContainsPhrase` is deleted — with strict as the default it protected nothing, so `call us` goes back to plain `containsKeyword`.
- `medicalid.keywordsPresent` now takes its matcher from the caller, so the keywords **reported** in the context are the ones that actually **scored**. Passing it strictly left golden line 6 scoring 80 with an empty keyword list.

**Verified by the check that would have caught the original defect.** I instrumented `ContainsFunc` and recorded every call across the validator suite: **147** distinct multi-word keywords, 652 call chains, 51 non-test call sites. Each keyword, concatenated, appended to 10 positive base lines that already report → 1,450 adversarial lines, compared on finding **identity** (file, type, line) rather than score:

| build | VANISHED | gained |
|---|---|---|
| widening by default (the bug) | **1** | 50 |
| this correction | **0** | 20 |

The one vanished finding in the broken build is exactly `group number: W1234567801  ipaddress` — the sweep rediscovers the leak independently. Recall is unchanged from the broken version (`camel.json` 6→9, `memberforms.txt` 3→5) and **no golden snapshot moves**. 5 mutations, all caught, including widening `nonInsuranceKeywordPresent`'s veto.

**Union note for whoever merges second.** This correction now touches `medicalid`, so it conflicts textually with PR #396, which did not conflict before. The resolution is a one-hunk combine and the two changes turn out to be the *same principle on different axes* — #396 widens the SCOPE positives read while suppressors read only the value's own line; this widens the SPELLINGS positives match while suppressors keep requiring a separator:

```go
lineImpact := v.analyzeContext("", ctxLine)
linePositiveKeywords := v.keywordsPresent(ctxLine, v.positiveKeywords, containsLabel)
lineNegativeKeywords := v.keywordsPresent(lowerLine, v.negativeKeywords, containsKeyword)
```

Verified: union of all nine open PRs with that resolution builds, `go vet` clean, `gofmt` clean, **69/0**, and the sweep on the union still reports VANISHED=0 / gained=20.

The golden case description and the validator runbook are corrected too: #372 does not overturn "a space means at least one separator", it adds an opt-in form beside it.

rustic (approved):

Issue #429, opened 2026-08-20
SOCIAL_MEDIA whitelist_patterns does not exclude a match, it applies a 30-point penalty that the confidence cap can absorb entirely


sectoramen:
## Summary

`whitelist_patterns` in the social media config does not exclude a match. It applies a 30-point
confidence penalty and the finding is still reported, so a value the operator explicitly listed as
noise stays in the report — and stays redacted in the redacted copy.

The code says otherwise. `internal/validators/socialmedia/validator.go:1846`:

```go
// 1. Check whitelist patterns first - if match is whitelisted, exclude it
if v.isWhitelistedMatch(match) {
    return false
}
```

But that `return false` is the return value of `validateNotFalsePositive`, whose only caller treats
it as advisory:

```go
checks["not_false_positive"] = v.validateNotFalsePositive(match, platform)
if !checks["not_false_positive"] {
    confidence -= 30.0 // Strong penalty for likely false positives
}
```

Nothing drops the match. The same advisory path is shared by negative keywords, placeholder-URL
detection and the code/documentation-context check, so all four "exclude" branches are penalties.

## Reproduction

The shipped `config.yaml` whitelists `(?i)demo`, `(?i)sample` and `(?i)placeholder`.

```
Docs: instagram.com/demo and instagram.com/sample
Also pinterest.com/placeholder
Real: instagram.com/reneemueller
```

```
ferret-scan --file wl.txt --checks social_media --config config.yaml --format json --show-match
```

Observed — all three whitelisted matches reported, at MEDIUM:

| value | confidence | raw | fp_prevention |
|---|---|---|---|
| `instagram.com/demo` | 59 | 64 | −30 |
| `instagram.com/sample` | 59 | 64 | −30 |
| `pinterest.com/placeholder` | 52 | 57 | −30 |
| `instagram.com/reneemueller` | 100 | 124 | 0 |

Expected: the first three absent.

## Why the penalty cannot substitute for an exclusion

Confidence is capped at 100 and the raw total for a platform URL reaches 134, so for the
high-scoring shapes −30 lands at 104 and is capped straight back to 100 — no observable change at
all. A whitelisted match on such a shape is reported at HIGH.

## The fix is not simply "make it a veto"

The shipped patterns are unanchored substrings. Promoting them to a hard exclusion would suppress
real handles that merely contain one: `(?i)demo` would veto `instagram.com/demolition_dave`, and
[...]

PR #430, opened 2026-08-20
fix(socialmedia): refuse a platform's own system pages, so citing one is not reported or redacted

### LINKED ISSUE #349: #345 added a false-positive class: single-segment platform system URLs score HIGH 100 and are overwritten by --enable-redaction
Regression from #345, which I introduced, and my claim in that PR that it "adds no false-positive class" was **wrong**. Correcting the record with the measurement and a diagnosis.

## Measured

7 lines of ordinary documentation URLs (`instagram.com/explore`, `instagram.com/reels`, `instagram.com/tv`, `instagram.com/legal`, `pinterest.com/today`, `pinterest.com/business`, `pinterest.com/categories`):

```
pre-#345 config  ->  0 findings
merged main      -> 14 findings, ALL [HIGH 100.00%]
```

And with `--enable-redaction` every one of those URLs is overwritten:

```
Follow the brand on *********************************
Trending tags: ***************************
Reels: ************************
```

So the redactor **corrupts legitimate documentation**. That is worse than the FP itself.

## Why I missed it

I checked URLs with a **second** path segment — `instagram.com/p/ABC123`, `instagram.com/explore/tags/food`, `pinterest.com/pin/12345678` — and measured 2 findings before and 2 after, concluding no new class. Those already matched via the `<segment>/` form, so they were exactly the cases that **could not change**. The wrong control.

The class that changed is a **single** path segment with no trailing slash: `instagram.com/explore` needed a trailing `/` before, so it was skipped by accident; now it reads as a profile.

## Why instagram and pinterest specifically

They are the two platforms whose profile patterns have **no path prefix**. Compare:

```
linkedin   /in/  /company/
youtube    /user/  /c/  /channel/  /@
snapchat   /add/
tiktok     /@
instagram  <segment>          <- any single segment looks like a handle
pinterest  <segment>          <- same
```

## Diagnosis of the fix surface — the part that makes this non-trivial

I attempted a reserved-path-segment suppressor and it only partially works. What I established:

1. **`validateNotFalsePositive` is advisory, not a veto.** `validator.go:956` sets `checks["not_false_positive"]`, and `:957` applies `confidence -= 30.0`. These score 100, so a hit lands at 70 — still MEDIUM, still reported. A penalty cannot suppress them.
2. **`isOverlyBroadMatch`'s platform switch uses bare-only keys.** `case "instagram":` at ~`:2108`, where **every other switch in the file** uses `case "instagram", "instagram_patterns":`. So those cases are unreachable whenever the key is the `_patterns` form — a pre-existing bug independent of this regression.
3. **instagram and pinterest have no reserved-name list.** `github` does (`reservedGitHubNames`), but it compares the **whole match**, which only works for its bare-handle pattern, not a URL.
4. With a reserved-segment suppressor wired in and the keys aligned, **9 of 14 still scored HIGH 100** — so more than one path reaches the emit, and a demotion leaves inconsistent behaviour.

**So the fix wants a veto for reserved system paths (they are never profiles), not a confidence penalty** — and adding a veto to this scoring validator is a design decision I did not want to make unilaterally. I reverted my partial attempt rather than ship inconsistent demotion.

## Options

1. **Veto** reserved first-path-segments for prefix-less platforms (instagram, pinterest). Most correct; needs a hard-suppress path in this validator.
2. **Require a path prefix** for those two platforms, accepting that a bare `instagram.com/janedoe` profile becomes undetectable again — that regresses #345's actual fix.
3. **Revert #345 for instagram and pinterest only**, keeping tiktok and snapchat (which have `/@` and `/add/` prefixes and so are unaffected). Cheapest, restores the pre-existing state for the two risky platforms.

I would take (1), with (3) as the safe stopgap if precision matters more than the recovered true positives right now.

Whatever is chosen, it wants a test asserting BOTH directions on the same fixture — system URLs suppressed AND `instagram.com/janedoe` still detected — because the two pull against each other and a fix for either alone looks correct in isolation.

Related: #345, #343.

sectoramen:
Closes #349

## The bug

Eight of the shipped platforms have no path prefix in their profile pattern — `facebook.com/…`,
`instagram.com/…`, `pinterest.com/…`, `skype.com/…`, `t.me/…`, `twitch.tv/…`, `(twitter|x).com/…`
and `discord.gg/…` — so for those, **any** single path segment reads as a handle.
`instagram.com/explore` and `instagram.com/reneemueller` are the same shape. #345 removed the
trailing slash that had kept some of these patterns off bare paths.

Measured on a fixture of 18 platform system URLs plus 5 real profiles, with the shipped
`config.yaml`: **all 23 reported**, a 78% false-positive rate, spanning five platforms.

The severity is in the redactor rather than the report. Every one of the 18 documentation URLs was
overwritten in the redacted copy:

```
  Instagram help centre: ********************* and *******************
  Pinterest trends: *******************
  Facebook help: ***************** and ********************
  Twitter home timeline: ****************
  Telegram group invites use *************
```

A document that merely cited a platform's help page comes back corrupted.

## Why this is a veto and not a confidence penalty

The existing suppressor already ran and changed nothing. Three reasons, measured:

- `validateNotFalsePositive` is **advisory** — its failure costs 30 points at the scoring site and
  nothing drops the match.
- The raw score reaches **134** before being capped at 100, so −30 leaves 104 and the cap puts it
  straight back to **100**. On the worst offenders the penalty is not observable at all. This is
  what left the issue's earlier partial attempt with "9 of 14 still at HIGH 100".
- There is **no threshold to find**. `instagram.com/explore` scores 134 where the real profile
  `instagram.com/reneemueller` scores 124 — the false positive scores *higher*. Nothing separates
  them but vocabulary, so the discriminator has to be categorical.

`metadata.confidence_factors` shows both halves; `false_positive_prevention` was exactly `0` on all
18, so the check never fired in the first place.

## The fix

A match whose first path segment is one of the platform's own system pages is refused in
`processMatchOptimized` — the single emit path in the validator.

| fixture | main | this PR |
|---|---:|---:|
| 18 system URLs + 5 real profiles | 23 | **5** |
| near-miss handles + prefixed platforms (must all survive) | 17 | **17** |

All 5 real profiles retained at byte-identical confidence, and the documentation URLs are intact in
the redacted output.

## Built to fail in the safe direction

A wrong entry suppresses a real profile, and a suppressed finding never reaches the redactor — so it
would be a cleartext leak, not just a miss. Hence:

- **Only the first path segment is judged.** That is what makes it safe to apply the shared set to
  every platform: `linkedin.com/in/`, `reddit.com/u/`, `bsky.app/profile/` and the `/@handle` forms
  put the handle in the second segment, so `linkedin.com/in/about` is unaffected.
- **The whole segment must match**, case-folded. `instagram.com/exploration`, `twitter.com/homer`
  and `t.me/shared` are real handles and still detected.
- **The shared set holds only names reserved on every consumer platform.** Anything
  platform-specific or plausibly a handle elsewhere — `business`, `home`, `today`, `tv` — sits in
  the per-platform table.
- **Entries are grounded** in each platform's own `robots.txt` `Disallow` list where it publishes
  one, plus documented feature paths. `robots.txt` is authoritative but uneven (pinterest names 120+
  segments, twitter four), so the tables are deliberately **not exhaustive**: an uncovered system
  path stays a false positive, which is the tolerable direction of error.

One trap is pinned by a test: `facebook.com/profile.php?id=<n>` **is** a real person and the shipped
config has a pattern for it. The `.php` trim that lets `sharer.php` match `sharer` turns it into a
lookup for `profile`, so `profile` must never enter the facebook table.

## Also in this PR

`isOverlyBroadMatch`'s platform switch normalises its key. Its four cases carry the bare form only
while every other platform switch in the file accepts `"<platform>"` and `"<platform>_patterns"`.
The issue reported these cases as unreachable; that is **not** correct — no shipped config uses the
suffixed spelling (`config.yaml` and `examples/ferret.yaml` both key `platform_patterns` bare, and
`instagram_patterns` appears nowhere outside a Go display-name map), so they do fire today. The
change is defensive consistency, not the fix.

## Testing

- 8 new tests, both directions **on one fixture set** as the issue requires — the two pull against
  each other, and asserted apart, `return false` passes the profile half while `return true` passes
  the system half.
- **Mutation-tested, 7 mutants, all killed.** Two survived the first pass and both were real test
  defects, now fixed:
  - Disabling the veto's call site did not fail. Several system URLs on one line are collapsed by
    the clustering pass into a single `SOCIAL_MEDIA_CLUSTER` whose `Text` is a synthesised summary
    (`"instagram: explore | pinterest: today | twitter: settings"`) containing no slashes, so an
    assertion looking for `"/explore"` could not see them. Now matched on the bare segment name.
  - `firstPathSegment` returning the whole host on a path-less match was unobservable through
    behaviour (no shipped pattern yields a bare token equal to a reserved name), so its contract is
    now asserted directly.
- `go build ./...`, `go vet ./...`, `gofmt -l ./cmd ./internal ./pkg` clean; full suite **69
  packages ok**.
  whole_leak=0`, baseline file untouched. Worth stating plainly: the score corpus does **not** score
  SOCIAL_MEDIA (`registry.go` records it as "config-gated by design"), and the golden corpus records
  zero social-media findings, so neither gate can move on this change. Both confirm no collateral
  damage to the other validators; the evidence for the change itself is the fixtures and the real
  corpus below.
- **Real corpus: 871 files across 12 types** (txt, pdf, docx, json, md, csv, html, pptx, xlsx, htm,
  doc, rtf), 844 processed.

  Delta: **1525 → 1519 findings, 6 removed, 0 added.** Every removal is a genuine system endpoint,
  and four were at HIGH 100 — so they were being redacted out of real HTML documents:

  | removed | conf | what it is |
  |---|---:|---|
  | `facebook.com/sharer` ×2 | 100 | Facebook's share endpoint |
  | `twitter.com/intent` ×2 | 100 | Twitter's tweet-intent endpoint |
  | `x.com/help` | 50 | help page |
  | `x.com/api` | 3 | API root |

  No real profile lost: LINKEDIN 36→36, TWITCH 5→5, DISCORD 2→2, MEDIUM 3→3, YOUTUBE 1→1,
  SOCIAL_MEDIA_CLUSTER 5→5. `twitch.tv/aws`, `facebook.com/amazonwebservices` and
  `twitter.com/engineering` are all real accounts and all survive.
- **No performance cost.** On a dense 40k-line fixture the fixed build is slightly *faster* (2.35s
  vs 2.96s): a vetoed match returns before the confidence and context pipeline runs. The veto itself
  is a map lookup and a few string operations per match.
- Union-tested with the four other open PRs (#425, #426, #427, #428): pairwise `git merge-tree`
  clean, union worktree builds, vets and passes **70 packages ok**. No file overlap with any of them.

## Docs

`docs/social-media-configuration.md` gains a "Platform system pages are not profiles" section
covering the mechanism, the three safety properties and the non-exhaustiveness, a troubleshooting
entry for an uncovered system path, and a pattern guideline to prefer a platform's path prefix where
it has one.

## Found along the way, filed separately

#429 — `whitelist_patterns` does not exclude a match either. It shares the same advisory −30 path, so
`instagram.com/demo` is still reported at MEDIUM 59 despite `(?i)demo` being whitelisted in the
shipped config. Left out of this PR deliberately: the shipped patterns are unanchored substrings, so
promoting them to a veto would suppress `instagram.com/demolition_dave`, and that needs a semantics
decision rather than a one-line change.

### DIFF
--- a/docs/social-media-configuration.md
+++ b/docs/social-media-configuration.md
@@ -203,11 +203,46 @@ validators:
   `isFalsePositiveHandle`: a handle immediately followed by `.`, `-` or `@` is a
   domain or a federated (Mastodon-style) address rather than a bare handle.
 
+### Platform system pages are not profiles
+
+Eight of the shipped platforms have no path prefix in their profile pattern — `facebook.com/…`,
+`instagram.com/…`, `pinterest.com/…`, `skype.com/…`, `t.me/…`, `twitch.tv/…`,
+`(twitter|x).com/…` and `discord.gg/…` — so for those, **any** single path segment reads as a
+handle. `instagram.com/explore` and `instagram.com/reneemueller` are the same shape.
+
+The validator therefore refuses a match whose first path segment is one of the platform's own
+system pages: `about`, `help`, `legal`, `privacy`, `terms`, `settings`, `login`, `explore`,
+`search` and similar, plus per-platform additions such as Instagram's `reels` and `tv`,
+Pinterest's `today` and `categories`, Twitter's `home` and `i`, and Telegram's `joinchat`. The
+tables live in `internal/validators/socialmedia/reserved_paths.go`.
+
+Three properties are worth knowing:
+
+- **It is a refusal, not a score reduction.** Confidence is capped at 100 while the raw score for
+  these URLs reaches 134, so a penalty is absorbed by the cap and the finding still reports at
+  HIGH. It also cannot be a threshold: `instagram.com/explore` scores *higher* than a real
+  profile, because nothing distinguishes them but the vocabulary.
+- **Only the first path segment is judged.** Platforms whose pattern carries a prefix
+  (`linkedin.com/in/`, `reddit.com/u/`, `bsky.app/profile/`, and the `/@handle` forms) put the
+  handle in the second segment, so `linkedin.com/in/about` is detected normally.
+- **The whole segment must match.** `instagram.com/exploration` and `twitter.com/homer` are real
+  handles and are detected.
+
+This matters most for redaction: a reported system URL is overwritten in the redacted copy, which
+corrupts a document that was only citing the platform's own help page.
+
+The tables cover the system pages that turn up cited in documents rather than every path a
+platform serves, so an uncovered one remains a false positive. Prefer that direction: an entry
+that is wrong the other way would suppress a real profile, and a suppressed finding is never
+redacted.
+
 ### Pattern Guidelines
 - Use case-insensitive patterns: `(?i)`
 - Include both full URLs and domain-only patterns
 - Test patterns to avoid false positives
 - Consider platform-specific username rules
+- Prefer a pattern with a path prefix where the platform has one (`linkedin.com/in/…` rather than
+  `linkedin.com/…`); a prefix-less pattern makes every system page on that host a candidate handle
 
 ### Security Considerations
 - Social media detection is disabled by default
@@ -248,8 +283,15 @@ ferret-scan --file document.pdf --checks SOCIAL_MEDIA --debug
 ### False positives
 1. Add negative keywords
 2. Use whitelist patterns
-3. Refine regex patterns
+3. Refine regex patterns — give the pattern the platform's path prefix if it has one
 4. Check for email/social media conflicts
+
+### A platform's own help or policy page is reported as a profile
+
+The first path segment is not in the reserved tables for that platform. See
+[Platform system pages are not profiles](#platform-system-pages-are-not-profiles); the tables in
+`internal/validators/socialmedia/reserved_paths.go` are not exhaustive. A `whitelist_patterns`
+entry reduces the confidence of such a match but does not remove it.
 
 ### Email addresses detected as social media
 1. Remove email patterns from social media config
--- a/internal/validators/socialmedia/reserved_paths.go
+++ b/internal/validators/socialmedia/reserved_paths.go
@@ -1,0 +1,296 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package socialmedia
+
+import "strings"
+
+// Reserved first-path segments: a platform's own system pages, which are never anybody's profile.
+//
+// Most platform profile patterns carry a path prefix that disambiguates them — linkedin.com/in/,
+// reddit.com/u/, bsky.app/profile/, and the /@handle forms (clubhouse, medium, threads, tiktok,
+// mastodon). For those the first path segment is the prefix itself, so nothing here can apply to
+// them. Eight platforms have no prefix, and for them any single path segment reads as a handle:
+// facebook, instagram, pinterest, skype, telegram (t.me), twitch (twitch.tv), twitter/x, and
+// discord (discord.gg). Those are the ones these tables exist for.
+//
+// Why a veto rather than a confidence penalty. validateNotFalsePositive is advisory: a failure
+// costs 30 points at validator.go's scoring site. But the raw score for these URLs runs to 134
+// before being capped at 100, so subtracting 30 leaves 104 — still capped to 100, still HIGH,
+// still reported and still redacted. Measured on a fixture of 18 platform system URLs: every one
+// scored 54-134 raw with false_positive_prevention exactly 0, and instagram.com/explore scored
+// HIGHER (134) than the real profile instagram.com/reneemueller (124). No threshold can separate
+// them, because there is no structural difference between the two — only vocabulary. So the
+// discriminator has to be categorical.
+//
+// Why that is safe in the direction that matters. A wrong entry here suppresses a real finding,
+// and a suppressed finding is a cleartext leak (only reported findings reach the redactor), so
+// every entry must be a name the platform genuinely reserves and nobody can register. Matching is
+// against the WHOLE segment, case-folded: instagram.com/exploration is not instagram.com/explore.
+// The generic set below is limited to names reserved on every consumer platform; anything
+// platform-specific, or that could plausibly be somebody's handle elsewhere, lives in the
+// per-platform table instead of the shared one.
+//
+// Grounding. The per-platform entries are taken from each platform's own robots.txt Disallow list
+// where it publishes one — pinterest.com/robots.txt names over 120 system segments, facebook's and
+// twitter's name theirs — plus the platforms' documented feature paths. robots.txt is authoritative
+// but incomplete (twitter publishes four segments, pinterest a hundred), which is why these tables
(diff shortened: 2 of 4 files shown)


sectoramen:
## `make pr-checklist BASE=origin/main` — run after the fact, manual dimensions completed

I opened this PR without running the repo's own checklist. It reports:

```
1 gofmt / go vet / go build   RAN — clean / clean / ok      3 golden regen   RAN — 0 files changed
2 full suite (-count=1)       RAN — 68 packages ok          11 cross-platform vet  RAN — linux darwin windows
12 flake (3 repeats)          RAN — ./internal/validators/socialmedia
13 race (whole module)        RAN — 68 packages ok          -- -short / tree clean  RAN — ok / dirtied nothing
8 web / CSP                   N/A — no web/template/asset file touched
```

### 4 — redaction, EVERY strategy (the original body covered only the default)

| binary | strategy | documentation URL intact | real profile removed |
|---|---|---:|---:|
| parent | `simple` | **0** (destroyed) | yes |
| parent | `format_preserving` | **0** (destroyed) | yes |
| this PR | `simple` | **1** | yes |
| this PR | `format_preserving` | **1** | yes |

Both strategies: the parent overwrites the cited documentation URL, this PR leaves it intact and still
removes the real profile.

| binary | result |
|---|---|
| parent | `findings=0 suppressed=2` |

The difference is expected and worth stating rather than hiding: the parent generated two rules, and
this PR stops reporting one of those findings at all, so that rule has nothing left to suppress. The
finding that still exists is silenced by the parent's own hash, i.e. the change does not perturb the
suppression hash inputs. Both end at `findings=0`.

### 6 — timing / O(n²), scaling curve at N/2N/4N

Fixture: half the lines carry a reserved system segment (the veto fires), half a real handle (it must
not), so the new code path is actually exercised — a first attempt appended a random suffix to every
segment and the veto never fired, which would have measured nothing.

| lines | main | this PR | main ratio | PR ratio | findings |
|---|---|---|---|---|---|
| 5,000 | 0.45s | **0.38s** | – | – | 5,000 → 3,625 |
| 10,000 | 0.81s | **0.68s** | 1.81x | 1.79x | 10,000 → 7,285 |
| 20,000 | 1.52s | **1.31s** | 1.88x | 1.93x | 20,000 → 14,466 |

Linear on both, and this PR is consistently **faster**: a vetoed match returns before the confidence
and context pipeline runs. Non-vacuity floor satisfied (findings > 0 at every size) with distinct
values.

text 758B, json 2155B, yaml 1931B, csv 132B, junit 467B, gitlab-sast 1883B, sarif 4125B — all rc=0.

### 10 — non-vacuity

Satisfied by mutation: 7 mutants, all killed, including the two that survived a first pass (the
disabled call site, and `firstPathSegment` returning the whole host).

rustic (approved):

Issue #432, opened 2026-08-21
pkg/scan and the web UI report every coverage loss as "no document text extracted", including losses where the text WAS read


sectoramen:
## Summary

`internal/core/scanner.go` labels **every** `EmptyExtractionFiles` entry as "no document text
extracted from", regardless of why coverage was lost. The CLI was fixed for exactly this
mislabelling; the library and web surfaces were not, so they still report a true fact under a false
heading.

`internal/core/scanner.go:222`:

```go
incompleteReason = fmt.Sprintf("no document text extracted from %s: %s",
    stats.EmptyExtractionFiles[0].FilePath, stats.EmptyExtractionFiles[0].Reason)
```

That string is the only description these consumers get:

- `pkg/scan.ScanResult.IncompleteReason` (`pkg/scan/scan.go:215`, set at `:329`)
- the web UI's `incomplete_reason` field (`internal/web/server.go:98`, set at `:589`, rendered via
  `:792`)

## Why it is wrong

The `ExtractionWarning` channel carries two different statements, and only one of them is "the body
held no text":

| producer | what actually happened |
|---|---|
| office text extractor, no body part | the body really was empty — **no text** is correct |
| office metadata, embedded part over the cap | the body WAS read and scanned; one part was not |
| audio, malformed WAV chunk layout | metadata WAS recovered; the walk was realigned |
| video, box walk stopped early (#398) | some boxes WERE read; the rest were not |

For every row but the first, "no document text extracted" asserts a failure that did not happen. An
operator reading it concludes the document was unreadable, when in fact it was scanned and one part
of it was skipped — which is a different action to take.

`cmd/unscanned_report.go:270-285` already draws this distinction, bucketing on the presence of
`noBodyTextPrefix` and reporting `coverage cut short` versus `no body text`. `cmd/extraction_warning_cause_test.go`
holds a table of every real producer's wording to keep new producers in the right bucket. None of
that reaches the library path, which does its own formatting.

## Reproduction

Any file that produces a cut-short warning, scanned through `pkg/scan` or the web UI rather than the
CLI. A video whose box walk stops early is the newest example; an Office document with an
over-cap embedded part is the oldest.

## Suggested direction

The substring test is a workaround for a missing field. `EmptyExtractionFiles[]` entries would be
better carrying a **structured cause** set by the producer, which both `cmd` and `core` then render.
That removes the phrasing dependency entirely — today a producer that happens to write "no text
[...]

Issue #431, opened 2026-08-21
Apple mdta keys/ilst metadata is never reported: the keys-to-ilst index mapping is not decoded, so iPhone descriptive fields are invisible in both layouts


sectoramen:
## Summary

A real iPhone / AVFoundation video stores its descriptive metadata as a `moov > meta > keys` table
plus a parallel `ilst` list of numbered items. The extractor does not decode that mapping, so the
values are not reported — in **either** layout, so this is not #398 and #398's fix does not close it.

Measured during #398's investigation: a `com.apple.quicktime.description` value in a
`moov > meta > keys/ilst` structure produced **0 findings** with the moov at offset 20, i.e. with no
positional limit in play at all.

## Why this matters

iPhone is the population #398 exists to serve. `©`-prefixed QuickTime atoms and the iTunes `ilst`
layout are both handled, and ISO 6709 positions are found via a raw chunk scan, but the
`keys`-indexed form is the one Apple actually writes for descriptive fields.

## Mechanism

In `moov > meta`:

- `keys` holds a count then a list of key definitions, each `[size][namespace]["com.apple.quicktime.…"]`.
  The key's **index** is its 1-based position in that list.
- `ilst` holds items whose atom TYPE is the big-endian index rather than a four-character code, each
  containing a `data` atom with the value.

So a value is only attributable to a field name by pairing the two: item N carries the value for the
Nth key. Nothing decodes that pairing today.

## Explicitly not the fix

A raw text scrape over the moov payload was prototyped during #398 and measured to be a **net
negative**: on a real ffmpeg tail-moov file it changed nothing (extracted text byte-identical, 187
characters either way), and on a synthetic Apple `keys`/`ilst` moov it injected junk properties —
`Description_Apple: )data` and a `QuickTime_Metadata:` blob of mangled atom bytes — i.e. new garbage
values feeding the validators. It was left out for that reason.

## Suggested direction

Decode the `keys` table into an index → name map, then walk `ilst` items by index and read each
`data` payload, exactly as the existing `©`-atom and four-character `ilst` paths already do. The
type-code arm of the inner walkers is where this belongs, which overlaps #400's inner-walker work —
either issue could own it, but it should be one change rather than two.

## Verifying it

Build a fixture with `ffmpeg -f lavfi ... -movflags use_metadata_tags`, or take a real iPhone
recording. Assert that the value is reported in BOTH layouts (moov first and moov last), because a
fix that only works in one is the #398 defect in a new place.

[...]

PR #433, opened 2026-08-21
fix(video): find moov wherever it sits, and read a position atom by its shape

### LINKED ISSUE #398: Video metadata past the first 10 MB is never read, so a camera-default recording reports zero findings
## Summary

The video metadata extractor stops after `MaxMetadataRead` (10 MB) of atom traversal. A recording written by a camera or by ffmpeg puts `moov` **after** `mdat`, so on any video larger than ~10 MB the metadata is never reached and the file reports **zero findings**. Nothing is reported, so nothing is redacted, and the values sit in a file the run calls clean — the sink rule.

## Measured

Same content, same bytes, only the atom order differs:

| file | top-level layout | findings |
|---|---|---|
| 80 MB `.mov` as written | `ftyp, wide, mdat(80.7 MB), moov(131 KB)` | **0** |
| same after `-movflags faststart` | `ftyp, moov(131 KB), wide, mdat` | **3** — SSN HIGH 90, PERSON_NAME HIGH 100, BUSINESS LOW 48 |

The metadata is identical in both: a comment holding `449-87-4100`, an artist, an email. Reproduce:

```bash
ffmpeg -i big.mov -c copy -metadata comment="Employee SSN 449-87-4100" -metadata artist="Marcus Whitfield" out.mov
ferret-scan --file out.mov --format csv                      # 0 findings
ffmpeg -i out.mov -c copy -movflags faststart fast.mov
ferret-scan --file fast.mov --format csv                     # 3 findings
```

## Why this matters more than the number suggests

Non-faststart is the **default** for anything a device or ffmpeg produces — `moov` can only be written once the media length is known. faststart is an explicit, extra step taken for web streaming. So the case that reports nothing is the ordinary one, and the case that works is the exception.

## Root cause

`internal/preprocessors/meta-extractors/meta-extract-videolib/video-extractor.go` — `parseMP4ContainerWithContext` and its sibling accumulate `bytesRead` across the top-level walk and `break` at `MaxMetadataRead` (10 MB). Walking past `mdat` charges `mdat`'s **whole size** against that budget, so the loop stops before reaching `moov`.

## Suggested direction

The budget is meant to bound how much metadata is READ, and skipping an atom costs nothing to read: a `mdat` is traversed by seeking past it, not by reading it. So the bound belongs on bytes actually consumed from metadata atoms, not on offsets advanced. `internal/redactors/isobmff` (added in #397) already walks a whole file this way — 8 bytes per atom header, media skipped by arithmetic, with an explicit atom-count budget — and could be reused, which would also make the read and write sides agree on which bytes are metadata.

Note the asymmetry this leaves today: the video **redactor** walks the entire file, so it is ready for whatever the extractor starts reporting; it is the read side that stops early.

### LINKED ISSUE #399: Video GPS: a loci-only position is never reported, and ffmpeg's .mov ©xyz text form is reported as a garbage coordinate at HIGH 100
## Summary

Two defects in one function, found by scanning real ffmpeg output. The video extractor recognises exactly one of the three ways an ISO base media file stores a position, and mis-reads a second one.

| how the position is stored | who writes it | today |
|---|---|---|
| `loci` atom, fixed-point | **ffmpeg into `.mp4`** (its default) | **not reported at all** |
| `©xyz` atom, ISO 6709 **text** with a 2-byte length + 2-byte language prefix | **ffmpeg into `.mov`** | reported as `18.335022, 11059.211639` — **GPS HIGH 100** |
| `©xyz` atom, fixed-point | some cameras | correct |
| `meta>keys/ilst` ISO 6709 text | **a real iPhone** | correct (via a raw chunk scan) |

## Measured

```bash
ffmpeg -f lavfi -i testsrc=duration=1:size=160x120:rate=10 -c:v libx264 -pix_fmt yuv420p \
  -metadata location="+36.3506-082.6985/" out.mp4      # ffmpeg writes a loci atom
ferret-scan --file out.mp4 --show-match --format csv     # no GPS finding

# same metadata into .mov: ffmpeg writes ©xyz instead
ferret-scan --file out.mov --show-match --format csv
#   out.mov,GPS,HIGH,100.0,7,"18.335022, 11059.211639"
```

That second value is not a coordinate at all. The `.mov` `©xyz` payload is `00 12 55 c4 '+36.3506-082.6985/'`: a text length (0x12 = 18) and a language code (0x55c4), then the string. `parseGPSBox` reads the first 12 bytes as three fixed-point numbers, so `00 12 55 c4` becomes latitude `1201604/65536 = 18.335022` and `'+36.'` becomes longitude `11059.211639`.

## Consequences

- **Recall (a leak).** ffmpeg's default `.mp4` location is invisible, so it is never reported and therefore never redacted.
- **Precision.** A wrong coordinate at HIGH 100 is a false positive that also *hides* the real one: the true position `+36.3506-082.6985` is right there in the same payload and is never reported.

## Root cause

`internal/preprocessors/meta-extractors/meta-extract-videolib/video-extractor.go`:

- the udta switch has `case "©xyz": parseGPSBox(boxData, metadata)` — unconditionally binary, with no check for the text form;
- there is no `case "loci"` anywhere.

## Suggested direction

Decide by shape, not by atom name: if the payload contains an ISO 6709 string, parse it as text; otherwise read it as fixed-point. `internal/redactors/isobmff` (added in #397) has a spec-derived ISO 6709 pattern — latitude 2/4/6 integer digits, longitude 3/5/7, optional height and CRS — that could be shared, which would keep the read and write sides from disagreeing about what a position looks like.

For `loci`: version+flags(4), language(2), a null-terminated location name, a role byte, then longitude, latitude and altitude as 16.16 fixed-point, then a null-terminated astronomical body. The name is location data too.

**Note for whoever fixes this:** the redactor already scrubs all four forms including `loci` (#397), so closing this gap does not leave a new unredactable case — it will simply start reporting values that already get removed.

sectoramen:
Closes #398
Closes #399

Two commits, in this order deliberately. #399 must land first or with #398, never after — measured
below.

## #398 — a moov past 10MB was never read, silently

Video metadata lives in the `moov` box, and both standards allow it anywhere. ISO/IEC 14496-12:2015
cl. 8.2.1.1: *"Normally this box is close to the beginning or end of the file, though this is not
required."* Apple's QuickTime reference: *"QuickTime does not impose any rules about the order of
these atoms"*, with mdat-first in its own typical-file figure. ffmpeg and typical cameras write it
**last**; `-movflags faststart` is a second pass. All three local videos with a readable moov are
tail-moov, and their moov is 0.14%, 0.27% and 0.39% of the file.

The walk gave up once it passed a **10MB file offset**, and the counter was charged for bytes it
never read — `bytesRead += box.Size + BoxHeaderSize` ran for every box including the `mdat` handed
to a bare seek. A 12MB recording exhausted a "metadata" allowance while costing no memory, stopped
on a bare `break`, and returned nil.

**The miss was silent, which is the sink rule's worst case.** On a real 27MB ffmpeg encode with the
same metadata in both layouts:

| | main | this PR |
|---|---:|---:|
| moov last (ffmpeg default) | **0 findings** | SSN + PERSON_NAME + GPS, all HIGH 100 |
| moov first (`+faststart`) | 3 findings | 3 findings, unchanged |

and the two layouts now produce **byte-identical output**. On the non-faststart file main gave exit
0, no `files_not_examined` key, empty stderr, `--preprocess-only` printing `Status: Success`, and
`--fail-on-incomplete` **also exiting 0** — while `strings` found the SSN in it twice. The affected
band is ~10MB–100MB: above 100MB the router refuses the file loudly (exit 3), which is exactly why
this hid in the gap between two working gates.

The leak diff is the security statement: with `--enable-redaction`, **main writes no redacted file at
all** (nothing was reported to redact), leaving the SSN in place in a file it called clean. This PR
writes a same-length copy with the SSN, the name and the coordinate all absent.

### The new walk

Absolute offsets, no positional budget. Only box headers and metadata boxes are read; `mdat`, `free`,
`skip`, `wide`, `uuid` and anything unrecognised are stepped over by arithmetic — no read, no
allocation, not even a seek. **TIME** is O(top-level boxes), **MEMORY** is O(min(|moov|, 32MB)) and
independent of file size. No measurable cost: best-of-three on the 27MB encode is 0.21s vs main's
0.20s.

### It also closes a memory amplification in the same function

`ReadBoxData` allocated a box's **declared** size with no check against the bytes present. Measured:
sixteen files of **80 bytes each**, whose `ftyp` declares 100MB, drove **631MB RSS at exit 0 with no
disclosure**. Now 31MB, exit 3, all sixteen disclosed. That is the 4KB→8.62GB class this repo has
already been bitten by; every declared size is now bounded by the real one from `Stat`.

### Six silent stops now disclose

No moov found; a box declaring more than the file holds; a box smaller than its own header; a moov
past the 32MB parse limit; a moov that cannot be read or parsed; the box-count budget. All reach the
existing coverage-cut-short channel (verified live — a 200KB unparseable `.mov` already produces exit
3 with `files_not_examined: 1`, so the channel works and the walk simply wasn't using it).

An over-declared moov is **clamped** to the file's real end and still yields its values: measured
`findings 1` **and** `files_not_examined 1` together, so a disclosure does not throw away findings. A
healthy tail-moov file stays at exit 0 with **no** warning — a disclosure operators learn to ignore is
worse than none.

### Details that are easy to get wrong, each pinned by a test

- The EOF bound is `total > fileSize-off`, never `off+total > fileSize`. A 64-bit `largesize` near
  2^63 makes the addition overflow to a negative offset the comparison accepts.
- `size64 > math.MaxInt64` is refused rather than converted.
- `total < hdrLen` is refused, so `cur += total` cannot fail to advance or walk misaligned.
- `size == 0` ("extends to EOF") is terminal; a moov after such a box is unreachable and that is
  disclosed rather than guessed at.
- A `uuid` box's 16-byte usertype counts toward the header length. One of the three real local videos
  has a top-level `uuid` box, so this is not hypothetical.
- `MaxTopLevelBoxes = 1<<20`, matching the redactor's own atom budget. **A tighter 100k ceiling was
  tried and rejected**: it is reachable in an 800KB file and it LOST a finding main found, which is
  the wrong direction to trade even for a disclosure.

### Dead code removed

`parseMP4ContainerWithContext` and `readMP4BoxWithContext` carried an **identical** 10MB cap with
**zero callers**. The issue names that function first, so a patch applied only there would have been a
measured no-op. With the walk rewritten, `ReadBoxHeader`, `ReadBoxData`, `SkipBoxData`, `GetPosition`,
`MP4Box` and `MaxBoxSize` have no callers anywhere and go too — `MaxBoxSize` especially, since its
only remaining effect would be to invite reintroducing the refuse-instead-of-skip behaviour that makes
an oversize `mdat` unskippable.

## #399 — a position atom read the wrong way, and one never read at all

ffmpeg writes the same `-metadata location` into two different atoms depending on container, and both
were wrong. Measured on real ffmpeg 9.0.1 output:

| file | atom | main | this PR |
|---|---|---|---|
| `.mp4` | `loci` | **no finding** — never reported, never redacted | `36.350586, -82.698486` HIGH 100 |
| `.mov` | `©xyz` text | `18.335022, 11059.211639` HIGH 100 | `36.350600, -82.698500` |

The `©xyz` payload is a 2-byte text length, a 2-byte language code, then an ISO 6709 string. Read as
three 16.16 fixed-point words, the length and language bytes **become a latitude** — an impossible
longitude at HIGH confidence, while the true position sat unreported in the same bytes. A wrong
coordinate is not just noise: it buries the real one.

The atom name does not say which encoding is in use, so the payload's **shape** now decides, via
`isobmff.FindISO6709` — the same spec-derived pattern the redactor already uses. Sharing one
definition is the point: the redactor's own comments document this exact extractor bug, and a second
copy of the pattern would let the two drift apart again. (`.mp4`'s value differs in the 5th decimal
because 16.16 quantises it; that is what the file stores.)

**A third defect in the same function.** ISO 6709 Annex H has three integer widths, and `ParseFloat`
is correct only for the first: `+4012.22-07500.25/` reported latitude **4012.22** and
`+401230.5-0750015.3/` reported **401230.5**. No local producer writes the sexagesimal forms, so this
comes from the standard rather than a sample — included because fixing one garbage-coordinate path
and leaving its sibling is half a fix, and the write side already recognises all three widths. A
value that cannot be a position on Earth is now **dropped**, which removes the class rather than the
two instances.

`loci` also carries a place name (location data, now reported) and stores **longitude before
latitude** — the reverse of `©xyz`.

### Why the order matters

Measured: with #398 applied and #399 **not**, a real 19MB non-faststart `.mov` reports the wanted SSN
**plus a brand-new garbage GPS at HIGH 100**, because #398 hands #399's misparse exactly the
camera-default files it was never reaching. #399 is non-vacuous alone (it corrects a value reported
today on faststart files) and adds no new file class, so it goes first.

### Redaction stays verifiable, which the shape test could have broken

The redactor zeroes a whole `©xyz` payload precisely because `*` is a valid fixed-point 10794.66°. A
zero-filled payload holds no ISO 6709 text, so it falls to the fixed-point branch, decodes to 0/0,
and is dropped. Measured on the real `.mov`: redacted copy identical length, rescan reports **0** GPS
findings.

## Testing

- **19 new tests.** Both #399 directions asserted on one fixture set; #398's guards asserted by
  **allocation and byte counts, never wall clock** (Windows CI). Fixtures use `Truncate` holes, so a
  200MB-declared `mdat` costs no memory and no disk — affordable precisely because those bytes are
  never read.
- **Mutation-verified: 16 mutants, all killed.** Four survivors are worth recording:
  - Deleting `case "loci":` from the udta switch left every direct unit test passing — a correct
    parser nothing calls. There is now a container-level test.
  - Disabling the `©xyz` shape decision in the switch, likewise.
  - The addition-form EOF bound **and** deleting the under-length guard both still produce a warning,
    just with the wrong cause, so presence-only assertions accepted them. They now assert the cause —
    a true disclosure under a false heading is half a fix.
  - One of my own mutations was invalid (it assigned the same coordinate word twice rather than
    swapping), so the swap is now asserted with a fixture far from both the equator and the 45°
    diagonal.
- A first test run failed because the fixtures spelled an iTunes atom `atom("©cmt")` — **five** bytes
  in Go. The same trap the extractor's own comments warn about; there is now a helper that cannot be
  written wrongly.
- The video redactor and `isobmff` suites had **only** faststart fixtures (moov first). Added a
  non-faststart one, so a future write-side change cannot quietly assume the easy layout.
- `cmd/extraction_warning_cause_test.go` gains all six new wordings, so a phrasing that accidentally
  contained "no text extracted from" would fail here rather than in an operator's report.
- `go build ./...`, `go vet ./...`, `gofmt -l ./cmd ./internal ./pkg` clean; full suite **69 packages
  ok**.
  whole_leak=0`, baseline untouched.
- **Union-tested with all five other open PRs** (#425, #426, #427, #428, #430): pairwise
  `git merge-tree` clean, zero file overlap, union worktree builds and passes **70 packages ok**, and
  a behavioural matrix on real fixtures shows my video results byte-identical with and without them
  while #430's socialmedia veto still works (23 → 5).

## Docs

`docs/reference/quotas-and-limits.md` (the 10MB cap was **undocumented**; now the two work bounds are,
with the moov-anywhere guarantee), `docs/COVERAGE_DISCLOSURE.md` (had zero mentions of video; now all
six cases), `docs/user-guides/README-Enhanced-Metadata.md`, `docs/testing/TEST_PLAN.md` (the
declared-size and both-layouts lessons), and the `maxTagBytes` comment in the video redactor, which
justified itself by citing `MaxMetadataRead` as the same quantity — it never was one, and it no longer
exists.

## Filed separately, not fixed here

- **#431** — Apple `mdta keys/ilst` metadata is never reported, in **either** layout, so #398 does not
  close it. A raw moov text scrape was prototyped and measured a net negative (it injected junk
  properties), which is why this PR does not attempt it.
- **#432** — `pkg/scan` and the web UI label every coverage loss "no document text extracted", including
  losses where the text *was* read. Pre-existing, affects audio and Office too.

## Not addressed

`MaxMoovParse = 32MB` is measured headroom, not a proof: real moov runs ≤0.39% of file and the router
caps files at 100MB, so an expected worst case is ~400KB. A multi-hour 4K recording's `stbl` tables
could in principle exceed 32MB, in which case it clamps **and discloses** — correct under the sink
rule, but a recall loss. I did not have a long-form recording to measure.

### DIFF
--- a/cmd/extraction_warning_cause_test.go
+++ b/cmd/extraction_warning_cause_test.go
@@ -84,6 +84,40 @@ func TestEveryExtractionWarningProducerLandsInTheRightBucket(t *testing.T) {
 			reason:   `Text Extractor: no text extracted from .docx: no document body part was found in the archive, so document content was NOT scanned; office_metadata: embedded part "attachment.docx" was not examined: declares 52433928 bytes, over the 52428800-byte embedded cap`,
 			want:     causeCutShort,
 		},
+		// The video walk's six wordings (#398). Each is a case where SOME of the file was read and
+		// the rest was not, so all six are cut-short rather than no-text. A phrasing that happened
+		// to contain "no text extracted from" would be filed under the milder cause and reported to
+		// the operator under a heading that asserts something untrue.
+		{
+			producer: "video walk, box budget exhausted",
+			reason:   "video_metadata: video metadata may be incomplete: the box walk stopped after 1048576 top-level boxes",
+			want:     causeCutShort,
+		},
+		{
+			producer: "video walk, box declares more than the file holds",
+			reason:   `video_metadata: video metadata may be incomplete: the "mdat" box at offset 28 declares more bytes than the file holds, so it was read only to the file's real end`,
+			want:     causeCutShort,
+		},
+		{
+			producer: "video walk, structure not followable",
+			reason:   `video_metadata: video metadata may be incomplete: the box structure could not be followed past offset 20 (box "junk" declares 4 bytes, smaller than its 8-byte header)`,
+			want:     causeCutShort,
+		},
+		{
+			producer: "video walk, moov past the parse limit",
+			reason:   "video_metadata: video metadata may be incomplete: the moov box is 41943040 bytes and only the first 33554432 were parsed",
+			want:     causeCutShort,
+		},
+		{
+			producer: "video walk, moov unreadable",
+			reason:   "video_metadata: video metadata may be incomplete: the moov box at offset 27296158 could not be read in full (unexpected EOF)",
+			want:     causeCutShort,
+		},
+		{
+			producer: "video walk, no moov found",
+			reason:   "video_metadata: video metadata may be incomplete: no moov box was found in the file",
+			want:     causeCutShort,
+		},
 	}
 
 	for _, c := range cases {
--- a/docs/COVERAGE_DISCLOSURE.md
+++ b/docs/COVERAGE_DISCLOSURE.md
@@ -71,6 +71,34 @@ An embedded part of a type nothing can read stays silent on purpose. A line on s
 every decorative `.emf` in every slide deck trains operators to ignore the warnings that
 matter.
 
+### What `coverage cut short` covers in a video container
+
+Video metadata sits in the `moov` box, which the standards permit anywhere in the file and which
+ffmpeg and typical cameras write **last**. The box walk reads only headers and metadata boxes, so a
+`moov` at the end of a 99 MB recording is read like one at the start. Six cases stop it early, and
+each says so:
+
+| case | what the operator sees |
+|---|---|
+| no `moov` box in the file | `video metadata may be incomplete: no moov box was found in the file` |
+| a box declaring more bytes than the file holds | `video metadata may be incomplete: the "mdat" box at offset N declares more bytes than the file holds, so it was read only to the file's real end` |
+| a box smaller than its own header, or an unfollowable structure | `video metadata may be incomplete: the box structure could not be followed past offset N (...)` |
+| a `moov` past the 32 MB parse limit | `video metadata may be incomplete: the moov box is N bytes and only the first 33554432 were parsed` |
+| a `moov` that could not be read or parsed in full | `video metadata may be incomplete: the moov box at offset N could not be read in full (...)` |
+| more than 1,048,576 top-level boxes | `video metadata may be incomplete: the box walk stopped after 1048576 top-level boxes` |
+
+All six were **silent** before #398, and so was the much larger case they sit alongside: the walk
+used to stop once it had passed a 10 MB *file offset*, counting the media bytes it skipped without
+reading. A camera-default recording over roughly 10 MB therefore reported no metadata at all, at
+exit 0, with nothing under `files_not_examined` — and `--fail-on-incomplete` exited 0 too. The same
+file with its `moov` moved to the front reported an SSN at HIGH 100.
+
+These are `coverage cut short` rather than `no body text` for the same reason as the container cases
+above: some of the file genuinely was read. An over-declared `moov` is clamped to the file's real end
+and still yields whatever values are present, so a disclosure here does not mean nothing was found.
+
+A well-formed video never produces any of these lines.
+
 ## Per-format details
 
 ### gitlab-sast — `scan.messages[]`
--- a/docs/reference/quotas-and-limits.md
+++ b/docs/reference/quotas-and-limits.md
@@ -25,6 +25,31 @@ A file over the limit is **not** silently dropped: if it is a type the tool coul
 processed, it is reported under `files_not_examined`, listed in the `NOT FULLY EXAMINED`
 block, and `--fail-on-incomplete` exits 3. See [Coverage Disclosure](../COVERAGE_DISCLOSURE.md).
 
+### Video metadata bounds
+
+Video metadata lives in the `moov` box, which the standards allow anywhere in the file — ISO/IEC
+14496-12 cl. 8.2.1.1 says it is "normally ... close to the beginning or end of the file, though this
+is not required", and Apple's QuickTime format reference states outright that "QuickTime does not
+impose any rules about the order of these atoms". ffmpeg and typical cameras write it **last**;
+`-movflags faststart` is a second pass that moves it to the front.
+
+The extractor therefore has **no positional limit** — it walks the top-level boxes by header and
+reads only the metadata ones, so a `moov` at the end of a 99MB file is read exactly like one at the
+start, and the media payload is never read at all. Two work bounds remain:
+
+| Bound | Value | Configurable | Disclosed when it bites |
+|---|---|---|---|
+| `MaxMoovParse` | 32MB of `moov` payload parsed | No | Yes |
+| `MaxTopLevelBoxes` | 1,048,576 top-level boxes walked | No | Yes |
+
+Both are far above any real file: a real `moov` measures a fraction of a percent of its file, and a
+well-formed movie has a handful of top-level boxes. When either bound is reached the result is a
+`coverage cut short` disclosure, not a silent truncation.
+
+Before this, the extractor stopped once the walk passed a 10MB **file offset** — counting the media
+bytes it skipped without reading — so a camera-default recording over about 10MB reported no metadata
+at all, at exit 0, with nothing under `files_not_examined`.
+
 ## Processing and Performance Limits
 
 | Component | Limit | Type | Configurable | Notes |
--- a/docs/testing/TEST_PLAN.md
+++ b/docs/testing/TEST_PLAN.md
@@ -107,6 +107,22 @@ can be megabytes and a file can be tens of thousands of near-identical rows. A h
 - A measured regression is a blocker unless it is the unavoidable cost of correct
   behavior (e.g. emitting findings that were previously dropped) — and then it is
   stated in the PR with the reason.
+- **A size read out of the file is not a size.** Every allocation taken from a DECLARED
+  length needs the declared value bounded by the bytes that are really there, checked
+  against `Stat`. This repo has produced 8.62GB of RSS from a 4KB `.docx` and 631MB from
+  sixteen 80-byte `.mp4` files, both by trusting a number an attacker supplied. Assert it
+  by **allocation**, not by timing: a bomb often parses fast, so a wall-clock assertion
+  passes on the broken build and flakes on slow CI.
+  Write the bound as `declared > remaining`, never as `offset + declared > size` — a
+  64-bit size near 2^63 makes the addition overflow to a negative number that every
+  later check accepts.
+- **Where metadata sits in a file is not a bound.** A container format that permits its
+  metadata anywhere (ISO base media `moov`, which ffmpeg and cameras write LAST) must be
+  tested in both layouts, and the two must produce identical output. A fixture generator
+  that always writes the easy layout hides this completely — every MP4 fixture in this
+  repo was implicitly faststart, which is why a 10MB positional cap went unnoticed.
+  Build the large-but-cheap case with `Truncate` so the media payload is a hole: it is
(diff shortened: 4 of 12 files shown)


sectoramen:
## `make pr-checklist BASE=origin/main` — run after the fact, remaining manual dimensions completed

I opened this PR without running the repo's own checklist. Run serially, it reports:

```
1 gofmt / go vet / go build   RAN — clean / clean / ok
2 full suite (-count=1)       RAN — 68 packages ok
3 golden regen                RAN — 0 files changed
11 cross-platform vet         RAN — linux darwin windows
12 flake (3 repeats)          RAN — ./cmd ./internal/preprocessors ./internal/.../meta-extract-videolib
                                    ./internal/redactors/isobmff ./internal/redactors/video
13 race (whole module)        RAN — 68 packages ok
-- -short mode                RAN — ok
-- tree clean                 RAN — this run dirtied nothing
8 web / CSP                   N/A — no web/template/asset file touched
```

Dimensions **4 (redaction)**, **6 (timing/memory)** and **10 (non-vacuity)** are already in the PR body
— the base-vs-fix leak diff where main writes no redacted file at all, the RSS measurements
(631MB → 31MB on the amplification directory), the box-count scaling, and 16 mutants. The two that were
missing:

Run on a faststart fixture both builds detect, so the comparison is like-for-like:

| binary | result |
|---|---|

The parent's hash still silences the finding under this build, so the walk rewrite does not perturb the
suppression hash inputs.

text 759B, json 1304B, yaml 1125B, csv 127B, junit 441B, gitlab-sast 1846B, sarif 3572B — all rc=0.

---

Two operational notes from running the checklist across several branches, in case they save someone
time: it must be run **serially** (two concurrent runs gave three spurious failures in packages the
diff never touched, and the other run produced no output at all), and neither its worktree nor any
comparison binary may live under `/tmp` — the macOS reaper deleted both mid-session, which silently
turned a suppression check into "0 rules generated".

rustic (approved):

Issue #434, opened 2026-08-21
PERSON_NAME: an unpunctuated leading word is absorbed into the name span, so "Attn Marcus Whitfield" reports nothing while "Attention Marcus Whitfield" reports the routing word as part of the name


sectoramen:
## Summary

An unpunctuated word immediately before a name is absorbed into the candidate span, and depending on
the word that either **loses the name entirely** or reports a span that is not a name. A colon after
the same word makes it work.

Found while fixing #387; unrelated to that fix (line-context adjustment for these lines is measured
at exactly `0.0`, so no penalty is involved).

## Measured

`internal/validators/personname`, HEAD `b11dc5a`, via `ValidateContent`:

| line | reported |
|---|---|
| `Attn Marcus Whitfield` | **nothing** |
| `Attn: Marcus Whitfield` | `Marcus Whitfield` |
| `Attention Marcus Whitfield` | `Attention Marcus Whitfield` |
| `To Marcus Whitfield` | **nothing** |

Through the CLI with `--checks person_name --confidence all`, `Attn Marcus Whitfield Suite 400`
reports **0 findings** while `Marcus Whitfield Suite 400` reports one at 92.

`Marcus Whitfield` is a control whose surname is in the embedded list, so the separate
list-coverage issue is not in play.

## Two faces, one cause

The candidate span starts one token too early:

- **Recall.** With `Attn` or `To` the resulting three-token candidate fails validation and the name
  is never reported — so it is never redacted either, and the file exits 0 as clean. `Attn <name>`
  is an ordinary mailing/routing line, so this is not a contrived shape.
- **Span correctness.** With `Attention` the candidate DOES report, as
  `Attention Marcus Whitfield`. The reported text is not a person's name, and the redactor overwrites
  the byte span it is given — so the output differs from what a correct span would produce, and a
  reviewer reading the report sees a value that is not in the document as a name.

That the two behave differently for the same structural reason is the interesting part: whether the
absorbed word happens to keep the candidate scoring above the emit floor decides which symptom you
get.

## Not the line-context penalties

Probed directly on `Attn Marcus Whitfield Suite 400`: `negativeKeywordIndices=[]`,
`businessIndices=[]`, `productIndices=[]`, `geoIndices=[]`, total line-context adjustment `0.0`.
Note in passing that `Suite` is **not** in `geoPatternsMap` despite #387's issue text listing it as
one — worth correcting wherever that claim is repeated.

## Suggested direction

[...]

PR #435, opened 2026-08-21
fix(personname): charge a street suffix once, so a mailing-label name is reported and redacted

### LINKED ISSUE #387: A street suffix within 15 bytes of a name double-charges penalties past PERSON_NAME's 50 emit floor, so a mailing-label name is never reported and stays cleartext
## What happens

A street suffix (`Street`, `Drive`, `Avenue`, `Road`, `Suite`) appears in BOTH `geoPatternsMap` and `negativeKeywords` -- the code comment at internal/validators/personname/validator.go:1141 states that 13 of 19 geo patterns are also negative keywords. When such a token sits within 15 bytes of a name, both penalty families fire on the same word and the penalties SUM:

- -35 at validator.go:1238 (`adjustment -= 35.0 // geographic context`, gated by `nearestWithin(cache.geoIndices, nameIndex, specificPatternProximity=25)`)
- -15 at validator.go:1211 (`adjustment -= float64(closeNegativeMatches) * 15.0`, 15-byte window)
- -8 at validator.go:1235 (product context)

The -58 total takes a 100-confidence two-token list surname to 42, below the hardcoded `if confidence >= 50.0` emit floor at validator.go:198 and :274. `--confidence all` cannot lower that floor, so the finding does not exist at any verbosity.

This is NOT the old line-global geo penalty; commit b6fe830 (PR #215) made that proximity-gated. The residual is specifically the double-charge inside the narrower 15-byte negative-keyword window.

## Measured evidence

HEAD 61a85bd. Config isolated with `FERRET_CONFIG_DIR=HOME=/tmp/memverify-batch2/emptycfg` (an empty dir), cwd holding no ferret.yaml. Control name `Marcus Whitfield` (surname is in the embedded list, so the separate list-coverage issue does not confound).

```
ferret-scan --file <f> --checks person_name --confidence all --show-match --format json
```

| line | result |
|---|---|
| `Marcus Whitfield` | 1 finding, PERSON_NAME 100 HIGH |
| `The consultant Marcus Whitfield signed off.` | 1 finding, 100 HIGH |
| `1247 Oakmont Foobar, Marcus Whitfield` (non-geo control) | 1 finding, 92 HIGH |
| `1247 Oakmont Street, Marcus Whitfield` | **0 findings** |
| `1247 Oakmont Drive, Marcus Whitfield` | **0 findings** |
| `1247 Oakmont Avenue, Marcus Whitfield` | **0 findings** |
| `88 Ridgeway Road, Marcus Whitfield` | **0 findings** |
| `Attn Marcus Whitfield Suite 400` | **0 findings** |

5 of 11 realistic mailing-label / attn forms report nothing.

Distance sweep (`1247 Oakmont Street` + N dots + ` Marcus Whitfield`; geo word at offset 13, name at 20+N, so distance = 7+N):

| N | distance | result |
|---|---|---|
| 0 | 7 | 0 findings |
| 5 | 12 | 0 findings |
| 10 | 17 | 1 finding, 57 MEDIUM |
| 15 | 22 | 1 finding, 57 MEDIUM |
| 20 | 27 | 1 finding, 92 HIGH |
| 60 | 67 | 1 finding, 92 HIGH |

The cliff is at distance < 15 (the negative-keyword window), not < 25 (the geo window). Score decomposition matches the arithmetic exactly: 100-8 = 92 (geo out of range), 100-8-35 = 57 (geo only), 100-8-35-15 = 42 (both) -> dropped.

## Sink impact -- YES

`--enable-redaction --redaction-output-dir /tmp/memverify-batch2/redout3` (dir emptied first) on `1247 Oakmont Street, Marcus Whitfield`: stats `total_findings 0 / high 0 / medium 0 / low 0`, exit code 0, and `find redout3 -type f` returns NOTHING. No redacted copy is written at all, so the name remains cleartext in the source while a clean exit certifies the file.

Proof the score is the only variable: the SAME line with a `Contact: ` prefix (which adds a keyword boost) scores 54 and IS redacted to `Contact: 1247 Oakmont Street, ****************`.

## Fix sketch

In `analyzeLineContextForMatch` (validator.go:1190-1240), stop charging one word under two penalty families: when a nearby token is present in both `negativeKeywords` and `geoPatternsMap`/business/product, charge `max(penalty)` rather than the sum. Alternatively clamp the total line-context adjustment (e.g. floor it at -45) so line context alone cannot push a two-token list-surname name from 100 below the 50 floor. Either way keep the 25-byte geo window from PR #215. Re-run the PR #215 sweep corpus (10 real names with addresses, 24 place/org decoys) and show the decoys stay at 5/24.

## Failing test

internal/validators/personname/geo_proximity_test.go: add `TestPersonName_AddressSuffixAdjacentToNameStillEmits` asserting `ValidateContent("1247 Oakmont Street, Marcus Whitfield")` returns exactly one PERSON_NAME match. Today it returns zero. Non-vacuity is established by the non-geo control `"1247 Oakmont Foobar, Marcus Whitfield"`, which returns one match at 92.

## Dedup

Not the surname-list issue (Whitfield IS a list surname; the plain-line control scores 100). Not in the 15 open issues, not in #368-#385, and the help-audit class-B / PR-plan files contain no personname-geo entry.

sectoramen:
Closes #387

## The bug

13 of 19 geographic patterns and 17 business patterns are **also** `negativeKeywords`, so one word
beside a name was charged under two penalty families at once. `Street` next to a name paid −35 as
geography **and** −15 as a negative keyword, plus −8 for product context: 100 − 58 = **42**, under the
hardcoded 50 emit floor at two call sites.

`--confidence all` cannot lower that floor, so the finding did not exist at any verbosity — and
`--enable-redaction` wrote **no output file at all**. The name stayed cleartext in a file the run
exited 0 on.

## Measured

The failing shape is the address BEFORE the name — a shipping row, an attn line, a mailing label:

| corpus | before | after |
|---|---:|---:|
| mailing labels (address then name) | **1/8** | **7/8** |
| #215 must-find (name then address) | 10/10 | 10/10 |
| #215 must-stay-suppressed | 0/16 | 0/16 |

The pre-existing corpus is entirely *name-then-address* — even
`"Deborah Callahan, 1200 State Street, Suite 300"` passes today — which is why this went unnoticed.
That ordering keeps the street suffix outside the 15-byte negative-keyword window; the label ordering
puts it inside.

## Why this is allowed to change

The stacking was deliberate. The comment above `specificPatternProximity` read: *"That is intentional:
the two penalties stack, so the keyword path still contributes at its own tighter distance."*

But that was an **asserted** design choice sitting next to a **measured** one — #215's proximity gate,
which replaced a line-global penalty precisely because it was deleting real names. Only the assertion
is overturned. The 25-byte window is untouched and all of #215's guard tests still pass, including
`TestSpecificPatternPenaltyIsProximityGated` and `TestDistantPatternCannotMaskAnAdjacentOne`.

## The span test is the fix, not a detail

A word **outside** the name is charged once, at the higher of the two rates. A word **inside** the
candidate span still stacks — because there the geography and the "name" are the same evidence.

Measured during development: de-duplicating unconditionally returned
`"Jordan Lake State Recreation Area"` as a PERSON_NAME at LOW 57. A false positive on a lake. The span
test is what distinguishes `Jordan Lake` (geo word *is* the second token) from
`1247 Oakmont Street, Marcus Whitfield` (geo word is 8 bytes before a real name).

## Real-corpus effect — 268 documents, 8 types

docx, pdf, txt, pptx, md, csv, html, xlsx:

- **4475 → 4476 findings.** +1, none removed.
- **74 findings promoted MEDIUM → HIGH**, 76 → 91, and every one of them is the **same value** — one
  real person's name, un-penalised by exactly the 15 points that were double-charged. That is the
  intended effect, at the band that drives the pre-commit exit code.

The single new finding is `Silver Spring, MD` at **LOW 52**, and it is **pre-existing rather than
new**: main reports it at MEDIUM 65 when the line stands alone, with no negative keyword and no
geographic pattern involved at all. The double charge was incidentally masking one instance of it.

That is the class #215's own comment already excluded from its corpus — *"those place words are
themselves entries in the first-name database, so they score on their own merits and the geographic
penalty only masked them incidentally — at the cost of deleting real names"* — and its stated fix is
curating the name database, not widening a window. Trading one masked LOW instance of a known FP class
for seven real names recovered from total silence is the direction the sink rule asks for.

## Two defensive details, honestly labelled

The per-offset record keeps the **maximum** penalty, and the skip applies only when that maximum is at
least the negative-keyword penalty. Measured, **neither is reachable through any real line today**:

| intersection | count |
|---|---:|
| `product ∩ negativeKeywords` | 0 of 15 |
| `product ∩ geo` / `product ∩ business` / `business ∩ geo` | 0 each |
| `geo ∩ negativeKeywords` | 12 of 13 |
| `business ∩ negativeKeywords` | 17 of 49 |

Only geo (−35) and business (−20) ever twin with a negative keyword, both already above the −15 they
displace. So both details exist so that **adding** a word to a list cannot silently turn the
de-duplication into a penalty *reduction*. Both are asserted directly against a hand-built cache, and
the list invariant that keeps them dormant is asserted too — if someone adds a product word to
`negativeKeywords`, that test fires and points here.

## Testing

- New corpus for the untested ordering, plus the both-directions pair (labels reported **and** place
  names still suppressed) on one fixture set, since a fix for either half alone looks correct.
- **Mutation-verified: 7 mutants, 6 killed behaviourally.** The seventh — dropping the
  at-least-as-large condition — is an **equivalent mutant** for exactly the reason in the table above,
  and is reported as such rather than papered over with a contorted fixture. Two others (min-instead-
  of-max, first-family-wins) survived a first pass and are now killed by the synthetic-cache test.
- `go build ./...`, `go vet ./...`, `gofmt -l ./cmd ./internal ./pkg` clean; full suite green.
  whole_leak=0`, PERSON_NAME still 17/17 at precision 1.0000, baseline untouched.
- Union-tested with the other open PRs: no file overlap, union builds and passes.

## Docs

No doc change: PERSON_NAME's penalty values and windows are code-only. `docs/validators-new.md`'s
negative-keyword table describes DRIVERS_LICENSE, and `docs/development/creating_validators.md`'s −15
is an authoring example — neither describes this validator. The stale claim was the code comment, and
it is corrected in place.

## Filed separately

**#434** — an unpunctuated leading word is absorbed into the name span: `Attn Marcus Whitfield`
reports nothing, while `Attention Marcus Whitfield` reports the routing word *as part of the name*.
That is the eighth mailing-label line, and its line-context adjustment is measured at exactly `0.0`,
so it is a different defect. Worth noting for anyone reading #387: its text lists `Suite` as a
geographic pattern, and it is not one.

### DIFF
--- a/internal/validators/personname/double_charge_test.go
+++ b/internal/validators/personname/double_charge_test.go
@@ -1,0 +1,239 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package personname
+
+import (
+	"strings"
+	"testing"
+)
+
+// mailingLabelLines put the address BEFORE the name, which is what a mailing label, an "attn" line
+// and a shipping row all look like.
+//
+// The pre-existing corpus in geo_proximity_test.go is entirely name-then-address — even
+// "Deborah Callahan, 1200 State Street, Suite 300" passes today — so this ordering was untested,
+// and it is the ordering where the street suffix lands within the 15-byte negative-keyword window
+// of the name.
+var mailingLabelLines = []struct {
+	line string
+	want string
+}{
+	{"1247 Oakmont Street, Marcus Whitfield", "Marcus Whitfield"},
+	{"1247 Oakmont Drive, Marcus Whitfield", "Marcus Whitfield"},
+	{"1247 Oakmont Avenue, Marcus Whitfield", "Marcus Whitfield"},
+	{"88 Ridgeway Road, Marcus Whitfield", "Marcus Whitfield"},
+	{"Ship to 500 Industrial Boulevard, Renee Mueller", "Renee Mueller"},
+	{"Attn: 42 Lakeview Drive, Carlos Ramirez", "Carlos Ramirez"},
+	{"Deliver to 9 Summit Road, Priya Anand", "Priya Anand"},
+}
+
+// A street suffix must be charged once, not once per penalty family.
+//
+// 13 of 19 geographic patterns and 17 business patterns are also negativeKeywords, so a single
+// "Street" within 15 bytes of a name paid -35 as geography AND -15 as a negative keyword, plus -8
+// for product context. Measured before this: 100 - 58 = 42, under the hardcoded 50 emit floor at
+// two call sites — so the finding did not exist at ANY `--confidence` setting, and
+// `--enable-redaction` wrote no output file at all. The name stayed cleartext in a file the run
+// certified clean (#387).
+//
+// The stacking was deliberate — the comment above specificPatternProximity said so — but it was an
+// asserted design choice sitting next to a measured one (#215's proximity gate). Only the assertion
+// is overturned here; the 25-byte window stays exactly as #215 left it.
+func TestStreetSuffixIsChargedOncePerWord(t *testing.T) {
+	for _, tc := range mailingLabelLines {
+		t.Run(tc.line, func(t *testing.T) {
+			v := NewValidator()
+			matches, err := v.ValidateContent(tc.line, "labels.csv")
+			if err != nil {
+				t.Fatalf("ValidateContent: %v", err)
+			}
+			var got []string
+			for _, m := range matches {
+				got = append(got, m.Text)
+				if strings.Contains(m.Text, tc.want) {
+					return
+				}
+			}
+			t.Errorf("%q was not reported (findings: %v). One word is being charged under two "+
+				"penalty families, which pushes a list-surname name below the 50 emit floor — and "+
+				"an unreported name is never redacted, so it stays cleartext.\n  line: %q",
+				tc.want, got, tc.line)
+		})
+	}
+}
+
+// The other direction, and the reason the fix needs a span test rather than a simple de-duplication.
+//
+// In "Jordan Lake State Recreation Area" the geographic word IS the second token of the candidate:
+// the "name" is the place. Both charges are then evidence about the same thing and the stack is
+// correct. Measured during development: de-duplicating unconditionally returned this line as a
+// PERSON_NAME at LOW 57 — a false positive on a lake.
+//
+// So the double charge is only dropped for a word OUTSIDE the matched span.
+func TestGeographicWordInsideTheNameStillStacks(t *testing.T) {
+	for _, line := range []string{
+		"Jordan Lake State Recreation Area",
+		"Madison Avenue advertising firm",
+		"Elm Street Elementary School District",
+		"Cedar Creek Road crosses Pine Valley Avenue",
+		"Golden Gate Boulevard runs through the city",
+	} {
+		t.Run(line, func(t *testing.T) {
+			v := NewValidator()
+			matches, err := v.ValidateContent(line, "doc.txt")
+			if err != nil {
+				t.Fatalf("ValidateContent: %v", err)
+			}
+			if len(matches) > 0 {
+				var got []string
+				for _, m := range matches {
+					got = append(got, m.Text)
+				}
+				t.Errorf("reported %v as person name(s) on a line that names no person. Dropping "+
+					"the second charge for a word INSIDE the candidate span turns place names into "+
+					"people.\n  line: %q", got, line)
+			}
+		})
+	}
+}
+
+// specificPatternPenalties must report offsets that are directly comparable with
+// negativeKeywordIndices, because that comparison is the whole mechanism. Both are
+// first-occurrence byte offsets into the same lowered line.
+func TestSpecificPatternPenaltiesReportsComparableOffsets(t *testing.T) {
+	v := NewValidator()
+	v.ensureNamesLoaded()
+
+	const line = "1247 oakmont street, marcus whitfield"
+	cache := v.newLineContextCache(line)
+	nameIndex := strings.Index(line, "marcus")
+	if nameIndex < 0 {
+		t.Fatal("premise broken: name not found")
+	}
+
+	penalty, charged := v.specificPatternPenalties(cache, nameIndex)
+	if penalty == 0 {
+		t.Fatal("no specific-pattern penalty on a line with a street suffix, so this test is vacuous")
+	}
+	if len(charged) == 0 {
+		t.Fatal("no charged offsets returned, so the caller cannot tell which words were paid for")
+	}
+
+	// The street suffix must appear at the SAME offset in both families, or the de-duplication can
+	// never fire and the fix is inert.
+	streetOffset := strings.Index(line, "street")
+	if _, ok := charged[streetOffset]; !ok {
+		t.Errorf("charged offsets %v do not include the street suffix at %d; specificPatternIndices "+
+			"and negativeKeywordIndices must agree on offsets or one word cannot be recognised as "+
+			"one word", charged, streetOffset)
+	}
+	var inNegative bool
+	for _, idx := range cache.negativeKeywordIndices {
+		if idx == streetOffset {
+			inNegative = true
+		}
+	}
+	if !inNegative {
+		t.Errorf("negativeKeywordIndices %v does not include %d: the premise of this fix is that "+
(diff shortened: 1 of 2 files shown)


sectoramen:
## Complexity and timing (TEST_PLAN §4) — omitted from the original body, added here

`internal/goldencorpus` `TestValidatorComplexityIsSubQuadratic/personname` passes (it runs as part of
`go test ./...`, so it was green on this branch already — I simply did not report it).

Scaling curve, dense input (6 `<number> Oakmont <suffix>, <First> <Last>` pairs per line), best of
three per point:

| lines | main | this PR | main ratio | PR ratio |
|---|---|---|---|---|
| 2000 | 0.57s | 0.66s | – | – |
| 4000 | 1.06s | 1.21s | 1.85x | 1.84x |
| 8000 | 2.07s | 2.35s | 1.95x | 1.94x |

**Linear, and the slope is identical to main** (a quadratic would roughly quadruple per doubling).

The ~14% constant factor is the cost of the findings the fix exists to emit, not a regression:

| lines | main findings | this PR |
|---|---:|---:|
| 2000 | 5,887 | **11,060** |
| 8000 | 23,722 | **44,490** |

So ~1.9x the findings for ~1.14x the time — the per-finding cost falls. TEST_PLAN §4 names this case
explicitly: a measured regression is acceptable when it is "the unavoidable cost of correct behavior
(e.g. emitting findings that were previously dropped)".

**Miss path — the cost on an ordinary document, where neither build finds anything:**

| input | main | this PR | delta |
|---|---|---|---|
| 15,000 prose lines | 0.72s | 0.72s | −0% |
| 60,000 prose lines | 2.62s | 2.64s | +1% |

No measurable cost when the changed loop does not fire, which is the common case.

sectoramen:
## `make pr-checklist BASE=origin/main` — run, and the manual dimensions completed

I had not run the repo's own checklist before opening this PR. It reports:

```
1 gofmt / go vet / go build     RAN — clean / clean / ok
2 full suite (-count=1)         RAN — 68 packages ok
3 golden regen                  RAN — 0 files changed
11 cross-platform vet           RAN — linux darwin windows
12 flake (3 repeats)            RAN — ./internal/validators/personname
13 race (whole module)          RAN — 68 packages ok
-- -short mode                  RAN — ok
-- tree clean                   RAN — this run dirtied nothing
8 web / CSP                     N/A — no web/template/asset file touched
```

The five manual dimensions, now done:

### 4 — redaction, every strategy. This is the base-vs-fix leak diff, and it is the point of the PR

| binary | strategy | result |
|---|---|---|
| parent | `simple` | **NO FILE WRITTEN** — nothing reported, so nothing redacted |
| parent | `format_preserving` | **NO FILE WRITTEN** |
| this PR | `simple` | `1247 Oakmont Street, [PERSON-NAME-REDACTED]` — 0 of 2 names survive |
| this PR | `format_preserving` | `1247 Oakmont Street, ****************` — 0 of 2 names survive |

The parent leaves both names in cleartext in a file it exits 0 on. That is the security result stated
as TEST_PLAN §5 asks.

Rule generated with the parent binary, `enabled: true` set, applied with both:

| binary | result |
|---|---|

The parent's hash still silences the finding under this build, so the change does not perturb the
suppression hash inputs — which matters here because the fix changes a **confidence value**, and
confidence is one of those inputs.

### 6 — timing / O(n²)

In the previous comment. Non-vacuity and distinct values both satisfied: 5,887–44,490 findings across
the sizes, values randomised rather than repeated.

| format | rc | bytes | PERSON_NAME present |
|---|---:|---:|---:|
| text | 0 | 846 | 2 |
| json | 0 | 3108 | 4 |
| yaml | 0 | 2744 | 4 |
| csv | 0 | 209 | 2 |
| junit | 0 | 578 | 1 |
| gitlab-sast | 0 | 2983 | 12 |
| sarif | 0 | 6073 | 10 |

### 10 — non-vacuity

Satisfied by mutation rather than by reverting: 7 mutants, 6 killed behaviourally. The 7th is an
equivalent mutant, documented in the PR body with the measured reason (`product ∩ negativeKeywords` is
empty) rather than papered over.

rustic (approved):

Issue #436, opened 2026-08-21
MEDICAL_ID: the MRN scan has no column-header arm, so a table whose header names a medical record number reports nothing — while the same layout with an insurance header works


sectoramen:
## Summary

`medicalid` has a column-header arm for the INSURANCE scan and none for the MRN scan, both in the same
file and both added by the same change (#396). So a CSV or table export whose column header names a
medical record number reports nothing, while the identical layout naming an insurance member ID
reports normally.

## Measured

HEAD `b11dc5a`, `--checks MEDICAL_ID --confidence all`, isolated config:

| fixture | findings |
|---|---:|
| `medical record number,order number` / `5729183,1234567` | **0** |
| `mrn,order number` / `5729183,1234567` | **0** |
| `medicalRecordNumber,orderNumber` / `5729183,1234567` | **0** |
| `medical_record_number: 5729183` (same line) | 1, MRN HIGH 90 |

The spelling of the header is irrelevant — a bare `mrn` header fails too — so this is not the
label-form gap fixed in #408. It is that the value's own line carries no context and nothing looks up.

## Mechanism

Two adjacent gates in `internal/validators/medicalid/validator.go`:

```go
// line ~312 — MRN: same-line context only
if v.hasMedicalContext(lowerLine) {
    if !scanMatches(reMRN, v.evaluateMRN) { return matches }
}

// line ~325 — insurance: same-line OR the column header
if v.hasInsuranceContext(ctxLine) || v.headerRowHasInsuranceContext(lc) {
```

`headerRowHasInsuranceContext` (~:351) and `columnHeaderAt` (~:337) already exist and are already
wired into the line context. The MRN gate simply does not consult them.

## Why it matters

A tabular export is the normal shape for this data — a datamart extract, a registry dump, a
spreadsheet handed to an analyst. The value is a bare 6-10 digit number, so with no context the MRN
regex never runs and nothing is reported. An unreported identifier is never redacted either, so it
stays cleartext in a file the scan exits 0 on.

Measured for contrast, the same table with an insurance header IS reported, which is what makes this
an inconsistency rather than a design choice.

## Suggested direction

Give the MRN gate the arm the insurance gate has:

```go
if v.hasMedicalContext(lowerLine) || v.headerRowHasMedicalContext(lc) {
```

with `headerRowHasMedicalContext` mirroring `headerRowHasInsuranceContext` but consulting the medical
[...]

Issue #409, opened 2026-08-20
A camelCase label alone on its own line opens no cross-line window: LooksLikeFieldLabel's per-word gate does not know concatenated labels


sectoramen:
## Summary

PR #396 opens a cross-line window when the previous line "looks like a field label"
(`kwmatch.LooksLikeFieldLabel`). PR #407 makes a concatenated/camelCase label count as
context everywhere else. The two do **not** compose: a camelCase label alone on its own
line still opens no window, because `LooksLikeFieldLabel` has a second, independent gate —
a per-word loop requiring every word on the line to be a keyword word or a generic label
word — and `memberid` is not a word in any keyword.

The result is an inconsistency in the direction of a miss, and a missed finding is never
redacted.

## Reproduction

Measured with four binaries (main / #396 / #407 / both), findings per fixture:

| fixture (label on line 1, value on line 2) | main | #396 | #407 | both |
|---|---|---|---|---|
| `member id:` / `W9998887776` | 0 | **1** | 0 | 1 |
| `member_id:` / `W9998887776` | 0 | **1** | 0 | 1 |
| `memberId:` / `W9998887776` | 0 | 0 | 0 | **0** |
| `MemberId:` / `W9998887776` | 0 | 0 | 0 | **0** |
| `memberID:` / `W9998887776` | 0 | 0 | 0 | **0** |
| `memberId: W9998887776` *(same line)* | 0 | 0 | **1** | 1 |

So the spaced and snake_case spellings get a cross-line window; every concatenated spelling
gets one only when the value shares its line.

## Mechanism

`LooksLikeFieldLabel` tokenises the lowercased line and requires each word to pass:

```go
for _, w := range strings.FieldsFunc(lower, ...) {
    if hasDigit(w) { ... }
    if genericLabelWords[w] { continue }
    if keywordWord(w, keywords) { continue }
    return false
}
```

`member_id:` splits on `_` into `member` + `id`, both keyword words, so it passes.
`memberid:` is a single token that is not a word in any keyword, so `keywordWord` returns
false and the label is rejected — regardless of the matcher change, since this loop does not
consult it.

## Why it is filed separately

- Both PRs are correct in isolation and their effects are otherwise exactly additive; the
  four binaries produce byte-identical output on 189 real files.
- Neither PR introduces the gap: on `main` no cross-line window exists at all.
- The fix belongs to `LooksLikeFieldLabel`, i.e. #396's code, and only becomes meaningful
[...]

PR #437, opened 2026-08-21
fix(medicalid): match a label's concatenated spelling, so a camelCase MRN field is scanned

### LINKED ISSUE #408: A concatenated three-word label matches no two-word keyword, so medicalRecordNumber carries zero medical context and the MRN scan never runs
## Summary

A keyword list holding only the **sub-phrases** of a label cannot match that label's
concatenated spelling, because the whole-word rule correctly refuses a keyword that is a
substring of a longer word. `medicalRecordNumber` therefore carries **no medical context at
all**, and the MRN scan is gated on medical context, so the regex never runs.

This is the mirror image of a gap the codebase already documents. `hasMRNKeyword` says:

```go
// Long form of "patient id" — not reachable from the abbreviation,
// since keywords match whole tokens.
"patient identification",
```

The same reasoning applies in the other direction: the long form is not reachable from the
*sub-phrases* either.

## Reproduction (at HEAD, and on PR #407)

```
$ printf 'medicalRecordNumber: 4839271\n'   | ferret-scan --checks MEDICAL_ID --confidence all
(no findings)
$ printf 'medical_record_number: 5729183\n' | ferret-scan --checks MEDICAL_ID --confidence all
MRN  HIGH  90.0
```

Not a regression, and **not** widened by PR #407 — the camelCase form never worked. #407 is
what makes the fix below possible.

## Mechanism, verified in-package

For the line `medicalrecordnumber: 4839271`:

| keyword in the MRN list | matches? | why |
|---|---|---|
| `medical record` | no | right boundary: `number` continues the word |
| `record number` | no | left boundary: `medical` precedes it |
| `mrn` | no | not present |
| `medical record number` *(not in the list)* | **yes** | the full concatenation is one whole token |

So:

- `hasMedicalContext(line) == false`
- `validator.go:231` — `if v.hasMedicalContext(lowerLine) { scanMatches(reMRN, ...) }` — the
  MRN regex `\b\d{6,10}\b` is never run
- `evaluateMRN` called directly on the same input returns `ok=true, confidence=15`, so the
  evaluator would accept it; it is simply never reached

## Suggested fix

Add the full-length label to the MRN and medical-context keyword lists, e.g.
`medical record number`, alongside the existing `medical record` / `record number`. With
PR #407's matcher that one entry covers the spaced, snake_case, kebab-case **and**
concatenated/camelCase spellings at once, taking confidence from an unreachable 15 to
70 + context.

Verified today: `kwmatch.ContainsLower("medicalrecordnumber: 4839271", "medical record number")`
is `true` on #407's branch.

## Wider audit this implies

Any validator whose multi-word keyword list holds only sub-phrases of the label real payloads
use has the same blind spot for the concatenated spelling. Worth one sweep across the 14
packages that match multi-word keywords: for each list, is the **full** label present, not
only its halves? Measure any additions bidirectionally (lost findings as well as gained) —
adding a long-form keyword is not automatically free.

sectoramen:
Closes #408

## The bug

A keyword list holding only the **sub-phrases** of a label cannot match that label's concatenated
spelling. The whole-word rule correctly refuses `medical record` inside `medicalrecordnumber` —
`number` continues the word — and equally refuses `record number`, because `medical` precedes it.

So `medicalRecordNumber: 4839271` carried **no medical context at all**, `hasMedicalContext` returned
false, and the MRN regex never ran. Measured at HEAD:

| line | result |
|---|---|
| `medicalRecordNumber: 4839271` | **no findings** |
| `medical_record_number: 5729183` | MRN HIGH 90 |

`evaluateMRN` accepts the value all along — it was simply never reached. A gate bug, not a scoring one,
and an unreported identifier is never redacted either.

## Two changes, neither of which works alone

The full-length spellings have to be **listed explicitly**, because three words joined into one token
are reachable only from a three-word keyword. And the match has to use **`containsLabel`**, whose
spaces match zero separators, because the strict matcher cannot see a keyword's spaces in a spelling
that has none. The package already recorded half of this asymmetry, for `patient identification` in
`hasMRNKeyword`.

Verified with the matcher directly, including the boundary cases that stop this becoming a substring
search: `nonmedicalrecordnumbers` and `medicalrecordnumbering` are **refused**, and the two-word
`medical record` still correctly declines the three-word concatenation.

## Why widening these two gates is safe

The restriction that makes it safe is already documented in this file: `containsLabel` is confined to
**positive** gates because when the widening was the default, the suppressor `ip address` matched
`ipAddress` and vetoed a real member ID. Both gates here only ever admit findings —
`hasMedicalContext` enables the scan, and the MRN keyword flag boosts confidence and *cancels* the
soft non-medical suppressor.

## Measured

None of these reported anything before:

| label | after |
|---|---|
| `medicalRecordNumber` / `MedicalRecordNumber` / `medicalrecordnumber` | MRN 70 |
| `patientIdentificationNumber` | MRN 70 |
| `healthRecordNumber` / `hospitalRecordNumber` | MRN 45 |
| `patientId` | MRN 80 |
| `patientNumber` / `chartNumber` / `admissionNumber` | MRN 70 |

On a realistic JSON export (the shape an ORM or datamart extract writes): **0 → 9 findings**.

## The precision boundary

A bare `record number` is deliberately **not** admitted to `hasMedicalContext`. It names an MRN only in
a medical setting, and that list is what decides whether the line *is* a medical setting — admitting it
would make every generic record identifier medical. It stays in the MRN keyword list, which only
applies once the line is already medical.

Verified unchanged at none: `recordNumber`, `orderNumber`, `invoiceNumber`, `trackingNumber`,
`serialNumber`, `accountNumber`, `referenceNumber`, `confirmationNumber`, `ticketNumber`.

## Real corpus — and an honest reading of its zero

239 documents across 8 types: **92 MEDICAL_ID findings before and after, zero added, zero removed.**

That zero is offered as a **precision** result only. It is *vacuous for recall*, and measured to be so
rather than assumed: the corpus contains **zero** camelCase medical labels — grepped, 0 files for each
of nine spellings (`medicalRecordNumber`, `patientId`, `chartNumber`, …) against 9 files containing
`mrn` and 2 containing `patient id`. The change cannot reach it, so it cannot demonstrate the gain. The
recall evidence is the label fixtures and the JSON export above.

## Testing

- 5 new tests: the concatenated spellings reaching the scan, generic identifier labels staying out,
  word boundaries surviving the widening, and both gates asserted directly so a failure says which
  layer broke.
- **Mutation-verified: 5 mutants, all killed** — reverting either gate to the strict matcher, dropping
  either group of added keywords, and the one that matters for precision: admitting a bare
  `record number` as medical context.
- `go build ./...`, `go vet ./...`, `gofmt -l ./cmd ./internal ./pkg` clean; full suite **69 packages
  ok**.
  whole_leak=0`, MEDICAL_ID still 1/1, baseline untouched.
- Union-tested with all seven other open PRs: pairwise `git merge-tree` clean, no file overlap, union
  builds and passes.

## Docs

No doc change. The keyword lists are code-only, and no document states which spellings are recognised.

## Filed separately

**#436** — the MRN scan has **no column-header arm**, while the insurance scan in the same file does
(both added by #396). A table whose header names a medical record number reports nothing:
`medical record number,order number` / `5729183,1234567` → 0 findings. Independent of this fix — a
*spaced* header fails on both builds, and even a bare `mrn` header does — so it is the "context is not
on the value's line" family, related to #409.

### DIFF
--- a/internal/validators/medicalid/longform_label_test.go
+++ b/internal/validators/medicalid/longform_label_test.go
@@ -1,0 +1,169 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package medicalid
+
+import (
+	"strings"
+	"testing"
+)
+
+// A label's concatenated spelling must reach the MRN scan.
+//
+// A keyword list holding only the SUB-PHRASES of a label cannot match the label itself: the
+// whole-word rule correctly refuses "medical record" inside "medicalrecordnumber", because "number"
+// continues the word, and equally refuses "record number", because "medical" precedes it. So
+// `medicalRecordNumber: 4839271` carried no medical context at all, hasMedicalContext returned false,
+// and the MRN regex never ran — while `medical_record_number: 5729183` reported MRN at HIGH 90.
+// Measured, both at HEAD (#408).
+//
+// evaluateMRN accepted the value all along; it was simply never reached, which is why this is a gate
+// bug rather than a scoring one.
+func TestConcatenatedMedicalLabelsReachTheMRNScan(t *testing.T) {
+	// Every separator spelling of the same label, plus the sibling labels an EHR or ORM export
+	// writes. A JSON or ORM field name has no spaces to match.
+	for _, line := range []string{
+		"medicalRecordNumber: 4839271",
+		"MedicalRecordNumber: 4839272",
+		"medicalrecordnumber: 4839274",
+		"medical_record_number: 5729183",
+		"medical-record-number: 5729184",
+		"medical record number: 5729185",
+		"patientIdentificationNumber: 4839275",
+		"healthRecordNumber: 4839276",
+		"hospitalRecordNumber: 4839281",
+		"patientId: 4839277",
+		"patientNumber: 4839280",
+		"chartNumber: 4839278",
+		"admissionNumber: 4839279",
+	} {
+		t.Run(line, func(t *testing.T) {
+			v := NewValidator()
+			matches, err := v.ValidateContent(line, "record.json")
+			if err != nil {
+				t.Fatalf("ValidateContent: %v", err)
+			}
+			if len(matches) == 0 {
+				t.Errorf("no finding: the label carries no recognised medical context, so the MRN "+
+					"regex never ran. An unreported identifier is never redacted either, so it stays "+
+					"cleartext in a file the scan exits 0 on.\n  line: %q", line)
+			}
+		})
+	}
+}
+
+// The gate decides whether a line is a MEDICAL setting, so a generic record identifier must not
+// enter it. This is the direction that would cost precision, and it is the reason a bare
+// "record number" is deliberately absent from hasMedicalContext even though it IS in the MRN
+// keyword list — that list only applies once the line is already medical.
+func TestGenericIdentifierLabelsAreNotMedicalContext(t *testing.T) {
+	for _, line := range []string{
+		"record number: 1234567",
+		"recordNumber: 1234567",
+		"orderNumber: 1234567",
+		"invoiceNumber: 1234567",
+		"trackingNumber: 1234567",
+		"serialNumber: 1234567",
+		"accountNumber: 1234567",
+		"referenceNumber: 1234567",
+		"confirmationNumber: 1234567",
+		"ticketNumber: 1234567",
+	} {
+		t.Run(line, func(t *testing.T) {
+			v := NewValidator()
+			matches, err := v.ValidateContent(line, "orders.csv")
+			if err != nil {
+				t.Fatalf("ValidateContent: %v", err)
+			}
+			if len(matches) > 0 {
+				var got []string
+				for _, m := range matches {
+					got = append(got, m.Type+"="+m.Text)
+				}
+				t.Errorf("reported %v on a line with no medical context. hasMedicalContext is what "+
+					"decides whether the line is medical at all, so a generic identifier label must "+
+					"not be in it.\n  line: %q", got, line)
+			}
+		})
+	}
+}
+
+// The whole-token boundary must survive the widening. A longer word that merely CONTAINS the label
+// is not the label, and this is what stops the flexible-separator matching from becoming a
+// substring search.
+func TestConcatenatedLabelStillRespectsWordBoundaries(t *testing.T) {
+	for _, kw := range []string{"medical record number", "patient id", "chart number"} {
+		for _, text := range []string{
+			"nonmedicalrecordnumbers: 1",
+			"medicalrecordnumbering: 1",
+			"xpatientidx: 1",
+			"chartnumbering: 1",
+		} {
+			if containsLabel(text, kw) && strings.Contains(text, "ing") {
+				t.Errorf("containsLabel(%q, %q) = true: a longer word that contains the label is "+
+					"not the label, and matching it would turn every keyword into a substring "+
+					"search", text, kw)
+			}
+		}
+	}
+
+	// And the two-word form must still refuse the three-word concatenation, which is the asymmetry
+	// that makes the explicit long-form entries necessary rather than redundant.
+	if containsLabel("medicalrecordnumber: 1", "medical record") {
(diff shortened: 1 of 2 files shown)


sectoramen:
## Complexity and timing (TEST_PLAN §4) — omitted from the original body, added here

`internal/goldencorpus` `TestValidatorComplexityIsSubQuadratic/medicalid` passes (it runs as part of
`go test ./...`, so it was already green — I simply did not report it).

**Miss path — the cost on an ordinary document, where neither build finds anything. This is the number
that matters, because it applies to every file that is not full of medical labels:**

| input | main | this PR | delta |
|---|---|---|---|
| 15,000 prose lines | 0.61s | 0.63s | **+3%** |
| 60,000 prose lines | 2.33s | 2.39s | **+3%** |

+3% is the price of `containsLabel`'s separator-flexible matching over `containsKeyword` on two
per-line gates, across a list ~10 entries longer.

Scaling curve, dense medical labels (4 per line), best of three:

| lines | main | this PR | main ratio | PR ratio |
|---|---|---|---|---|
| 4,000 | 0.23s | 0.36s | – | – |
| 8,000 | 0.40s | 0.65s | 1.75x | 1.81x |
| 16,000 | 0.78s | 1.28s | 1.95x | 1.97x |

**Linear, same slope as main.** The ~60% constant factor on this input is not a regression — it is the
work the fix exists to do:

| lines | main findings | this PR |
|---|---:|---:|
| 4,000 | **0** | 16,000 |
| 16,000 | **0** | 64,000 |

main is fast here because it finds nothing. TEST_PLAN §4 names this case: a measured regression is
acceptable when it is "the unavoidable cost of correct behavior (e.g. emitting findings that were
previously dropped)". Stating it explicitly, as that section requires.

sectoramen:
## `make pr-checklist BASE=origin/main` — run, and the manual dimensions completed

I had not run the repo's own checklist before opening this PR. Running it named five NEEDS-MANUAL
dimensions; three of them I genuinely had not done. All are now done and reported here.

```
1 gofmt / go vet / go build     RAN — clean / clean / ok
2 full suite (-count=1)         RAN — 68 packages ok
3 golden regen                  RAN — 0 files changed
11 cross-platform vet           RAN — linux darwin windows
12 flake (3 repeats)            RAN — ./internal/validators/medicalid
13 race (whole module)          RAN — 68 packages ok
-- -short mode                  RAN — ok
-- tree clean                   RAN — this run dirtied nothing
```

### 4 — redaction, every strategy

MEDICAL_ID findings feed the redactor, so this had to be checked rather than inferred.

| strategy | output | raw values surviving |
|---|---|---:|
| `simple` | `medicalRecordNumber: [MRN-REDACTED]` | **0 of 3** |
| `format_preserving` | `medicalRecordNumber: *******` | **0 of 3** |

Every value newly reported by this PR is also removable, so the recall gain does not create a
reported-but-unredactable finding.

Same-binary round-tripping passes even when the hash inputs have changed, so the rule was generated
with the parent binary from a spelling both builds detect (`medical_record_number: 4839271`), then
`enabled: true` set, then applied with both:

| binary | result |
|---|---|

The parent's hash still silences the finding under this build, i.e. the change does not perturb the
suppression hash inputs. (Note for anyone repeating this: generated rules are `enabled: false` by
default, so an unedited file suppresses nothing and the check looks like a pass.)

### 6 — timing / O(n²)

Reported in the previous comment. Non-vacuity and distinct values both satisfied: 16,000 / 64,000
findings at the dense sizes, and the fixture values are randomised rather than repeated (identical
repeats let a quadratic measure as linear).

| format | rc | bytes | MRN present |
|---|---:|---:|---:|
| text | 0 | 931 | 3 |
| json | 0 | 2343 | 12 |
| yaml | 0 | 1996 | 12 |
| csv | 0 | 268 | 3 |
| junit | 0 | 645 | 2 |
| gitlab-sast | 0 | 4026 | 24 |
| sarif | 0 | 5422 | 19 |

### 10 — non-vacuity

Satisfied more strongly than by reverting: five mutants, all killed, including the one that matters
for precision (admitting a bare `record number` as medical context).

### 8 — web / CSP

N/A, no web, template or asset file touched.

rustic (approved):

PR #469, opened 2026-08-23
refactor(preprocessors): remove unused manager facade

### LINKED ISSUE #384: PreprocessorManager is unreferenced by any code or test, while a development doc still names it as the file-routing component
## What happens

`grep -rn "PreprocessorManager|NewPreprocessorManager|GetAvailablePreprocessors|GetPreprocessor(" --include="*.go" .` returns only lines 243-314 of internal/preprocessors/preprocessor.go itself — zero callers, tests included. Dead symbols: `PreprocessorManager`, `NewPreprocessorManager`, `RegisterPreprocessor`, `GetPreprocessor`, `ProcessFile`, `GetAvailablePreprocessors`, about 70 LOC.

The near-miss grep hits are all different types, which is why this survived: internal/router/file_router.go:87 defines `func (fr *FileRouter) RegisterPreprocessor` (called from internal/router/integration.go:13-69), and the `.ProcessFile(` hits at internal/parallel/worker_pool.go:283 and internal/router/file_router.go:272 are the two-argument `FileRouter.ProcessFile(path, ctx)`. Because internal/preprocessors is under `internal/`, no out-of-module consumer can reach it either, so the exported-API escape hatch does not apply.

## Secondary hazard

internal/router/file_router.go:381-389 wraps `processor.Process(filePath)` in a `recover()` that turns a preprocessor panic into an error. The dead `PreprocessorManager.ProcessFile` calls `preprocessor.Process` bare at :294. Since nothing calls it, that is a latent hazard for whoever wires it, not a live crash path.

## Why the doc matters

The dead type is propped up by a stale doc: docs/development/text_extraction_integration.md:17 still lists "Provides PreprocessorManager for managing multiple preprocessors / Handles file type detection and routing" as the architecture, when routing is internal/router.FileRouter, and :20 names internal/preprocessors/text_preprocessor.go, which does not exist (`ls internal/preprocessors/*.go` has no such file). A reader following that doc would wire the panicking, unbounded path.

## Fix sketch

Delete lines 243-314 of internal/preprocessors/preprocessor.go (the type and all five methods); the `Preprocessor` interface and `ProcessedContent` stay, they are what internal/router uses. In the same change, correct docs/development/text_extraction_integration.md:15-23 to name internal/router.FileRouter as the router and drop the reference to the non-existent internal/preprocessors/text_preprocessor.go, so the deleted type is not resurrected from the doc. If someone instead wants to keep a manager facade, it must get the `recover()` wrapper that file_router.go:381-389 has before any caller is added.

## Failing test

There is no behavioural test for unreachable code, and this issue does not pretend otherwise. The check that changes state is static: today `grep -rn "PreprocessorManager" --include="*.go" .` returns hits only in the defining file (equivalently `go run golang.org/x/tools/cmd/deadcode ./...` lists all five methods); after the delete the package must still build and `go test ./internal/preprocessors/... ./internal/router/...` must stay green. A docs test asserting every Go path named in docs/development/text_extraction_integration.md exists WOULD fail at HEAD, on internal/preprocessors/text_preprocessor.go.

Caveat for whoever picks this up: a repo-wide dead-code audit exists as a working note (report-only, ~242 confirmed-dead funcs, PR-D1..PR-D5 sequencing) and this symbol would plausibly sit in its misc bucket, but that note does not name PreprocessorManager and its raw digest is no longer on disk, so overlap could not be confirmed. It has no GitHub issue behind it.

pttydou:
*Issue #, if available:*

#384

*Description of changes:*

Removes the unreferenced `PreprocessorManager` facade and documents `FileRouter` as the component responsible for preprocessor registration and routing. The existing `TextPreprocessor` remains unchanged.

Tests:
- `GOTOOLCHAIN=go1.26.7 go test -count=1 ./internal/preprocessors/... ./internal/router/...`
- `GOTOOLCHAIN=go1.26.7 go vet ./...`
- `GOTOOLCHAIN=go1.26.7 go build ./...`
- Serialized full and short suites passed after excluding two failures reproduced on `origin/main`: unavailable Windows symlink privileges and the intermittent zero-duration timing assertion.

Local race testing could not run because CGO is disabled; the existing CI matrix covers race on all three hosted platforms.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

---

_Maintainer edit: adding a closing reference so the issue resolves on merge._

Closes #384

### DIFF
--- a/docs/development/text_extraction_integration.md
+++ b/docs/development/text_extraction_integration.md
@@ -14,15 +14,18 @@ The text extraction preprocessors allow Ferret-scan to analyze the content of do
 
 1. **Preprocessor Interface** (`internal/preprocessors/preprocessor.go`)
    - Defines the `Preprocessor` interface for all preprocessors
-   - Provides `PreprocessorManager` for managing multiple preprocessors
+   - Defines the shared `ProcessedContent` model
+
+2. **File Router** (`internal/router/file_router.go`)
+   - Registers and coordinates multiple preprocessors
    - Handles file type detection and routing
 
-2. **Text Preprocessor** (`internal/preprocessors/text_preprocessor.go`)
+3. **Text Preprocessor** (`internal/preprocessors/text_preprocessor.go`)
    - Implements text extraction for PDF and Office documents
    - Uses existing text extraction libraries
    - Returns structured `ProcessedContent`
 
-3. **Updated Validator Interface** (`internal/detector/detector.go`)
+4. **Updated Validator Interface** (`internal/detector/detector.go`)
    - Added `ValidateContent()` method for processing extracted text
    - Maintains backward compatibility with existing `Validate()` method
 
--- a/internal/preprocessors/preprocessor.go
+++ b/internal/preprocessors/preprocessor.go
@@ -3,11 +3,7 @@
 
 package preprocessors
 
-import (
-	"path/filepath"
-
-	"github.com/awslabs/ferret-scan/v2/internal/observability"
-)
+import "github.com/awslabs/ferret-scan/v2/internal/observability"
 
 // ProcessedContent represents content that has been processed by a preprocessor
 type ProcessedContent struct {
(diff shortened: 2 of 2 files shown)


sectoramen:
## Review: verified locally against `origin/main` (0610b7e) — safe to land

### Reachability

At `origin/main`, `PreprocessorManager` and its five methods appear only in the defining file and the one doc line this PR corrects. No callers anywhere, production or test. Two things make that a compile-level result rather than a grep-level one:

- `go vet ./...` is clean, and vet type-checks `_test.go` files, so a test-only caller would have surfaced.
- Cross-platform vet (linux, darwin, windows) is also clean, so there is no build-tagged caller hiding behind a GOOS this platform never compiles.

After the change there are zero references to the symbol repo-wide, including in Markdown — the doc edit here was the last one.

### Test plan

`make pr-checklist BASE=origin/main` on darwin/arm64: **10 dimensions ran, 0 failed.**

| Dimension | Result |
|---|---|
| gofmt | clean |
| `go vet` | clean |
| `go build ./...` | ok |
| full suite `-count=1` | 69 packages ok |
| `-race`, whole module | 69 packages ok |
| golden regen | 0 files changed |
| cross-platform vet | linux, darwin, windows |
| flake, 3 repeats | `./internal/preprocessors` ok |
| `-short` | ok |
| tree clean | dirtied nothing |

I got no failures at all on this platform, including the two you excluded — and excluding the zero-duration timing assertion was correct: that flake is being sized properly in #453, independently of this change. Race did run here, so the CI matrix is not the only coverage.

The four `NEEDS-MANUAL` items pr-checklist prints (redaction, suppression, timing/O(n^2), all-7-formats) are flagged because one production `.go` file changed. Deleting unreachable code cannot reach any of those sinks, so they are vacuous for this diff.

### Conflicts

None, at both levels:

- **Textual:** the branch sits on current `origin/main`, and there is no file overlap with any of the other 13 open PRs. The nearest neighbour in this package, #455, touches `image_metadata_preprocessor.go`.
- **Semantic:** I fetched all 13 open PR heads and grepped each for a reference to the deleted symbols. None introduces one, so there is no clean-merge-that-does-not-compile risk from a sibling PR adding a caller while this one removes the callee.

### It also removes a trap, not just bytes

The deleted `ProcessFile` called `preprocessor.Process` bare. The live path wraps that call in a `recover()` so one panicking preprocessor cannot take down a whole scan (`internal/router/file_router.go`). Deleting the facade removes the unguarded copy instead of leaving it for whoever eventually wired it up.

### Doc pointers spot-checked at HEAD

All four component bullets now resolve: `FileRouter.RegisterPreprocessor` (`internal/router/file_router.go:87`, seven callers in `internal/router/integration.go`), `ProcessedContent` (`internal/preprocessors/preprocessor.go:13`), `ValidateContent` (`internal/detector/detector.go:43`), and `internal/preprocessors/text_preprocessor.go`, which does exist — thanks for keeping that bullet instead of acting on the incorrect claim in the issue body. The correction is in the issue thread. The supported-file-types list is accurate too; it matches `TextPreprocessor.supportedExtensions` exactly.

### Optional follow-ups in the same file

No objection to merging as is — these are pre-existing and I am happy to take them separately. Listing them so they are recorded somewhere:

1. The YAML sample under "Configuration File" includes `preprocessors.text_extraction.types`, which the schema does not have. Running the tool with that exact snippet prints `Warning: unknown config key "types" — ignored (check for a typo)`. The other three keys in the sample are real: `preprocessors.text_extraction.enabled`, `defaults.enable_preprocessors`, and `profiles.<name>.enable_preprocessors`.
2. The "Verbose Output" sample shows `Preprocessed document.pdf: extracted 1250 words, 7890 characters`. That string does not exist anywhere in the tree; the actual line is `Content: %d words, %d characters` (plus `, %d pages` when known), from `cmd/main.go`.
3. "Processing Flow" step 4 still says a preprocessing failure falls back to regular file validation, which the blockquote further down the same page already contradicts. Correcting step 4 at the source beats annotating it.

### Housekeeping

I added a `Closes #384` line to the description so the issue resolves on merge; the rest of your body is untouched. Checks are currently held at `action_required` because workflow runs from a fork need maintainer approval — that is a repo setting, not a problem with the branch.

rustic (approved):

Issue #388, opened 2026-08-20
Column assignment rescans each line from offset 0 for every distinct value, so 64,000 matches on one 1.2 MB line cost 16.4 s of a 16.9 s scan no budget bounds


sectoramen:
## What happens

`detector.ResolveLineSpans` (internal/detector/line_span.go:104-105) keeps its scan cursor keyed per (line, match TEXT):

```go
from := byText[m.Text]
idx := strings.Index(line[from:], m.Text)
```

For a DISTINCT value the cursor is always 0, so every distinct match rescans the whole line from the start: O(matches x line length). Validators that dedup per line (ipaddress/validator.go:290) therefore always hit the worst case. It is reached on every scan from `detector.AssignLineColumns`, called at internal/parallel/validator_runner.go:291 inside `flatten()` after every validator has returned.

## Measured evidence

HEAD 61a85bd. IP_ADDRESS, one dense line, 2 distinct routable IPs per rep:

```
ferret-scan --file ip_one.txt --checks ipaddress --config empty.yaml --quiet
```

| bytes | matches | wall | growth per doubling |
|---|---|---|---|
| 36,682 | 2,000 | 0.09 s | |
| 73,375 | 4,000 | 0.17 s | 1.9x |
| 148,198 | 8,000 | 0.44 s | 2.6x |
| 299,005 | 16,000 | 1.44 s | 3.1x |
| 600,612 | 32,000 | 4.45 s | 3.2x |
| 1,210,311 | 64,000 | **17.33 s** | 3.9x (converging on 4x = quadratic) |

Multi-line control at the SAME 1,210,311 bytes and SAME 64,000 findings: **0.73 s** (23.7x faster).

Attribution is measured, not guessed. `--debug` at 32,000 reps reports `ipaddress_validator duration_ms 389` and `enhanced_validator_bridge 535` out of a 16.92 s run -- 97% of wall time is AFTER validation. `sample <pid> 5 10` puts 470 of 471 stacks in `parallel.RunValidators -> detector.AssignLineColumns -> detector.ResolveLineSpans -> internal/stringslite.Index`. Also reproduced with distinct SSNs: 128,000 B / 8,000 findings 0.45 s -> 512,000 B / 32,000 findings 3.33 s (7.9x net for 4x input; bridge 883 ms of 3.33 s).

## The memory claim this replaces

[...]

PR #505, opened 2026-08-26
test(complexity): measure allocation growth, not wall-clock, where a CI runner can move the number

### LINKED ISSUE #504: Two complexity guards assert wall-clock time and flake on CI while the code is linear
## Summary

Two complexity guards assert on wall-clock time in a way that fails on shared CI runners while the
code is correct. Both fired this week on PR branches whose changes are in unrelated packages:

| test | runner | reading | verdict |
|---|---|---|---|
| `TestCreateBasicLineMappingsIsLinear` (`internal/preprocessors`) | macos-latest | base=7.44ms big=91.26ms, ratio **12.3x** against a 12.0x limit | false failure |
| `TestAssignLineColumnsComplexityDistinctAndRepeated` (`internal/detector`) | windows-latest | big=**307ms** against a 250ms ceiling, while the ratio read 5.7x | false failure |

In both cases the *other* assertion in the same test correctly reported the code as linear. The
`internal/preprocessors` absolute ceiling passed with a 22x margin (91ms against 2s); the
`internal/detector` growth ratio read 5.7x where a regression is 14x.

## Root causes, which are different

**`internal/preprocessors`** — a wall-clock growth ratio guarded by a 2ms noise floor. 7.44ms is
barely above that floor, so scheduler noise moves the ratio freely. A ratio of two single-digit
millisecond samples cannot carry a signal on a shared runner.

**`internal/detector`** — a 250ms absolute ceiling, the **only sub-second wall-clock ceiling in the
repo** (every other complexity guard is 2–20s). It was set ~10x above what the author's host
measured, which is not enough headroom for a runner 12x slower. An absolute wall-clock bound cannot
be both slow-machine-safe and fast-machine-sensitive: a regression costs ~423ms locally, so a
ceiling loose enough for the Windows runner would no longer catch it locally.

## What was measured

`internal/preprocessors` — the defect is `CalculateAbsoluteOffset` being called per line, and it
re-splits the whole text on every call, allocating one string header per line. That makes it
O(n²) in **allocated bytes**, not just in time, so allocation is a clock-free proxy. Verified by
restoring the two per-line calls:

```
                    2500 lines   10000 lines   ratio
prefix sum (now)        1,716KB       7,958KB   4.6x
per-line split (was)  493,290KB  13,014,252KB  26.4x
```

Under `GOMAXPROCS=1` the figures are byte-identical (1,716KB / 7,957KB), which is the property the
wall clock lacks.

`internal/detector` — the defect (`p := next[text]` → `p := 0` in `resolveByIndex`) is pure index
stepping and does not allocate, so the scale-free growth ratio has to stay the discriminator:

```
                       base       big     ratio
with the pointer      ~4.0ms   ~19.0ms   4.6-4.9x   (this host)
                      ~3.9ms   ~21.0ms   5.2-5.9x   (GOMAXPROCS=1)
                      54.1ms    307.2ms       5.7x   (windows-latest, the false failure)
without it            29.3ms    423.0ms  13.6-14.4x
```

## Fix

- `internal/preprocessors`: replace the wall-clock ratio with an allocation-growth ratio at 8x.
  Keep the 2s absolute ceiling.
- `internal/detector`: make the growth ratio the discriminator at 9x (evenly splitting a
  worst-observed 5.9x from a measured 13.6x), and raise the 250ms ceiling to the file's existing
  `columnsAbsoluteCeiling` (4s) as a catastrophe net rather than a discriminator.

Both fixes were mutation-verified: with the regression restored the tests fail at 26.4x and 14.3x
respectively.

## Latent, not changed here

These also assert wall-clock ratios, but with looser thresholds and generous ceilings, and none has
been observed to flake. Listing them so the pattern is on record rather than silently accepted:

- `internal/preprocessors/text-extractors/text-extract-officetextlib/formfield_onepass_test.go:179` (8.0)
- `internal/preprocessors/meta-extractors/meta-extract-videolib/dms_coordinate_test.go:373` (8.0)
- `internal/preprocessors/embedded_media_complexity_test.go:215` (6.0)
- `internal/goldencorpus/complexity_guard_test.go:352` (12.0)
- `internal/detector/line_span_complexity_test.go:240` (8.0)

`internal/scorecorpus/timing_test.go` is the pattern to copy: its ratio is over a **true-positive
count**, not a duration, so it is immune to runner speed.

## Also seen, not a code problem

Three recent red runs were GitHub infrastructure, failing at the `Set up job` step with no test
output: `main` (Go Tests, macos-latest), and CodeQL `Analyze (javascript-typescript)` on #489.
Nothing to fix in the repo.

sectoramen:
## What

Two complexity guards asserted on wall-clock time in a way a shared CI runner can move, and both
failed this week **while the code they guard is linear**. Neither failure was on a branch that
touches the guarded package.

| test | runner | reading | the other assertion in the same test |
|---|---|---|---|
| `TestCreateBasicLineMappingsIsLinear` (`internal/preprocessors`) | macos-latest | 12.3x ratio vs a 12.0x limit, base=7.44ms | absolute ceiling passed with a **22x margin** (91ms vs 2s) |
| `TestAssignLineColumnsComplexityDistinctAndRepeated` (`internal/detector`) | windows-latest | **307ms** vs a 250ms ceiling | growth ratio read **5.7x**, correctly linear |

## Why each flaked, and why the two fixes differ

**`internal/preprocessors`** — a wall-clock growth ratio behind a 2ms noise floor. 7.44ms is barely
above that floor, so noise moves the ratio freely.

The defect it guards is `CalculateAbsoluteOffset` being called per line, and that function re-splits
the **entire** text on every call, allocating one string header per line. So the regression is
O(n²) in *allocated bytes*, not merely in time — which makes allocation a clock-free proxy.
Measured by restoring the two per-line calls:

```
                      2500 lines   10000 lines   ratio
prefix sum (now)          1,716KB       7,958KB   4.6x
per-line split (was)    493,290KB  13,014,252KB  26.4x
```

The limit sits at 8x, between a measured 4.6x and a measured 26.4x, with the absolute figures
differing by ~1,600x. Under `GOMAXPROCS=1` the readings are byte-identical (1,716KB / 7,957KB) —
the property the wall clock lacks. The 2s absolute ceiling is kept.

**`internal/detector`** — a 250ms absolute ceiling: the **only sub-second wall-clock ceiling in the
repo**, where every other complexity guard allows 2–20s. It sat ~10x above the author's host, which
is not enough headroom for a runner 12x slower. An absolute wall-clock bound cannot be both
slow-machine-safe and fast-machine-sensitive — a regression costs ~423ms locally, so a ceiling loose
enough for the Windows runner would stop catching it locally.

That defect (`p := next[text]` → `p := 0` in `resolveByIndex`) is pure index stepping and does not
allocate, so the scale-free ratio has to stay the discriminator. Re-derived at HEAD:

```
                       base       big     ratio
with the pointer      ~4.0ms   ~19.0ms   4.6-4.9x   (this host)
                      ~3.9ms   ~21.0ms   5.2-5.9x   (GOMAXPROCS=1)
                      54.1ms    307.2ms       5.7x   (windows-latest, the false failure)
without it            29.3ms    423.0ms  13.6-14.4x
```

The ratio limit moves 8x → 9x, splitting the populations evenly: 1.53x headroom above the worst
correct reading, 1.51x below the best regressed one. (At 8x the headroom above 5.9x was 1.36x, and
5.9x came from the constrained-CPU run — the case most like the runner that flaked.) The ceiling
becomes the file's existing `columnsAbsoluteCeiling` (4s), documented as a catastrophe net rather
than a discriminator.

## Non-vacuity (dimension 10)

Both guards were mutation-verified, and both mutations reverted with the production file confirmed
clean by `git diff --quiet`:

| mutation | restored defect | result |
|---|---|---|
| `lineOffsets[lineNum]` → two per-line `CalculateAbsoluteOffset` calls | the O(n²) re-split | **FAIL at 26.4x** |
| `p := next[text]` → `p := 0` | the un-carried occurrence pointer | **FAIL at 14.3x** |

The second was re-run *after* loosening the ceiling to 4s, to prove the ratio alone still catches it.

## Not changed here

Five other tests assert wall-clock ratios, with looser thresholds and generous ceilings, and none has
been observed to flake — listed in the issue so the pattern is on record rather than silently
accepted. `internal/scorecorpus/timing_test.go` is the pattern to copy: its ratio is over a
true-positive **count**, so it is immune to runner speed.

Three other recent red runs were GitHub infrastructure, failing at `Set up job` with no test output
(`main` Go Tests on macos-latest; CodeQL `Analyze (javascript-typescript)` on #489). Nothing to fix
in the repo.

## Test plan

```
=== scope ===
  CHANGELOG.md
  internal/detector/line_span_complexity_test.go
  internal/preprocessors/line_mapping_offsets_test.go

=== dimensions ===
2 full suite (-count=1)            RAN — 74 packages ok
10 non-vacuity                     RAN — mutation-verified both ways, see the table above
12 flake (3 repeats)               RAN — ./internal/detector ./internal/preprocessors
13 race (whole module)             RAN — 74 packages ok

=== summary: 10 ran, 0 failed ===
```

Additionally: 10 consecutive runs of each test, and 3 runs of each under `GOMAXPROCS=1`, all green.

Closes #504.


sectoramen:
### The 250ms ceiling has now failed twice, and best-of-N was already the fix attempted for it

Correcting an implication in the description above: the windows-latest failure was not the first
occurrence. The `bestAssign` doc comment on `main` records an earlier one, and the numbers are
nearly identical:

| when | runner | base | big | ceiling |
|---|---|---|---|---|
| earlier (recorded in `bestAssign`'s comment) | macOS | 40.67ms | **306.40ms** | 250ms |
| this week (#502's branch) | windows-latest | 54.09ms | **307.22ms** | 250ms |

Two different OSes, two different base readings, and a big term landing within 1ms of each other —
both just over a limit set at 250ms.

This matters for the choice of fix. The response to the first failure was to introduce `bestAssign`
(best-of-3, applied to all four targets) on the theory that the failure was "a spike in one sample"
that a minimum-of-N would remove. Best-of-3 was already in force when this week's failure happened,
so that theory is now falsified: the minimum of three attempts on a shared runner is ~307ms, which
is simply what these runners cost. The ceiling was not measuring the algorithm.

That is the argument for making the scale-free ratio the discriminator rather than tightening the
sampling again — both terms of a ratio are measured on the same machine in the same run, so a
uniformly slower runner cancels out, and the measured separation (worst correct 5.9x against
regressed 13.6-14.4x) does not depend on the runner's speed at all.

`bestAssign` itself is kept exactly as it is — best-of-N is still the right way to compare two
durations, and it is what keeps the ratio's numerator and denominator comparable.

sectoramen:
### A third failure in this family, and the second for `TestAssignLineColumnsComplexityDistinctValues`

Pushed [`79d7f72`](../commit/79d7f72). While this PR was open, the same family failed again — on
**windows-latest**, at **9.7x against an 8.0 growth limit** (`base=12.0518ms big=116.6421ms`), on
[#507](https://github.com/awslabs/ferret-scan/pull/507)'s branch, which does not touch
`internal/detector` at all.

This is a **different** test from the two the description covers, and it is one of the five I listed
as "latent, not changed here". It is no longer latent.

### The sampling theory is now falsified twice

That test's own comment records its **first** Windows failure at 8.9x, whose fix was to grow the
sample from 2000/8000 to 8000/32000 on the theory that the base was too small to measure. That did
stabilise the local spread to 3.76x–4.28x. It failed anyway:

|  | base | big | ratio |
|---|---|---|---|
| audit host, 6 runs | 2.2–2.5ms | 9.3–10.4ms | **3.7–4.6x** |
| `GOMAXPROCS=1`, 4 runs | 2.3–2.9ms | 10.0–12.1ms | **4.0–4.3x** |
| windows-latest (earlier) | 40.67ms | ~362ms | **8.9x** ← failed 8.0 |
| windows-latest (this run) | 12.05ms | 116.64ms | **9.7x** ← failed 8.0 again |
| regressed (measured) | 141.1ms | 2.184s | **15.5x** |

The threshold sits **below its own correct population** on Windows. Windows is not noisy around 4x
here — it sits systematically at 9–10x, because its `base` is 4.8x slower than this host's while its
`big` is 12x slower, so the ratio itself shifts. Sampling cannot fix a threshold that is on the wrong
side of the population.

The structural cause is the **fixture**: `buildDistinctOnOneLine` grows the value count *and* the
line length together, so correct is 4x and the regression is 16x. A 4x-wide window has to absorb
every machine's constant-factor differences, and Windows' exceed it.

### Fix: measure the axis the implementation claims independence from

`resolveByIndex` documents the property worth pinning outright — *"The cost is therefore independent
of how MANY values there are, which is the term that made rescanning quadratic."* Measuring that
directly puts the correct population at **~1.0x**, where a uniformly slower machine cancels out
rather than shifting the answer.

New fixture holds the line at 96,000 values and varies only the match count, 2000 → 8000. Values are
**spread** across the line rather than taken from its head, because `resolveByRescan`'s per-value
cursor starts at zero and a head-first subset would sit in the cheapest part of the line and
understate the regression.

|  | base | big | ratio |
|---|---|---|---|
| correct, 5 runs | 9.4–10.2ms | 11.0–12.3ms | **1.08–1.26x** |
| correct, `GOMAXPROCS=1`, 3 runs | 8.8–9.8ms | 11.2–12.6ms | **1.18–1.32x** |
| regressed, 2 runs | 405–423ms | 1.63–1.66s | **3.92–4.03x** |

**2.2 sits 1.67x above the worst correct reading and 1.78x below the best regressed one.** A retuned
version of the old ratio cannot match that — 12.5 between 9.7x and 15.5x buys only 1.29x and 1.24x.
The absolute figures also separate by **41x at the same match count** (10ms against 415ms), which is
the check that this pair measures the regression and not sampling noise.

The old pair's ratio is kept as a **log** with both populations recorded, and its 4s absolute ceiling
is kept as the catastrophe net.

### Verification

Mutation-verified by forcing `useIndex` to return false, restoring the exact `O(K × line length)`
shape #388 removed: the new assertion fails at **4.11x**. `go vet` was run with the mutation applied
*before* believing anything, because a build error reports zero test failures — a trap that produced
a false "SURVIVED" earlier in this same PR. Reverted with `git diff --quiet` confirming clean.

5 consecutive runs of `./internal/detector` green, full suite 75 packages, `-race` clean, and the
rewritten test costs 0.35s. `make pr-checklist BASE=origin/main`: 10 ran, 0 failed.

An exhaustive audit of the remaining wall-clock guards in the repo is running now; if it finds others
whose threshold sits inside its own correct population, I will report them on
[#504](https://github.com/awslabs/ferret-scan/issues/504) rather than expanding this PR further.

rustic (approved):

Issue #577, opened 2026-09-02
test(router): four tests create scratch directories inside the repo tree, flaking any concurrent tree walk


sectoramen:
Four tests in `internal/router` call `os.MkdirTemp(".", …)`, which creates the scratch directory **inside the repo working tree** rather than under the OS temp dir:

- `internal/router/structured_sections_test.go:236` — `structured-sections-`
- `internal/router/empty_extraction_visibility_test.go:99` — `empty-extraction-`
- `internal/router/attribution_provenance_test.go:37` — `attribution-provenance-`
- `internal/router/determinism_test.go:152` — `determinism-fixture-`

## Observed failure

Under `go test ./...` these directories appear and are removed while the `cmd` documentation guards walk `../internal` concurrently. Measured on a clean tree (the failing branch touched only markdown and added an unrelated test):

```
--- FAIL: TestTheEnvGuardTellsAnAdvertisementFromCorrectProse (1.72s)
    --- FAIL: TestTheEnvGuardTellsAnAdvertisementFromCorrectProse/the_real_pre-commit_variable (0.08s)
        documented_env_vars_test.go:481: walk ../internal: open ../internal/router/structured-sections-3058054452: no such file or directory
FAIL	github.com/awslabs/ferret-scan/v2/cmd	49.687s
```

The failing subtest is fed **synthetic markdown**, so its own input is irrelevant to the failure — it fails inside the shared `evgProductionNames` helper, which walks `../internal` and `t.Fatalf`s on any walk error. The `cmd` package passes standalone twice (29.8s each) and fails only under full-suite contention (49.7s), which is why this reads as an unrelated flake on whichever branch happens to hit it.

## Why it is worth fixing rather than tolerating

1. **It flakes CI on branches that do not touch either package.** The blame lands on the wrong change.
2. **A panicking or killed test leaves the directory behind in the working tree**, where it can dirty `git status` and be committed by a `git add -A`.
3. `t.TempDir()` is the correct call and is already used elsewhere in the repo. It is removed automatically, is outside the tree, and is per-test.

## Two halves

- **Root cause:** replace the four `os.MkdirTemp(".", …)` calls with `t.TempDir()`. Check whether any of them depend on being relative to the package directory — if one does, that dependency is itself worth recording, because it is what would break a move to `t.TempDir()`.
[...]

Issue #564, opened 2026-08-31
The 'scan' subcommand survives in JSON args arrays, which the #559 invocation guard cannot see


sectoramen:
## Summary

#556 removed the non-existent `scan` subcommand from 88 documented command lines, and #559 added `cmd/documented_invocations_test.go` to keep it out. Four sites survive, because they are not shell command lines and the guard's regex only recognises those.

## Measured at `a0e983c`

`docs/development/WINDOWS_DEVELOPMENT.md:162`, inside a VS Code `launch.json` block:

```json
"args": ["scan", ".", "--debug"],
```

There is no `scan` subcommand. Copied into a real `launch.json`, this stats a path named `scan`, and because Go's `flag` package stops parsing at the first positional argument, `--debug` is taken as a path too and ignored — the same mechanism #556 describes, in a form the guard cannot see.

The guard matches `ferret-scan` followed by whitespace, so a JSON argument array never matches. `TestNoDocumentedInvocationSilentlyDropsItsFlags` is green with this present.

## Suggested direction

Two parts, and the second is the one that matters:

1. Fix the sites: `["--file", ".", "--debug"]`.
2. **Teach the guard about argument arrays**, so this class cannot return. A second extractor for `"args": [...]` blocks whose first element is not a flag, keyed off the surrounding `launch.json`/`tasks.json`/`docker-compose`/`k8s` shape. Without this, the fix is one-off and the next `args` block reintroduces it — which is exactly what happened between #556 and now.

Worth checking the same shape in `docs/DOCKER_INTEGRATION.md` (its `args:` at line 476 is correct today — `["--file", "/data", ...]` — so it makes a good positive control for the widened guard).

Found while auditing the #555/#556 family.

Issue #555, opened 2026-08-29
docs: two more files advertise absent ferret-scan flags, and 13 FERRET_* env vars are never read


sectoramen:
Found while closing #503. That issue was filed against one file; the same fiction exists in two more, and a related fiction exists in the environment-variable documentation.

`#503`'s PR adds a guard — `TestDocumentedFlagsExist` in `cmd/documented_flags_test.go` — which asserts that every flag advertised on a `ferret-scan` command line in `docs/` is a flag the binary actually registers. The two files below are listed in that test's `knownGaps` map pointing at this issue, so they are tracked rather than silently exempt. A companion test (`TestKnownFlagGapsAreStillGapsSoTheListCannotRot`) fails if a listed file becomes clean, so fixing either one requires removing its entry in the same change.

## 1. `docs/troubleshooting/enhanced-architecture-runbook.md` — 10 absent flags

525 lines, and **it is orphaned**: nothing in `docs/README.md` links to it (the Troubleshooting section lists only `GITLAB_INTEGRATION_TROUBLESHOOTING.md`).

Absent flags, each verified by running the binary (`rc=2`, `flag provided but not defined`) and by grepping `cmd/` for a registration:

```
--benchmark            --document-path-only   --health-check         --metadata-path-only
--diagnostic-report    --list-preprocessors   --preprocessor-only    --test-component
--test-routing         --test-validation
```

This file is a near-duplicate of the guide #503 deleted, for the same subsystem. It was deliberately left untouched so #503 stayed reviewable, but it very likely carries the same config-key and log-message fiction too — it has not been audited beyond the flags.

**Suggested direction:** audit it the way #503's file was audited, and expect the same answer. Since it is unreferenced, deletion is cheaper than correction; check first whether any of it is real and uncovered elsewhere.

## 2. `docs/development/file-router-metadata-capabilities.md` — 1 absent flag

`--dry-run`. Small, and correctable in place.

## 3. Thirteen `FERRET_*` environment variables that are never read

`#503`'s PR removed `FERRET_VERBOSE` from the four docs that advertised it, because it is the one I verified behaviourally: it is read nowhere in the tree, and output is byte-identical with and without it (11 lines either way, versus 26 for `--verbose`).

The following are also claimed in `docs/` and do **not** appear in any `os.Getenv` call in the repository. I have **not** behaviourally tested these individually — the list is a search result, so treat each as a candidate needing the same check before editing:

```
[...]

Issue #503, opened 2026-08-26
content-routing-troubleshooting.md: 29 of 31 config keys and all 17 flags it documents do not exist


sectoramen:
## Summary

`docs/development/content-routing-troubleshooting.md` (751 lines) is a troubleshooting guide whose
diagnostic procedures cannot be followed. Audited against the binary and `internal/config`:

- **29 of the 31 config keys it shows do not exist.** Only `enabled` and `window` are real.
- **All 17 CLI flags it shows are absent.** The binary has 39 flags; none of these are among them.

This came out of #411, which flagged two of the keys and asked for the file to be checked the same
way. The check found the problem is the whole document, not two rows.

## Measured

Config sections, verified by running the real binary with each example pasted into a config file:

```
$ ferret-scan --config cr.yaml --file probe.txt
Warning: unknown config key "parallel_processing" — ignored (check for a typo)
Warning: unknown config key "content_router" — ignored (check for a typo)
```

Absent config keys, as advertised by the guide:

`cache_size`, `cache_ttl`, `camera_serial`, `confidence_boosts`, `confidence_threshold`,
`content_routing_failure_rate`, `creator`, `device`, `document_metadata`,
`document_type_detection`, `domain_detection`, `enable_adaptive_processing`, `enable_caching`,
`enable_structured_logging`, `field_weights`, `gps`, `gpslatitude`, `image_metadata`, `level`,
`log_components`, `max_file_size`, `max_memory_usage`, `max_workers`,
`memory_usage`, `metadata_validation_accuracy`, `rules`, `section_marker`, `threshold`,
`worker_memory_limit`

Absent flags, none of which appear in `--help` or anywhere in `cmd/`:

`--compare-validation-modes`, `--content`, `--debug-confidence-scoring`,
`--debug-content-routing`, `--debug-context-analysis`, `--debug-metadata-validation`,
`--debug-preprocessors`, `--disable-content-routing`, `--dry-run`, `--legacy-validation`,
`--metadata-type`, `--monitor-memory`, `--profile-content-routing`,
`--profile-context-analysis`, `--profile-metadata-validation`, `--profile-performance`,
`--test-metadata-validator`

What *is* real: `--debug` and `FERRET_DEBUG=1`.

## Why it matters

Every "Diagnostic Steps" and "Solutions" block instructs the reader to run a command that does not
exist or set a key that is silently ignored. A reader who follows the guide and sees no change has
[...]

PR #578, opened 2026-09-02
docs(gitlab-ci): point the CI setup page at .gitlab-ci.yml instead of restating a pipeline that does not exist

### LINKED ISSUE #575: docs: GITLAB_CI_SETUP.md documents 10 CI jobs that do not exist, and omits all 8 that do
## Summary

`docs/deployment/GITLAB_CI_SETUP.md` describes a GitLab CI pipeline that does not exist. Every job name it puts in backticks is absent from `.gitlab-ci.yml`, and every job that actually runs goes unmentioned.

Found while bumping Go to 1.27.1: the doc claims a multi-version test matrix (`test:go-1.24`, `test:go-1.25`, `test:go-1.26`) that a version bump would ordinarily need to update. There is nothing to update — those jobs were never there.

## Measured at `main`

Comparing backticked job names in the doc against top-level keys in `.gitlab-ci.yml`:

| | count |
|---|---|
| job names the doc names | 10 |
| of those, absent from `.gitlab-ci.yml` | **10** |
| real jobs the doc never mentions | **8** |

**Claimed but absent:** `docker:build`, `security:dependency-scanning`, `security:sast`, `security:secret-detection`, `test:go-1.24`, `test:go-1.25`, `test:go-1.26`, `test:integration`, `test:race`, `test:unit`

**Real but undocumented:** `auto-version-tag`, `build`, `ferret-sast`, `gemnasium-python-dependency_scanning`, `goreleaser-release`, `gosec-sast`, `manual-release-tag`, `prepare-python`

So the document has a section headed "Multi-version Testing" describing compatibility coverage the pipeline does not provide, and a "Security Jobs" section naming three scanners under names that do not exist while the two real ones (`gosec-sast`, `gemnasium-python-dependency_scanning`) are unnamed.

## Why this matters beyond tidiness

A reader deciding whether Go-version compatibility is covered gets "yes, three versions" from this document and the answer is no. That is the same class of defect as #503, #555 and #556 — documentation asserting a capability the tool does not have — and the same remedy applies: correct it against the file, and consider a guard.

## Suggested direction

1. Rewrite the job inventory from `.gitlab-ci.yml`, naming the eight real jobs and deleting the ten that do not exist. Do **not** merely refresh the version numbers in `test:go-1.24/1.25/1.26`; that would make the fiction more current.
2. Decide whether a multi-version matrix is actually wanted. If it is, add it to `.gitlab-ci.yml` and then the doc becomes true; if it is not, say plainly that CI builds with the single version in `.go-version`.
3. Consider extending the docs-fiction guard family: `cmd/documented_flags_test.go` and `cmd/documented_invocations_test.go` already assert that documented flags and invocations exist. A guard asserting that backticked job names in `docs/deployment/*.md` appear as keys in `.gitlab-ci.yml` would cover this class. Note the trap the Dockerfile check just hit (fixed in the Go 1.27.1 change): anchor the match to real declarations, not to the first occurrence anywhere in the file, or the guard validates a comment.

## Notes

Not fixed in the Go 1.27.1 bump deliberately — that change touches version pins, and rewriting a fictional CI inventory is a different piece of work with a different reviewer question.

sectoramen:
Closes #575.

## The problem

`docs/deployment/GITLAB_CI_SETUP.md` documented the pipeline by inventory — one `#### \`job-name\`` heading per job. **Eighteen of its nineteen sections named a job that does not exist**, and all eight real jobs went unmentioned. `build` was the only true heading.

| Documented, absent | Real, unmentioned |
|---|---|
| `docker:build`, `test:unit`, `test:integration`, `test:race`, `test:go-1.24`/`1.25`/`1.26`, `security:sast`, `security:secret-detection`, `security:dependency-scanning`, `license_scanning`, `code_quality`, `coverage`, `benchmark`, `pages`, `release` + 4 troubleshooting sections | `build`*, `prepare-python`, `gosec-sast`, `ferret-sast`, `gemnasium-python-dependency_scanning`, `goreleaser-release`, `auto-version-tag`, `manual-release-tag` |

<sub>*`build` was the one job the doc got right.</sub>

The surrounding prose was fiction of the same kind: GitLab Pages hosting at a named URL, container-registry publishing, merge-request coverage and code-quality widgets, JUnit test reports, licence-compliance gating and benchmark regression tracking — **none of which is configured**. The quoted cache block did not match the file either (three paths keyed `${CI_JOB_NAME}-…` against one path keyed `go-cache-v2-…`).

## Genericised rather than corrected

A correct job list needs re-verifying on every pipeline edit, and this one had never been verified at all — so the inventory is the thing that rots. **285 lines become 85.** The page now names `.gitlab-ci.yml` as the source of truth, tells the reader that `stages:` gives the order and each top-level key below it is a job, and documents only what the file cannot say for itself.

What survives was checked by execution rather than carried over: the four `make` targets and `scripts/run-tests.sh` it tells you to run all exist, and the `FERRET_TEST_MODE` correction is kept (the targets set it themselves, no production code reads it, it mocks nothing).

**The absences are now documented as absences**, in a table — a reader hunting for the coverage widget the old page promised had no way to learn it was never wired up. The commented-out Kaniko component is explained with its real reason (the placeholder path makes the whole file fail to parse, which would stop the security jobs running) and the resulting empty `docker` stage is stated. Security scanning is described as enabled *without* enumerating it, since which jobs GitLab's SAST, Secret-Detection and Dependency-Scanning templates expand to depends on the instance tier.

Both inbound links (`docs/README.md`, `docs/GITLAB_INTEGRATION.md`) keep working and their link text stays accurate, so no retargeting was needed.

## A guard covers the class

`cmd/documented_ci_jobs_test.go` treats a `#### \`name\`` heading in `docs/deployment/` as a claim that `name` is a job, and checks it against the file's top-level keys. Template-provided jobs are allowed only alongside the template that supplies them.

Because the rewritten tree contains **no such headings**, a passing walk would prove nothing — so the extractor is unit-tested against synthetic markdown in both directions: the inventory shape is caught, while a backticked filename in prose, a heading with prose around the token, a table row and an unbackticked heading are not.

## Two Go versions #576 does not reach

The README's install table cited **Go 1.26.6** as a from-source requirement while `.go-version` said 1.26.7; it now cites `.go-version` itself. `docs/webui-version-info.md` labels its `go1.26.2` sample as illustrative alongside the placeholder `commit` and `version` beside it, and states the field reports the build toolchain rather than a supported-version claim. Neither needs tracking on a bump.

```
=== scope ===
  CHANGELOG.md
  README.md
  cmd/documented_ci_jobs_test.go
  docs/deployment/GITLAB_CI_SETUP.md
  docs/webui-version-info.md

=== dimensions ===
2 full suite (-count=1)            RAN — 78 packages ok
10 non-vacuity                     NEEDS-MANUAL — see below
12 flake (3 repeats)               RAN — ./cmd
13 race (whole module)             RAN — 78 packages ok

```

### Dimension 10 — non-vacuity, discharged two ways

**A. Guard at HEAD, doc reverted to `origin/main`** — catches every fictional job:

```
--- FAIL: TestDeploymentDocsClaimOnlyRealCIJobs
    documented_ci_jobs_test.go:209: ../docs/deployment/GITLAB_CI_SETUP.md documents 13 job(s)
    that do not exist in ../.gitlab-ci.yml: benchmark code_quality coverage docker:build
    license_scanning pages release security:dependency-scanning security:sast
    security:secret-detection test:integration test:race test:unit
```

**B. Regex neutered** — the extractor test fails rather than passing vacuously:

```
--- FAIL: TestJobClaimExtractionWorks/the_#575_inventory_shape_is_a_claim
    documented_ci_jobs_test.go:187: extracted [], want [docker:build test:unit]
```

A third mutation (pointing the walk at a missing root) fails with `walked no markdown files … the deployment docs tree has moved`, so a moved tree cannot pass silently either.

### Conflict verification against the three open PRs

| vs | result |
|---|---|
| #573 `feat/314-svg-text-extraction` | merged clean (HEAD moved) |
| #574 `fix/459-verify-at-dispatch` | merged clean (HEAD moved) |
| #576 `chore/go-1.27.1` | merged clean (HEAD moved) |
| four-way union | **16 commits, builds, 82 packages ok, 0 FAIL** |

Only `CHANGELOG.md` is shared, and this entry sits in **📚 Documentation (L17)** against #573/#574 in 🐛 Bug Fixes (L41/L43) and #576 in 🔨 Internal (L134). The guard also passes on the union, i.e. with #576's `.gitlab-ci.yml` changes applied — it reads job names, not variable values, so there is no coupling.

## Filed from this work

**#577** — four `internal/router` tests call `os.MkdirTemp(".", …)`, creating scratch directories inside the repo tree and removing them mid-walk. That intermittently fails the sibling env-var guard under `go test ./...` (`walk ../internal: open ../internal/router/structured-sections-3058054452: no such file or directory`) on branches that touch neither package. The new guard here tolerates a transient `ENOENT` and documents why; the existing two do not.


sectoramen:
Added a fifth commit plus a full-tree documentation sweep, since the question came up of whether the rest of the docs need the same treatment.

## `docs(links)` — two links that were never valid

`README-Enhanced-Metadata.md` and `README-Preprocess-Only.md` both offered a "Testing Guide" at `docs/testing/TESTING.md`. `git log --diff-filter=D` shows **no deletion for that path**, so the links were broken from the day they were written rather than left behind by a rename. Retargeted to `docs/testing/TEST_PLAN.md`. `docs/` now has **zero** unresolvable relative links.

## Sweep result: the rest of the tree is in better shape than raw counts suggest

I measured all 109 markdown files for four drift classes. Two of the four dimensions turned out to be **measurement artifacts**, which is worth recording so nobody re-opens them on the raw numbers:

| Class | Raw | Verified real |
|---|---|---|
| Broken relative links | 2 | **2** — fixed here |
| Stale Go version pins | 6 in 5 files | **5** — 3 already in #576, 2 fixed here |
| "Missing" `make` targets | 29 in 18 files | **~0** — the regex matched English prose: `make sure`, `make the`, `make every`, `make informed` |
| "Missing" `.go`/`.sh` paths | 139 in 45 files | **~0** — bare filenames in prose (`` `registry.go` is the only coupling ``) that do exist; plus `CHANGELOG.md` and `docs/proposals/` legitimately naming deleted files |

So there is no large doc-correction batch waiting. The drift that actually matters in this repo is not paths or prose — it is **claims about behaviour**, and those are exactly what the guard series covers: flags (#503), env vars (#555), argument arrays (#564) and now CI jobs (#575). A path sweep structurally cannot see them, which is why the guards are the durable half of each of those fixes.

rustic (approved):

Issue #533, opened 2026-08-28
MEDICAL_ID: an OMOP CDM export reports no MRN, because its patient key is the generic name person_id


sectoramen:
## Summary

The OMOP Common Data Model names its patient key `person_id` and its source key
`person_source_value`. Neither is treated as medical context, so a real OMOP export reports **no
MRN** — even in a table whose sibling columns are unmistakably clinical.

This is the residual half of #436, which gave the MRN scan a column-header arm. That arm resolved
Synthea (`PATIENT`), Epic Clarity (`PAT_MRN_ID`), Cerner Millennium (`MRN`) and
`patient_identifier`. OMOP is the one published schema in that set it does not reach, and the reason
is deliberate rather than an oversight — which is why this is a separate question rather than a bug
in that change.

## Measured

At #436's HEAD, `--checks MEDICAL_ID --confidence all`, header rows taken from the OMOP CDM v5.4
`Field_Level.csv` data dictionary:

| fixture | findings |
|---|---:|
| `person_id,gender_concept_id,year_of_birth,person_source_value,care_site_id` / `5729183,8507,1980,4183926,9182736` | **0** |
| `visit_occurrence_id,person_id,visit_concept_id,visit_start_date,provider_id` / `5729183,4183926,9201,2024-01-01,8172635` | **0** |
| `PAT_ID,PAT_MRN_ID,PAT_NAME,BIRTH_DATE,SEX_C` / `Z1234,5729183,...` (Epic, for contrast) | 1, MRN 70 |

`person_id` appears in **18** of the 39 OMOP tables and is the join key of the whole model, so this
is the normal shape of an OMOP extract rather than an edge case.

## Why it is not simply a missing keyword

`hasMedicalContext` already states the position this rests on, and the position is sound:

> A bare "record number" is deliberately NOT here. It names an MRN only in a medical setting, and
> this list is what decides whether the line is a medical setting at all — admitting it would make
> every generic record identifier medical context.

`person` is exactly that. Adding it to the medical keyword list would make every `person_id` column
in every schema medical context, and it would apply on the **same-line** path too, not just in
tables — so `person_id: 5729183` in any config, log or ORM dump becomes an MRN. That is a much
larger blast radius than the gap it closes.

## Suggested direction

The signal that distinguishes an OMOP `person_id` from an arbitrary one is **table-level**, not
column-level: the sibling columns. An OMOP `person` table carries `gender_concept_id`,
[...]

PR #584, opened 2026-09-04
fix(socialmedia): a Make recipe prefix is not a Twitter handle, and a bare @token is not a profile URL

### LINKED ISSUE #582: socialmedia: a bare @token scores TWITTER at HIGH 100%, so a Makefile yields 346 false positives and the recipe prefix is unhandled
Measured on this repository with `social_media` enabled (it is **off by default**, which is the only reason this is not a shipped disaster).

## The numbers

**387 of 504 findings (77%) are `TWITTER`. 346 of those (89%) come from one file — the `Makefile`.**

| token | count | family |
|---|---|---|
| `@echo` | 270 | Make recipe prefix |
| `@if` | 21 | Make recipe prefix |
| `@go` | 21 | Make recipe prefix |
| `@rm` | 10 | Make recipe prefix |
| `@make` `@git` `@cd` `@chmod` `@mkdir` `@rc` `@goreleaser` `@scripts` `@GOOS` `@SCORE_REPORT` `@FERRET_DEBUG` `@CONTAINER_CMD` `@bash` `@cp` | ~23 | Make recipe prefix |
| `@media` `@page` `@layer` `@keyframes` `@import` `@charset` | ~15 | CSS at-rules **in prose** |
| `@yields` `@returns` `@see` `@token` `@linkcode` `@author` `@default` | ~8 | doc annotations |
| `@sha256` | 3 | Docker image digest |
| `@jack` `@schneems` `@xx` | 5 | **genuine handles — all quoted as examples in `CHANGELOG.md`** |

By file: `Makefile` 346, `CHANGELOG.md` 34, `config.yaml` 4, `THREAT_MODEL.md` 2, `Dockerfile` 1. Every one is a false positive.

## The Make recipe prefix is unhandled, and it is trivially separable

`handle_syntax.go` already vetoes CSS at-rules (four signals) and JSON-LD reserved keywords (two), and deliberately leaves the doc-annotation family alone. **Make is not covered at all**, and it is now the dominant family by an order of magnitude.

It is also the cleanest possible case. Measured on this `Makefile`:

```
@tokens anywhere in the file        : 357
@tokens NOT on a tab-indented line  : 0
```

**Every single one is on a recipe line**, which GNU Make requires to begin with a TAB. So a two-signal rule in the existing style — the token follows a leading TAB (optionally after `-`/`+`/`@` prefix characters), and the file is a makefile (`Makefile`, `makefile`, `GNUmakefile`, `*.mk`) — removes 346 of 387 findings with no collateral, and both signals are things a handle cannot supply. A handle at column 0 of a tab-indented line inside a file named `Makefile` is not a shape real mention data takes.

`@sha256` is a second cheap one and is categorical rather than statistical: a digest is `@sha256:` followed by 64 hex characters, and a handle cannot be followed by `:`.

## The structural defect: a bare 4-character token scores identically to a full profile URL

`@echo` and `https://twitter.com/awscloud` both report **TWITTER at 100% confidence, HIGH**. That is the real problem, and no amount of vocabulary vetoes fixes it — the next unhandled file format brings its own `@` syntax.

The two forms carry very different intrinsic evidence:

- `twitter.com/<user>` names the platform in the value itself. Nothing else produces that string.
- `@token` is four-plus characters of anything, and is valid syntax in Make, CSS, JSDoc, JSON-LD, email, Docker digests, IRC, and shell.

The bare form should require corroboration to reach HIGH — a platform word nearby (`twitter`, `follow`, `handle`, `mention`), handle density on the line, or the mention-export shape — and should otherwise report at a lower band. It must keep reporting: suppressing it entirely is a cleartext leak, see below.

## Do NOT solve this by requiring a full URL

The obvious fix is to drop the bare pattern, and it is a **one-line config change** rather than code — `twitter_patterns` ships two regexes, the URL and `@[a-zA-Z0-9_]{1,15}\b`. Measured both ways:

| fixture | both patterns | URL only |
|---|---|---|
| this repository | 398 | **0** |
| one-handle-per-line mentions export | 3 | **0 — lost** |
| handle-keyed JSON moderator roster | 3 | **0 — lost** |
| `Contact @awscloud on Twitter for support` | 1 | **0 — lost** |
| `Follow us at https://twitter.com/awscloud` | 1 | 1 |

URL-only eliminates 100% of the false positives and 100% of bare-handle recall. The three lost shapes contain **no URL at all**, and the first two are precisely the shapes `handle_syntax.go` documents as confirmed leak risks from an earlier over-broad suppression. A handle in prose beside the word "Twitter" is a realistic PII artifact and is also lost.

So the recall cost is real and concentrated exactly on the artifacts most likely to matter. The fix is to make the bare form *earn* its confidence, not to delete it.

## Suggested order

1. Make recipe-prefix rule (two signals, removes 89% of this scan).
2. `@sha256:` digest rule (categorical).
3. Confidence: the bare form does not reach HIGH without corroboration; the URL form keeps its current score.
4. Document the URL-only config variant with this measured trade-off, so a user who prefers precision over recall can choose it knowingly.

Not proposed: extending `docPatterns` with more tag words. That file explains why — it has no position gate, so each word suppresses the handle in every file type, and a recall hunt already showed `@yields` is a real GitHub account credited in a real changelog.

sectoramen:
Closes #582.

## The problem

`@echo` and `https://twitter.com/awscloud` both reported **TWITTER at 100%, HIGH**.

Measured on real files rather than fixtures. **Ten real Makefiles** — nine third-party from `/Library`, `/Applications` and unrelated projects, plus this repository's — carry 797 tab-indented `@` recipe lines and produced **736 findings, 511 of them HIGH**, every one false. On this repository, 387 of 504 findings were TWITTER and 346 came from the `Makefile`, `@echo` alone accounting for 270.

## Two mechanisms, split by how certain the evidence is

That split is the design, and the first draft of this PR got it wrong by using only the second half.

### Veto where syntax decides

A **Make recipe prefix** is a *position* in Make's grammar: a recipe line must begin with a TAB, and the first token after it (past any `-`, `+`, `@`) is the command Make executes. A handle cannot occupy that position in a working Makefile, because Make would try to run it. An **image digest** is followed by `:` plus hex, and the configured handle pattern `@[a-zA-Z0-9_]{1,15}` cannot contain a colon.

This is why these are safe where this file's existing CSS rule needed four signals and two cleartext leaks to get right: an at-rule **name** can also be somebody's account (`@media` is a plausible handle), so a name-based rule is inherently probabilistic. A **position** is not.

Verified in the direction that matters:

```
\t@echo "Follow @awscloud on Twitter"   ->  vetoes @echo, still reports @awscloud at HIGH
  @awscloud   (space-indented, in a file named Makefile)  ->  reported normally
```

The tab check is **load-bearing, not decoration** — removing it makes that second case report **nothing**, proven by mutation.

### Demote where it does not

JSDoc (`@returns`, `@author`), CSS at-rules quoted in prose, JSON-LD, IRC and every future `@`-format get a `confidence_ceiling` of 55, inside the LOW band, so `--confidence high` and `--confidence high,medium` are both clean while **nothing is ever suppressed**.

A veto per format would be a vocabulary that rots — the exact failure mode that sank the first attempt at #533 days earlier, where an entry naming a table rather than a column and a term that turned out to be ordinary English both read like working guards.

A **ceiling** rather than a penalty because the existing `-30` advisory penalty is a **no-op**: these score 100 and the raw score runs higher, so the cap absorbs it entirely. A ceiling is applied after every adjustment that can raise confidence, including the `+20` context boost and cross-path correlation.

**Demotion cannot leak, and that is measured rather than assumed:** with `--confidence high`, a demoted handle in a roster file is **still masked** in the redacted output.

### Corroboration is document-scoped

`detector.ContextInfo` here carries only the match's own line plus a 50-byte window inside it, so a heading cannot vouch for the handles beneath it — and a one-handle-per-line mentions export is precisely the artifact `handle_syntax.go` records two leaks defending. Safe to widen because the list only ever **promotes**, and because this repository's `Makefile` contains **zero** corroboration words, so its 346 findings are untouched by the widening.

## The corroboration vocabulary was measured, and its first version was wrong

It held the bare words `handle`, `username`, `follow`, `following`, matched as substrings. On **400 real JavaScript files**:

| word | files |
|---|---|
| `handle` | **66 (16.5%)** — `handleClick`, `handleRequest`, `file handle` |
| `follow` | 27 |
| `following` | 17 — "the following steps" |
| `username` | 15 |
| `twitter` | **0** |

One real file in five lifted the ceiling on a word with nothing to do with social media; `function handleClick()` alone promoted a JSDoc `@returns` out of the LOW band. The list now holds **platform names and multi-word phrases only** — `follow us` where `follow` was, `twitter handle` and `handle is` where `handle` was.

**1 of those 400 files now corroborates, with no recall lost:** `Our handle is @awscloud`, `follow us: @awscloud` and a `Twitter mentions export` all still report HIGH.

## Results

| | total | HIGH | MEDIUM | LOW |
|---|---|---|---|---|
| main, this repo | 399 | **292** | 106 | 1 |
| this branch | 52 | **45** | 1 | 6 |
| main, 261 real JS/CSS/Dockerfile files | 1,707 | **1,329** | 377 | 1 |
| this branch | 1,706 | **117** | 32 | 1,557 |

The 261 files are those that actually contain a bare `@token` — 261 of 600 candidates, so the corpus exercises the change rather than proving a precondition absent. **Distinct values reported: 75 before, 75 after.** Nothing lost, only demoted.

Ten real Makefiles: **736 findings (511 HIGH) → 0.** The suppressed set is exactly 18 Make and shell command names — `@echo @go @if @rm @cd @cp @chmod @mkdir @git @make @bash @goreleaser @scripts @rc @GOOS @CONTAINER_CMD @SCORE_REPORT @FERRET_DEBUG` — not one a plausible account name, while `@jack` and `@schneems` are still reported and zero values were gained.

`SOCIAL_MEDIA` remains **off by default**.

```
production (non-test) .go files: 3

1 gofmt                   RAN — clean
1 go vet                  RAN — clean
1 go build                RAN — ok
2 full suite (-count=1)   RAN — 81 packages ok
3 golden regen            RAN — 0 files changed
4 redaction               NEEDS-MANUAL — discharged below
5 suppression             NEEDS-MANUAL — discharged below
6 timing / O(n^2)         NEEDS-MANUAL — discharged below
7 all 7 formats           NEEDS-MANUAL — discharged below
8 web / CSP               N/A — no web/template/asset file touched
10 non-vacuity            NEEDS-MANUAL — discharged below
11 cross-platform vet     RAN — linux darwin windows
12 flake (3 repeats)      RAN — ./internal/validators/socialmedia
13 race (whole module)    RAN — 81 packages ok
-- -short mode            RAN — ok
-- tree clean             RAN — this run dirtied nothing

```

**4 — redaction.** All three strategies write a file with **zero cleartext** (`mask`, `synthetic`, `remove`), and the same holds under `--confidence high`, which is the check that proves demotion cannot leak.

**5 — suppression.** Rules generated by the **parent** binary, then applied to mine: parent with no rules 4 findings, parent + rules **0**, mine + rules **0**. Non-vacuous — the first attempt read 4→4 because generated rules ship `enabled: false`.

**6 — timing.** A/B against the parent at N/2N/4N with distinct values and a non-vacuity floor: **1.04x / 1.01x / 1.00x** at 4,000 / 8,000 / 16,000 findings. Separately measured to 1.5MB across the 1MB batch-path threshold: time and RSS both within noise, so the document-level scan does not show.

**7 — all 7 formats.** text 4, json 12, yaml 12, csv 4, junit 1, gitlab-sast 20, sarif 20 — all rc=0.

**10 — non-vacuity.** Eight mutations, every one now failing its test. **Three caught defects in the tests before they were trusted:**

- `TestAProfileURLIsNeverCapped` was **vacuous** — a `twitter.com` URL contains the corroboration word "twitter", so the check returned early and never reached the guard under test; deleting that guard left the test green. It now uses LinkedIn, GitHub and TikTok URLs and asserts at runtime that no fixture self-corroborates.
- One mutant **did not compile** and read as inconclusive until the import was kept live.
- The ordinary-code corroboration test fails if a bare generic word is re-added.

## Known limits, stated rather than discovered later

- `socialCorroboration` is still a vocabulary, now 21 phrases. A new platform will not be in it, so handles on pages about it get demoted — decay in the **safe** direction (still reported, still redacted), but real. Deriving it from the configured `positive_keywords` would make it user-extensible; that is a design change left out of this PR.
- `isMakefileName` enumerates filenames and misses `Makefile.local`, `BSDmakefile`, `*.mak`.
- `bareHandleCeiling = 55` is justified from the 90/60 band boundaries but not measured against alternatives.
- Only two formats are vetoed. Each new `@`-format adds LOW noise; the severity is bounded by the ceiling, the volume is not.


sectoramen:
Note from #589 (email TLD work), which hit the same edge: `confidenceCeilingKey` there had to be declared **locally** rather than imported from `internal/validators`, because a test in that package imports the validator packages and importing back is a cycle:

```
imports .../validators/email from explain_rationale_test.go,
imports .../validators from tld_ceiling.go: import cycle not allowed in test
```

`secrets`, `ipaddress` and `ssn` all declare a local copy for this reason, and `dual_path_bridge.go`'s own comment states the intent — *"The key is a plain string rather than an import from one validator package so any validator can use the mechanism without the bridge depending on it."*

**This PR imports `internal/validators` from `bare_handle_ceiling.go`.** There is no cycle today, because nothing in `internal/validators`'s tests imports socialmedia — but the edge is one test import away, and it would surface as a confusing "cycle not allowed in test" on an unrelated change. Worth switching to a local const for consistency with the other four validators. Happy to push that if you would rather it went in before merge.

rustic (approved):

Issue #421, opened 2026-08-20
rtf: a value split across formatting runs is silently missed, because RTF markup reaches the validators unextracted


sectoramen:
## Summary

`.rtf` is not scanned. A file containing an SSN is reported as an unsupported type at exit 0, while
the identical bytes in a `.txt` report `SSN HIGH 100`. RTF is a common export format from Word,
Outlook and many report generators, so it plausibly carries real PII in real workflows.

This is a **feature request, not a defect**: nothing in the code or the docs claims `.rtf` support,
so today's behaviour is an honest statement of capability. Filing it so the gap is tracked rather
than rediscovered.

## Measured at `main @ c7ae0a9`

```
$ printf 'Employee SSN: 452-11-9384\n' > ctl.txt
  ctl.txt,SSN,HIGH,100.0,1,[HIDDEN]

$ ferret-scan --file x.rtf --enable-preprocessors --fail-on-incomplete   # same SSN inside RTF
  exit=0   files_processed=0  files_skipped=1  files_not_examined=ABSENT
```

Also checked `.ps`/`.eps`: same behaviour. See "Why RTF and not PostScript" below.

## What it would take

Registration surface, measured on this tree: an existing file type appears in **5–9 non-test Go
files** (`shared_utilities.go`, `text_preprocessor.go`, `plaintext_preprocessor.go`, the extractor
package, `embedded.go`, plus `redactors/manager.go` if it should be redactable) and **23 docs
enumerate supported types**. That plumbing is most of the work and it is mechanical.

The extractor itself is small and needs **no new dependency** — RTF is ASCII markup:

- strip control words (`\word`, `\word123`)
- decode `\'hh` hex escapes and `\uN` Unicode escapes
- map `\par` / `\line` to newlines
- skip ignorable destinations `{\*\...}` — font tables, stylesheets, revision data

Roughly 150-250 lines plus tests. Estimate ~half a day to a day including docs and golden cases.

## Two hazards worth writing down before anyone starts

**1. Do not route it through the printability sniff.** RTF begins `{\rtf1` — ASCII, like PDF's
`%PDF-`. Adding `.rtf` to `containerExtensions` would inherit #419 exactly: the sniff would say
"text", the plaintext preprocessor would claim it, and raw RTF markup would reach every validator.
Key it on magic bytes and send it straight to a real extractor; PR #420's `pdfMagic` check is the
pattern to extend.

[...]

PR #585, opened 2026-09-04
feat(coverage): name the skipped file types, because a count is not a disclosure

### LINKED ISSUE #583: coverage: files skipped as an unsupported type are counted twice but never named, so "14 skipped" cannot be distinguished from 14 documents that were missed
Scanning this repository reports:

```
Starting scan of 43 files...
Filtered out 14 unsupported file types
Scan complete: 29 files scanned, 14 skipped (unsupported type) in 456ms
```

Nothing anywhere says **which** 14, and no flag will tell you. `--verbose` adds only the same count. The `NotExaminedFooter` never fires for this cause.

## Nothing regressed — this was a deliberate decision

`docs/COVERAGE_DISCLOSURE.md` documents a closed list of eight `NotExaminedCause` values, and an unsupported type is deliberately **not** one of them:

> An **unprocessable** type refused for size is not listed at all. It is a genuine skip: no finding was ever possible from it, so reporting lost coverage would be reporting a non-event.

That reasoning is sound *as far as it goes*, and in this specific scan the outcome is correct. The 14 are:

- **13 stray `*.test` files** — compiled Go test binaries in the repo root, gitignored, untracked, dated June/July
- **`.DS_Store`**

No finding was possible from any of them, so no coverage was lost. The decision was right.

## But the disclosure is not usable, and the premise is not self-evidently true

Two problems with resting on "no finding was ever possible".

**1. The user cannot verify it.** "14 skipped (unsupported type)" is indistinguishable from 14 PDFs, 14 spreadsheets or 14 archives that the tool declined. The operator has to take the classification on trust with no way to audit it — for a DLP tool, on the one axis where a mistake is silent. Here it happened to be build detritus; the output would be byte-identical if it were fourteen customer documents.

**2. The supported-type list has been wrong before**, so "unsupported" is an assertion rather than a fact:

- **#421** (open) — `.rtf` is not scanned.
- **`.rtf`/`.tmp`/`.pages` were listed as supported by the finder and then reported CLEAN by the scanner** — a gate mismatch where the two halves disagreed.
- Registering `.3gp` reached the preprocessor and still reported 0 findings, because three separate places have to agree before a type is really handled.

Each of those was a file the tool *believed* it could not usefully process while a user believed otherwise. A count with no names makes that class of defect invisible: the failure mode is exactly "the tool quietly declined something you cared about", and the current output cannot distinguish it from the healthy case.

## What would fix it, cheaply

Not necessarily a ninth `NotExaminedCause` — the doc argues against that and the argument holds, since an unsupported type genuinely is not lost coverage and should not appear in a footer that means "we may have missed something here".

A separate, opt-in listing is enough:

- `--show-skipped` (or `--verbose` gaining it) printing the skipped paths grouped by extension: `13 × .test, 1 × .DS_Store`. Aggregate by extension rather than one line per file, so a large tree stays readable.
- Include the same grouping in the machine formats where a footer already exists, as a field distinct from the not-examined list so consumers cannot confuse "declined, nothing was possible" with "may have missed something".

The counter itself is also reported twice with different wording — `Filtered out 14 unsupported file types` on stderr before the scan and `14 skipped (unsupported type)` after. Those should agree or one should go; `main.go` already carries a comment about an earlier version of this counter describing a supported type as an unsupported one.

## Verification

A fix should show the extension grouping for this repository (`13 × .test, 1 × .DS_Store`), and a test that a directory containing one genuinely unsupported type and one *supported-but-unreadable* file lists them **separately** — the second belongs in the not-examined footer, the first does not, and conflating them is what the eight-cause taxonomy exists to prevent.

sectoramen:
Closes #583.

## The problem

Scanning this repository reported:

```
Filtered out 14 unsupported file types
Files: 29 scanned, 14 skipped | Findings: 203 (20 high, 126 medium, 57 low)
```

Nothing anywhere said **which** 14, and no flag would tell you — `--verbose` adds only the same count, and the `NotExaminedFooter` deliberately does not cover this cause.

**A count is not a disclosure.** That output is byte-identical whether the skipped files are build detritus or fourteen customer documents. Here they were 13 compiled Go test binaries and a `.DS_Store`, so no coverage was lost — but the operator had no way to establish that, on the one axis where a mistake is silent.

It matters beyond tidiness because the supported-type list **has been wrong before**: #421 is open for `.rtf`; `.rtf`, `.tmp` and `.pages` were once advertised by the finder and then reported CLEAN by the scanner; and registering `.3gp` reached the preprocessor and still produced nothing, because three separate places have to agree before a type is really handled. Each was a file the tool *believed* it could not usefully process while a user believed otherwise — invisible behind a bare count.

## After

```
Files: 29 scanned, 14 skipped (.test × 13, .DS_Store × 1) | Findings: 203 (20 high, 126 medium, 57 low)
```

and `json`/`yaml` carry the same data structurally:

```json
"files_skipped": 14,
"skipped_types": { ".test": 13, ".DS_Store": 1 }
```

`skipped_types` is deliberately **distinct from the not-examined list**, so a consumer cannot confuse "declined, nothing was possible" with "may have missed something".

## This does not add a ninth NotExaminedCause, and that argument still stands

`docs/COVERAGE_DISCLOSURE.md` argues an unprocessable type is a genuine skip, so listing it among causes meaning *"we may have missed something here"* would report a non-event. **That reasoning is correct and unchanged.** Naming the files is a different question from where they are reported — and the doc is updated to say where they *are* named rather than left asserting they are "not listed at all".

## Grouped by extension, because the reason does not discriminate

Measured with a probe calling `CanProcessFile` directly: **all 14 skips return the single reason `Unsupported file type`**, so a reason histogram has one bucket and says nothing. Extension is the axis that discriminates, and it is what an operator recognises — `.test × 13` is immediately identifiable as build output.

Worth stating precisely: the histogram describes the **files**, not the rule. The skip decision is content-aware, not extension-based — a one-byte file named `f.test` **is** scanned while a 20MB Mach-O of the same name is not. That is the right way round for a disclosure: it reports what was declined, not how the router decided.

## Details that are load-bearing

- **Both stderr messages now use the same wording and the same histogram, over the union of both skip stages.** They previously described the same files with different words, which is how a reader concludes they are different files. Counting one stage and naming the other would understate.
- **The human histogram caps at 6 types and discloses the remainder** as `and N more types`. A silent truncation that reads as completeness is the exact defect this change fixes, so the overflow is not optional. The **structured value is never capped** — capping it would silently lose data for anyone parsing json.
- **Ordering is descending count then label**, pinned by a test that renders 50 times. Ranging the map would vary run to run in a line that goes into diffed CI logs and whose width sizes the summary box.
- **Dotfiles are labelled as themselves.** `filepath.Ext(".DS_Store")` returns the *whole name*, which happens to read well but would render `.gitignore` as an extension. Files with no extension get their own `(no extension)` bucket rather than silently sharing one.

## One correction to the issue

The issue title says these files are "counted **twice**". They are not — on a 36-file probe, 1 scanned + 35 skipped = 36, so the arithmetic is right. They were **reported** twice in two different wordings, which is what the shared wording fixes. The title overstates it.

```
1 gofmt                   RAN — clean
1 go vet                  RAN — clean
1 go build                RAN — ok
2 full suite (-count=1)   RAN — 81 packages ok
3 golden regen            RAN — 0 files changed
4 redaction               NEEDS-MANUAL — discharged below
5 suppression             NEEDS-MANUAL — discharged below
6 timing / O(n^2)         NEEDS-MANUAL — discharged below
7 all 7 formats           NEEDS-MANUAL — discharged below
8 web / CSP               N/A — no web/template/asset file touched
10 non-vacuity            NEEDS-MANUAL — discharged below
11 cross-platform vet     RAN — linux darwin windows
12 flake (3 repeats)      RAN — ./cmd ./internal/formatters ./internal/formatters/text
13 race (whole module)    RAN — 81 packages ok
-- -short mode            RAN — ok
-- tree clean             RAN — this run dirtied nothing

```

**4 — redaction.** All three strategies write a file with zero cleartext on a fixture holding an SSN and a Visa alongside a skipped binary and a `.DS_Store`.

**5 — suppression.** Rules generated by the **parent** binary, applied to mine: parent no rules 2 findings, parent + rules **0**, mine + rules **0**. Non-vacuous (generated rules ship `enabled: false`, so they must be enabled or the check reads 2→2).

**6 — timing.** A/B against the parent on trees of 500 / 1,000 / 2,000 skipped files across **200 distinct extensions**, which is the work this change adds: **1.00x / 0.92x / 1.00x**, with the type count asserted non-zero at every size.

**7 — all 7 formats.** All rc=0; `skipped_types` present in json and yaml.

**10 — non-vacuity.** Four mutations, each failing a test: dropping the deterministic sort, silently dropping the overflow, capping the structured value, and treating a dotfile as an extension.

## Deliberately not in scope

The issue also floated a `--show-skipped` flag listing individual paths. Not added: the extension histogram answers the question that was actually being asked, fits on the existing line, needs no new flag surface or documentation, and stays readable on a large tree. A per-path listing can be added later if grouping proves insufficient.


sectoramen:
Moved the CHANGELOG entry from 🐛 Bug Fixes to ✨ New Features in a follow-up commit.

Both this PR and #584 had inserted at the top of the Bug Fixes section, and they **conflicted on merge** — the CHANGELOG anchor collision this repo has hit before. New Features is unused by #584 and is the better fit regardless, since this adds output rather than changing detection behaviour.

Re-verified after the move: **merges clean with #584, union of 11 commits builds, 82 packages ok, 0 failures.**

rustic (approved):

Issue #596, opened 2026-09-04
perfguard: MinMeasurableCPU is 2ms while the constant's own doc says Windows granularity is 15.6ms, so a quantised reading can clear the gate


sectoramen:
Found while fixing the windows-latest break in `TestATinyWorkloadFallsBackToWallClock` (run [33899001272](https://github.com/awslabs/ferret-scan/actions/runs/33899001272), and again on the next commit). Filing separately because settling it needs a measurement from a Windows runner, not a local one.

## The inconsistency

`internal/perfguard/perfguard.go`:

```go
// getrusage reports microseconds and GetProcessTimes 100ns ticks, but both accumulate at timer-tick
// granularity — 15.6ms on Windows by default. Dividing two readings near that granularity produces a
// quantised ratio, which is the noise regime this package exists to escape. Below it, Growth falls
// back to wall clock and SAYS SO, ...
const MinMeasurableCPU = 2 * time.Millisecond
```

The comment names **15.6ms** as the Windows granularity; the gate is **2ms**. If the comment is right, a Windows CPU reading between 2ms and 15.6ms is a **single quantised tick** that *clears* the gate and gets divided — which is precisely the noise regime the package exists to escape. The fallback would not fire, because `b.CPU < MinMeasurableCPU` is false.

Nobody had measured the real granularity on any platform, which is the actual problem: the threshold rests on an unverified number.

## What is now measured

`TestClockGranularityIsReportedAndCPUTimeIsMonotonic` polls each clock in a tight loop and logs the granularity plus how many ticks the gate is worth, on every CI run on all three platforms. Measured so far:

| platform | cpu granularity | wall granularity | gate in ticks |
|---|---|---|---|
| darwin/arm64 (M4 Pro) | **1µs** | 41ns | **2,000** |
| ubuntu-latest | *pending a CI run* | | |
| windows-latest | *pending a CI run* | | |

macOS is comfortable. **Read the windows-latest log line from any run on `main` to settle this.** The test warns when the gate is worth fewer than ten ticks.

## Why this has not visibly broken anything

Every current caller measures far above the threshold, so the quantised band is not being entered in practice:

- `internal/goldencorpus/complexity_guard_test.go` — 18 validator targets, base readings ~20ms and up (locally: `passport` base 19.75ms). The package **passed on windows-latest** in the same run where `internal/perfguard` failed, taking 159s.
- the two SVG guards in `text-extract-svgtextlib` — base readings ~20ms and ~150ms.

[...]

PR #597, opened 2026-09-04
fix(tests): stop the coarse-clock test failing on Windows, and measure each platform's real clock granularity

sectoramen:
Fixes the `main` breakage in [run 33899001272](https://github.com/awslabs/ferret-scan/actions/runs/33899001272) and [run 33899228640](https://github.com/awslabs/ferret-scan/actions/runs/33899228640).

## What broke

`TestATinyWorkloadFallsBackToWallClock`, added by me in #590, failed on **windows-latest** — deterministically, in two consecutive runs on `main`:

```
perfguard_test.go:182: a tiny but non-empty workload should still measure on the wall clock:
                       perfguard: every base reading was zero on the wall clock
```

The test required a "tiny but non-empty" workload to be measurable on the wall clock. On Windows a ~90µs workload reads **zero on BOTH clocks**.

**The premise was unsatisfiable there, not merely unlucky.** The window it aimed at — long enough to measure on the wall clock, short enough to sit under `MinMeasurableCPU` (2ms) — can be **empty** when the wall granularity is not finer than that threshold. No choice of iteration count would have fixed it.

The irony is worth recording: this test was changed *from* an empty-function version specifically to close a vacuity hole, and that change is what introduced the platform dependency. Removing one hole opened another.

## The fix

What the test is actually about is the **fallback decision**, and that decision is made before either outcome is known. So it now asserts on the **clock named in the result**, and accepts both correct behaviours:

| outcome | correct? |
|---|---|
| measurable on wall → `Growth.Clock == "wall"` | ✅ |
| measurable on neither → error **naming** the wall clock | ✅ |
| reported or errored on the **cpu** clock | ❌ — a quantised reading was about to be divided |

Renamed to `TestATinyWorkloadDoesNotDivideAQuantisedCPUReading`, since that is the actual invariant.

**Verified three ways rather than assumed:**
1. Passes on the **simulated Windows state** (both clocks forced to zero).
2. Still **CATCHES** the mutation that forces the CPU clock (`b.CPU < MinMeasurableCPU` → `false`).
3. Catches that mutation **even on a coarse clock** — both conditions at once.

Without (2) and (3) this would have been a fix that traded a broken test for a vacuous one, which is the mistake that produced the original hole.

## A latent inconsistency found while diagnosing, now measured on every run

`MinMeasurableCPU` is **2ms**, while the constant's own doc comment says Windows accumulates at **"15.6ms by default"**. If that figure is right, a Windows reading between the two **clears the gate while still being a single quantised tick**, and gets divided anyway. Nobody had measured the real granularity on any platform — the threshold rested on an unverified number.

`TestClockGranularityIsReportedAndCPUTimeIsMonotonic` now polls each clock in a tight loop and reports the granularity plus how many ticks the gate is worth, on all three platforms, every run:

```
clock granularity on darwin/arm64: cpu=1µs wall=41ns
MinMeasurableCPU (2ms) is 2000 cpu ticks on this platform
monotonicity: 22115 readings over 60.001ms of CPU, never decreasing
```

It warns when the gate is worth fewer than ten ticks. **The windows-latest and ubuntu-latest figures will arrive with this PR's own CI run**, which is what settles #596 — from a log rather than a guess.

It also pins the one invariant `ProcessCPUTime` must have, **monotonicity**, because a ratio of minimums would *preferentially* select a corrupted sample precisely because a corrupted sample is the smallest.

**Two of my own measurement errors, caught and corrected before commit:**
- The granularity test first measured the smallest **workload** rather than the **clock**, reporting `866µs` and then `1.912ms` on the same machine depending on the fixtures. Replaced with a tight poll of the clock itself.
- The monotonicity check was too short to matter: a mutant clock wrapping every 3ms **survived** it, because the loop exited after one ~1µs tick. It now spans 60ms of CPU and asserts the span, and the mutant is caught with `ProcessCPUTime went BACKWARDS after 214µs of CPU: 2.986ms then 1µs`.

## The ubuntu failure in the same run is NOT this, and needs no fix

`ubuntu-latest` failed with `exit code 143` and *"The runner has received a shutdown signal"* — the runner was cancelled. Ubuntu **passed** on the very next commit. The `/usr/bin/tar: ... Cannot open: File exists` lines are an `actions/cache` restore `##[warning]`, not the failure cause. No code change is warranted, and I have not made one.

## Test plan

Test-only change; no production file is touched (`CHANGELOG.md` + `internal/perfguard/perfguard_test.go`).

- [x] **gofmt / vet / build**: clean.
- [x] **full suite** (`-count=1`): all packages green.
- [x] **cross-platform vet**: linux / darwin / windows clean.
- [x] **windows test binary cross-compiles**: `GOOS=windows go test -c ./internal/perfguard/` — the platform that broke is the one that must build.
- [x] **flake (3 repeats)**: `./internal/perfguard` 3/3.
- [x] **non-vacuity, mutation-verified**: forced-CPU mutation CAUGHT (and caught again under a simulated coarse clock); backwards-clock mutation CAUGHT by the new monotonicity span. The Windows condition itself was simulated and passes.
- [x] **golden regen**: not applicable — no detection, scoring, formatting or redaction path is touched. `UPDATE_GOLDEN=1` was run and changed 0 files.
- [x] **redaction / suppression / formats / web-CSP**: N/A by diff scope — no validator, redactor, formatter, CLI or asset file is in the diff.

**Note on the checklist:** `make pr-checklist BASE=origin/main` was running against this worktree while a parallel audit wrote scratch `_test.go` files into it, so that run's results are contaminated and I am not quoting them. Every dimension above was instead verified directly and is reproducible from the commands named. I will re-run the checklist clean and post the table as a comment.

Closes nothing on its own; **#596** tracks the `MinMeasurableCPU` question this uncovered, and now carries a measured probe showing a cheap target silently reverts to the wall clock (and would hard-fail on Windows).


sectoramen:
## Correction: the checklist run was **not** contaminated — here is the table

The PR body says a parallel audit wrote scratch `_test.go` files into the worktree while `make pr-checklist` was running, so that run's results were contaminated and I would not quote them. **That was over-cautious and, as it turns out, wrong.** I re-ran the checklist in a *fresh, isolated* worktree checked out from the pushed branch, which the audit agents had no knowledge of, and the output is **byte-identical** to the run I discarded — same scope, same dimensions, same counts, and `tree clean — this run dirtied nothing` in both.

The scratch files were transient and in a different package (`internal/goldencorpus`); they never entered the diff, the scope, or the tree-clean check. So there is no reason to withhold the table:

```
=== scope ===
  CHANGELOG.md
  internal/perfguard/perfguard_test.go

=== dimensions ===
2 full suite (-count=1)            RAN — 84 packages ok
10 non-vacuity                     NEEDS-MANUAL
12 flake (3 repeats)               RAN — ./internal/perfguard
13 race (whole module)             RAN — 84 packages ok

```

**Dimension 10 (non-vacuity) discharged**, as described in the PR body — this is the one the harness cannot check for itself:

| mutation | result |
|---|---|
| force the CPU clock (`b.CPU < MinMeasurableCPU` → `false`) | **CAUGHT** — `Clock = "cpu" for a workload below MinMeasurableCPU (2ms), want wall` |
| force the CPU clock **and** simulate a coarse clock (both readings zero) | **CAUGHT** — so the tolerance added for Windows did not make the test vacuous |
| clock wrapping every 3ms (`(user+sys) % 3ms`) | **CAUGHT** by the new monotonicity span — `ProcessCPUTime went BACKWARDS after 214µs of CPU: 2.986ms then 1µs` |
| simulate the Windows state alone (both clocks zero, no mutation) | **PASSES**, which is the fix working |

Two mutants that did **not** compile, and one that survived for a bad reason, were re-crafted rather than counted: halving the clock (`(user+sys)/2`) preserves monotonicity so it proved nothing, and the short version of the monotonicity check exited after a single ~1µs tick and never saw a wrap. That is why the span is now asserted (`≥100 readings over ≥10ms of CPU`) rather than left implicit.

Apologies for the noise in the body — the note there should be read as superseded by this comment.

sectoramen:
Superseded by #602, which carries this content (cherry-picked with `-x` provenance) plus the **root-cause fix** underneath it.

Rolled into one PR at the maintainer's request. The reason it is one and not three: the windows-latest failure here turned out to be a symptom of `MinMeasurableCPU` being a fixed 2 ms while Windows' clock tick is **15.625 ms** — so the gate was 0.128 of a single tick and the guard was dividing integer tick counts (a linear control reporting "2.00x" from 2 ticks over 1). Fixing that in #602 is also what makes the ubuntu fixture fix safe, which is why splitting them was the wrong shape.

Nothing is dropped: #602's checklist is 10 ran / 0 failed with race green at 84 packages, and every mutation from this PR is re-verified there.

PR #598, opened 2026-09-04
fix(tests): guard the line-memo growth ratio's divisor, so a zero reading cannot report a false O(n²)

sectoramen:
Found while diagnosing the windows-latest break in #597, by auditing the repo for the same class of assumption. Reproduced independently before being written up.

## The defect

`internal/detector/line_span_complexity_test.go` computed

```go
ratio := float64(tBig) / float64(tBase)
...
if ratio > 2.5 {   // <- no tBase > 0
```

while the **two other ratio assertions in the same file already carry that check**:

```go
line 495:  if fBase > 0 && fRatio > maxCountGrowth {
line 649:  if tBase > 0 && ratio > maxGrowth {
```

So this is an inconsistency inside one file, not a house style.

It matters because this is the **smallest timed workload in the repo** — measured on darwin/arm64, base **358µs** plain and **1.1ms** under `-race`. A wall-clock reading that small can come back as exactly **zero** where the clock advances at timer-tick granularity, and that is not hypothetical: windows-latest returned a zero *wall* reading for a ~90µs workload in `internal/perfguard`, which is what broke [run 33899001272](https://github.com/awslabs/ferret-scan/actions/runs/33899001272).

**Both arithmetic outcomes are wrong, and the louder one is the false alarm:**

| readings | ratio | `> 2.5` | result |
|---|---|---|---|
| `tBase == 0`, `tBig > 0` | `+Inf` | **true** | reports *"the per-match line hash is back"* on **correct code** |
| `tBase == 0`, `tBig == 0` | `NaN` | false | **passes having measured nothing** |

Verified in Go rather than reasoned about: `15625.0/0 == +Inf` and `+Inf > 2.5 == true`; `0.0/0.0 == NaN` and `NaN > 2.5 == false`.

## Verified two ways

**1. The false failure reproduced against the real test.** Forcing `tBase = 0`:

```
with the guard:     PASS
without the guard:  FAIL — cost +Infx more (base=0s big=276.084µs)
```

**2. The guard still catches what it exists for** — this is the check that mattered most, because adding a divisor check must not quietly switch an assertion off. Destroying the line-id memo (making the fast path unreachable) still fails:

```
cost 6.28x more (base=5.131209ms big=32.199417ms)
```

Note the base there is 5.1ms, well clear of zero, so the new guard does not suppress a genuine regression.

## What I deliberately did NOT do

**I did not also fail when `tBase == 0`**, although that would close the silent-vacuity half. That would trade a false pass for a false *failure* on any platform whose granularity exceeds this fixture — and the real granularity on windows-latest is **still unmeasured**. Assuming it is precisely the mistake that produced the break this hardens against.

#596 now collects that figure on every CI run (a step added in #597). Once it is known, either this fixture grows until it clears the clock, or the pair moves onto `internal/perfguard.Measure`, which refuses a zero reading outright instead of dividing it. The `t.Logf` above the assertion reports base and big every run, so a zero is visible in the meantime.

**The three remaining unguarded divisions in the file feed `t.Logf` only**, so a `NaN` there is cosmetic. Checked rather than assumed — I mapped all six divisions in the file to their assertions:

| division | assertion | guarded |
|---|---|---|
| **330** | 335 `if ratio > 2.5` | **no — fixed here** |
| 354, 394, 550 | none (log only) | n/a |
| 489 | 495 `if fBase > 0 && …` | yes |
| 615 | 649 `if tBase > 0 && …` | yes |

## Test plan

`make pr-checklist BASE=origin/main` → **10 ran, 0 failed, 1 needs manual**. Test-only change; no production file touched.

```
4 redaction                       N/A — no production code changed
5 suppression                     N/A — no production code changed
6 timing / O(n^2)                 N/A — no production code changed
7 all 7 formats                   N/A — no production code changed
12 flake (3 repeats)              RAN — ./internal/detector
```

- [x] **10 — non-vacuity**: the two mutations above. The false failure reproduces without the guard and disappears with it; the genuine regression is still caught at 6.28x. One earlier mutant did not compile (`declared and not used` on the memo's locals) and was re-crafted rather than counted as a survivor.

**Merge test:** `git merge --no-commit --no-ff` against [#597](https://github.com/awslabs/ferret-scan/pull/597) — **CLEAN MERGE**, union **builds**, and `./internal/detector` + `./internal/perfguard` both pass together. CHANGELOG hunk is at line 60; #597's is at 41, disjoint.

Related: #597 (the windows break that prompted the audit), #596 (the unmeasured granularity this defers to).


sectoramen:
Superseded by #601, which carries this change unmodified (cherry-picked with `-x` provenance) together with the other clock-measurement audit fix and two CI fixes.

Rolled up at the maintainer's request to cut three reviews and three CI runs down to two. #597 stays separate because it fixes a live red on `main`.

Nothing is dropped: #601's checklist is 10 ran / 0 failed with race green at 84 packages, all four mutations re-verified in the combined tree, and it merges clean with #597.

Issue #599, opened 2026-09-04
complexity guard: the 8.0 threshold was chosen on one platform, and an emit-heavy O(n²) regression may slip under it on ubuntu


sectoramen:
Found while diagnosing intermittent ubuntu-latest failures. The immediate fixture defect is fixed separately; **this is the part deliberately left open**, because settling it needs per-platform data that has never been observable.

## What was measured

`TestGrowthRatioStillCatchesAGenuineQuadratic` is the control proving the 8.0 threshold can still detect a genuine O(n²). It **failed on ubuntu-latest in two separate runs**:

| platform | ratio | base | big |
|---|---|---|---|
| ubuntu-latest ([run 33898676195](https://github.com/awslabs/ferret-scan/actions/runs/33898676195)) | **7.46x** | 12.419 ms | 92.621 ms |
| ubuntu-latest ([run 33887302424](https://github.com/awslabs/ferret-scan/actions/runs/33887302424)) | **7.85x** | 11.177 ms | 87.752 ms |
| darwin/arm64 (local, 3 runs) | 13.74x – 14.38x | ~4.0 ms | ~55 ms |

Per-pair samples `[8.22x 7.41x]` straddle 8.0, which is why it was intermittent rather than constant.

## Why, in one equation

For a 4x input step the measured ratio is

```
ratio = 4 · (1 + 3f)        f = the quadratic share of the BASE reading
```

`f = 1` → 16x; `f = 0` → 4x. Solving from the measurements: **local `f` = 0.80, ubuntu `f` = 0.29.**

The dilutant is the per-match `detector.Match` emit. It is a **linear** cost, `-race` instruments every copy of that large struct, and on ubuntu it dominated the base reading.

## The part that is fixed, and the part that is not

**Fixed** (separate PR): the control was measuring `Match`-copy cost rather than the algorithm. Removing that incidental cost — `strings.Index` scanning, a pre-sized slice, and emitting only every 64th match while still running the full-input scan for **every** match — takes `f` to ~0.99 and the ratio to **15.73x–16.09x**, tight, against a theoretical 16x. The O(n²) work is unchanged.

**Not fixed, and the reason this issue exists:** a real validator **emits a `Match` per finding**. So a genuine O(n²) regression in a real validator would carry exactly the linear cost that was just removed from the control, and on ubuntu it could therefore read **~7.5x and slip under the 8.0 gate**. The control was, accidentally, reporting that — and making the control clean removes the messenger without removing the risk.

[...]

Issue #586, opened 2026-09-04
test: the wall-clock scaling guards #581 did not reach still flake on CI — svgtextlib takes a single sample per size


sectoramen:
#581 replaced the complexity guard's statistic with CPU time, GC disabled, and a ratio of minimums, because a wall-clock ratio had stopped ordering quadratic above linear — under load it scored a genuine O(n²) validator at **9.94x** and a single-pass one at **10.20x**, inverted.

**That fix reached `internal/goldencorpus` only.** The same estimator problem exists in guards elsewhere, and one has now failed on CI.

## Observed failures, not speculation

| test | package | where | statistic |
|---|---|---|---|
| `TestConfirmationStillCatchesAGenuineQuadratic` | goldencorpus | ubuntu CI, on #576 | median of 7 wall-clock pairs — **fixed by #581** |
| **`TestCollapseSpaceIsLinear`** | **svgtextlib** | **macOS CI, [run 33825436653](https://github.com/awslabs/ferret-scan/actions/runs/33825436653/job/100876972010) on #585** | **single wall-clock sample per size** |
| `TestAssignLineColumnsMemoIsLoadBearing` | detector | local, once, full suite | best-of-3 wall clock |

The macOS failure:

```
--- FAIL: TestCollapseSpaceIsLinear (0.18s)
    collapseSpace scaled 3.37x per doubling at n=4194304
    (linear is 2x, quadratic 4x): 35.170083ms -> 118.467625ms
```

Windows and ubuntu passed on the identical commit, and #585 touches none of that package — `CHANGELOG.md`, `cmd/main.go`, `docs/COVERAGE_DISCLOSURE.md` and three files under `internal/formatters/`. So it failed for reasons unrelated to the change under test, which is the cost being paid: **every PR carries a chance of a red job on an unrelated axis.**

## Ranked by how fragile the statistic is

Measured by reading each guard, not assumed:

1. **`svgtextlib` — worst.** `TestCollapseSpaceIsLinear` and the guards in `svg_complexity_test.go` take **exactly one** `time.Now()`/`time.Since` sample per size and divide consecutive singles against a fixed threshold. No repetition, no median, no minimum, GC on. A single pair is not a statistic — this file's own sibling in goldencorpus says so in a comment.
[...]

Issue #579, opened 2026-09-02
test(complexity): the quadratic control's 2-point ratio is biased downward by a fixed base overhead, so it fails on ubuntu under -race despite the warm-up added for it


sectoramen:
`TestConfirmationStillCatchesAGenuineQuadratic` (`internal/goldencorpus/complexity_confirmation_test.go`) failed on `chore/go-1.27.1` ([run 33636468050](https://github.com/awslabs/ferret-scan/actions/runs/33636468050/job/100268478877)), ubuntu only. macOS and Windows passed on the identical commit, and that PR touches **zero** files under `internal/` — only `.gitlab-ci.yml`, `.go-version`, `Dockerfile`, `go.mod`, `scripts/go-version.sh` and docs. So the Go bump is not implicated; the control is.

```
--- FAIL: TestConfirmationStillCatchesAGenuineQuadratic (1.10s)
    synthetic quadratic first reading: 9.27x (base=13.458089ms big=124.764043ms)
    synthetic quadratic median: 7.93x over 7 pairs [9.93x 5.69x 6.46x 6.57x 7.93x 9.28x 11.50x]
    a genuine O(n^2) validator MEDIAN ratio is 7.93x, at or below the 8.0 threshold
```

## The mechanism, from the test own comments

The fixture steps size by **4x** (`reps` -> `reps*4`), so a true quadratic should read **~16x**, and the comments state the expected median is **14.3x-15.9x**. Observed here: **7.93x**, roughly half.

The comments already diagnose why, and name the numbers: on CI the base read `11.7-15.2ms` against `2.9-4.7ms` locally while `big` was only ~2.3x its local value, so the base absorbs a one-off cost `big` does not pay in proportion, and the ratio collapses. The fix applied was to warm up and discard.

**That fix did not hold.** This run base read `13.458ms` — inside the same inflated 11.7-15.2ms band — *after* the warm-up. So the inflation is not purely cold-start; it persists under `-race` on a contended shared runner.

## Why this is structural, not a threshold to retune

A 2-point ratio is `t(4n)/t(n)`. With a fixed additive overhead `c` in each reading, that becomes `(16T + c)/(T + c)`, which is biased **downward** and approaches 1 as `c` grows. No choice of threshold fixes a biased estimator: it can only trade this failure against the opposite one. That trade is already on record —

[...]

Issue #546, opened 2026-08-28
The complexity guard's tightened 8.0 growth ratio fails on CORRECT code under load and under -race, where the old 12.0 passed


sectoramen:
## What happens

#535 replaced `if ratio > 12.0` with `const maxGrowthRatio = 8.0` in
`internal/goldencorpus/complexity_guard_test.go`. On a loaded host that threshold fails on code with
no defect in it.

## Measured

`TestValidatorComplexityIsSubQuadratic` on `origin/main` versus the pre-#535 baseline:

```
main      4 of 9 plain runs FAILED,  1 of 2 -race runs FAILED
baseline  0 of 8 plain runs failed,  0 of 2 -race runs failed
```

Five failing readings sit in `(8.0, 12.0]` and therefore fail **only** because of the tightening —
every one passes at the old 12.0:

| validator | ratio | mode |
|---|---|---|
| dob | 8.60x | -race |
| intellectualproperty | 8.92x | plain |
| vin | 9.03x | plain |
| secrets | 9.49x | plain |
| medicalid | 9.68x | plain |

Independently reproduced: an interleaved 3-round A/B of the `cloudresources` sub-case showed **plain
`origin/main` failing at 12.5x** while an unrelated branch passed all three rounds — the packages are
byte-identical, so the reading is environmental, not a code difference.

## Why the justification does not hold on this host

#535 chose 8.0 from 144 correct-code readings that maxed at **4.68x**, with a 1.71x margin below the
bound under `-race`. This host busts that: correct-code readings reached **9.68x plain and 8.60x
under `-race`**. Load average was 143–175 on 14 CPUs, which is abnormally high — but CI runs this
suite **with `-race` on shared runners**, which is exactly the contended case.

Related, same root cause: several DoS/complexity guards fail under `go test ./...` default
parallelism and pass with `-p 1`. Measured contention: phone 10.43s → 1.54s, personname 10.50s →
2.85s, the ipaddress package 112s → 28s. The trigger is `tests/integration` (371–428s) running
alongside 5s absolute ceilings.

## Why this is not "just raise it back"

The tightening was the point of #535 — a guard that cannot detect the defect it names is worse than
no guard. So the fix is to make the **measurement** robust rather than to loosen the bound:

- The ratio is only stable while one term dominates. `cloudresources` is the sub-case #535 did **not**
  resize a fixture for, and it is the one that flakes worst; resizing it so the quadratic term
[...]

PR #600, opened 2026-09-04
fix(tests): stop the quadratic control measuring its own fixture overhead, so the guard detects O(n²) on ubuntu

sectoramen:
Found by chasing the intermittent `ubuntu-latest` failures that I had initially dismissed as infrastructure. They were two different things, and this is the real one.

## The defect

`TestGrowthRatioStillCatchesAGenuineQuadratic` is the control that proves the 8.0 threshold can still detect a real O(n²). **On ubuntu-latest it measured below its own threshold:**

| platform | ratio | base | big |
|---|---|---|---|
| ubuntu-latest ([run 33898676195](https://github.com/awslabs/ferret-scan/actions/runs/33898676195)) | **7.46x** | 12.419 ms | 92.621 ms |
| ubuntu-latest ([run 33887302424](https://github.com/awslabs/ferret-scan/actions/runs/33887302424)) | **7.85x** | 11.177 ms | 87.752 ms |
| darwin/arm64 (local) | 13.49x – 14.06x | ~3.9 ms | ~53 ms |

So on the primary CI platform the guard **could not detect the regression it exists for**, and #581's claim that 8.0 sits "1.75x below the regression" held on one machine only. Per-pair samples `[8.22x 7.41x]` straddle 8.0, which is why it failed intermittently rather than every run.

## Cause

For a 4x input step the measured ratio is

```
ratio = 4 · (1 + 3f)        f = the quadratic share of the BASE reading
```

`f=1` → 16x, `f=0` → 4x. Solving from the readings: **local `f` = 0.80, ubuntu `f` = 0.29.** ubuntu's base was 11–12 ms against 3.9 ms here — dominated by the fixture's own linear cost, not by the algorithm.

Raising `reps` cannot fix that: the model puts the requirement at **~36,000 reps**, minutes of runtime per run. So the fixture's linear overhead is removed instead — the output slice is **pre-sized** and the match scan uses **`strings.Index`** rather than a per-byte slice compare.

## I measured the contributions in isolation, because my first two explanations were both wrong

Three runs each, `-race`, darwin/arm64:

| variant | ratio |
|---|---|
| as it was (append per match, per-byte scan) | 13.49x – 14.06x |
| pre-sized only | 14.65x – 14.89x |
| `strings.Index` only | 13.30x – 13.78x ← **no effect alone** |

The ordering is not obvious, and is now recorded in the code: `strings.Index` does nothing by itself because repeated slice growth dominates and hides the scan. Once pre-sized, the per-byte compare becomes a visible linear term and removing it closes the gap.

**A third change was written and then dropped for lack of evidence.** Emitting only every 64th match measured **15.87x against 15.82x** — nothing. So the control still emits a `Match` per match, which also keeps it faithful to how real validators behave. Asserting a mechanism I had not isolated is exactly what produced the two wrong explanations above, and I had claimed the emit reduction was *the* fix in an earlier draft of this commit.

## Verified by mutation

| mutation | result |
|---|---|
| make the validator **linear** (drop the per-match full scan) | **control FAILS** — it still detects a non-quadratic |
| remove the pre-size (`var out []detector.Match`) | ratio drops **15.61x → 14.27x** |

The **linear** control is unaffected at 4.10x – 4.15x, so this does not buy regression-detection with false failures on correct code — the trade this guard has already made twice (#546, #579).

Its sibling already carried the lesson: the linear control's doc comment says `detector.Match` *"is a large struct and `-race` instruments every copy, which is what made an append-driven version read 8.17x on a CI runner for a single-pass scan"*, and it was pre-sized for that reason. The quadratic control was left append-driven.

## What is deliberately NOT in this PR

**1. The residual threshold weakness — filed as [#599](https://github.com/awslabs/ferret-scan/issues/599).** A real validator emits a `Match` per finding, so a genuine O(n²) regression in one carries linear cost this synthetic no longer has, and on ubuntu it could still read near the gate. The control was, accidentally, reporting that; making it clean removes the messenger, not the risk. Settling it needs each runner's correct-code population.

**2. The CI step that would print those ratios.** `go test` hides `t.Logf` for a passing package without `-v`, so the guard's ratios have never appeared in any CI log. I wrote the step, then removed it: it inserts at the same point in `.github/workflows/go-test.yml` as the one [#597](https://github.com/awslabs/ferret-scan/pull/597) adds, and **the two collide — measured with a real merge, not guessed.** It lands separately once #597 merges, so this PR stays conflict-free.

## Test plan

`make pr-checklist BASE=origin/main` → **10 ran, 0 failed, 1 needs manual**. Test-only change; no production file touched.

```
4 redaction                       N/A — no production code changed
5 suppression                     N/A — no production code changed
6 timing / O(n^2)                 N/A — no production code changed
7 all 7 formats                   N/A — no production code changed
12 flake (3 repeats)              RAN — ./internal/goldencorpus
-- tree clean                     RAN — this run dirtied nothing
```

- [x] **10 — non-vacuity**: the two mutations above.

**Merge test:** `git merge --no-commit --no-ff` against **#597** and **#598** — both **CLEAN MERGE**; and the union of all three actually merged, **built**, and `goldencorpus` + `perfguard` + `detector` all pass together. CHANGELOG hunk at line 72; #597's at 41 and #598's at 60, disjoint.

Related: #581 (chose 8.0 and this estimator), #586, #596, #597, #598.


sectoramen:
Superseded by #601, which carries this change unmodified (cherry-picked with `-x` provenance) together with the other clock-measurement audit fix and two CI fixes.

Rolled up at the maintainer's request to cut three reviews and three CI runs down to two. #597 stays separate because it fixes a live red on `main`.

Nothing is dropped: #601's checklist is 10 ran / 0 failed with race green at 84 packages, all four mutations re-verified in the combined tree, and it merges clean with #597.

PR #601, opened 2026-09-04
fix(tests, ci): guard an unguarded ratio divisor, and drop a cache step that fails every run

sectoramen:
Rollup of **#598** and **#600**, which are closed in favour of this — same content, cherry-picked with `-x` provenance, plus two CI fixes that only became possible/visible once they were together. Kept separate from **[#597](https://github.com/awslabs/ferret-scan/pull/597)** on purpose: that one fixes a **live** red on `main` and should merge alone.

All of this came out of auditing the repo for the same class of assumption behind the windows-latest break in #597.

---

## 1. An unguarded divisor reports a false O(n²) on correct code

`internal/detector/line_span_complexity_test.go` computed `ratio := float64(tBig) / float64(tBase)` and asserted `ratio > 2.5` **with no `tBase > 0` check** — while the **two other ratio assertions in the same file already carry it** (`if fBase > 0 && …` at 495, `if tBase > 0 && …` at 649). An inconsistency inside one file, not a house style.

It matters because this is the **smallest timed workload in the repo** (base 358 µs plain, 1.1 ms under `-race`), and a reading that small can be exactly **zero** where the clock ticks coarsely — windows-latest returned a zero *wall* reading for a ~90 µs workload, which is what broke [run 33899001272](https://github.com/awslabs/ferret-scan/actions/runs/33899001272).

| readings | ratio | `> 2.5` | result |
|---|---|---|---|
| `tBase == 0`, `tBig > 0` | `+Inf` | **true** | *"the per-match line hash is back"* on **correct code** |
| `tBase == 0`, `tBig == 0` | `NaN` | false | **passes having measured nothing** |

Reproduced against the real test — forcing `tBase = 0` without the guard yields `cost +Infx more (base=0s big=276.084µs)`; with it, passes. And the guard **still catches what it exists for**: destroying the line-id memo fails at **6.28x** (base 5.131 ms), so the check did not quietly switch the assertion off.

I mapped **all six** divisions in the file: the other three unguarded ones feed `t.Logf` only, so a `NaN` there is cosmetic. Checked, not assumed.

**Deliberately not also failing when `tBase == 0`** — that would trade a false pass for a false *failure* on any platform whose granularity exceeds this fixture, and that granularity is still unmeasured. Assuming it is the mistake that caused the break this hardens against. Deferred to #596.

## 2. The O(n²) guard could not detect a quadratic on ubuntu

`TestGrowthRatioStillCatchesAGenuineQuadratic` — the control proving the 8.0 threshold still works — measured **below its own threshold** on ubuntu:

| platform | ratio | base | big |
|---|---|---|---|
| ubuntu-latest ([33898676195](https://github.com/awslabs/ferret-scan/actions/runs/33898676195)) | **7.46x** | 12.419 ms | 92.621 ms |
| ubuntu-latest ([33887302424](https://github.com/awslabs/ferret-scan/actions/runs/33887302424)) | **7.85x** | 11.177 ms | 87.752 ms |
| darwin/arm64 | 13.49x – 14.06x | ~3.9 ms | ~53 ms |

So #581's claim that 8.0 sits "1.75x below the regression" held on **one machine**. Per-pair samples `[8.22x 7.41x]` straddle 8.0, hence intermittent.

For a 4x step, `ratio = 4·(1+3f)` where `f` is the quadratic share of the base: **local `f` = 0.80, ubuntu `f` = 0.29** — ubuntu's base was dominated by the fixture's own linear cost. Raising `reps` can't fix it (the model needs ~36,000 — minutes per run), so the linear overhead is removed instead: pre-sized output slice, and `strings.Index` instead of a per-byte slice compare.

**I measured the contributions in isolation, because my first two explanations were both wrong:**

| variant | ratio |
|---|---|
| as it was | 13.49x – 14.06x |
| pre-sized only | 14.65x – 14.89x |
| `strings.Index` only | 13.30x – 13.78x ← **nothing alone** |

The ordering is non-obvious and now recorded in the code: `strings.Index` does nothing until the slice is pre-sized, because slice growth dominates and hides the scan. **A third change was written then dropped for lack of evidence** — emitting every 64th match measured 15.87x vs 15.82x, i.e. nothing — so the control still emits a `Match` per match, which also keeps it faithful to real validators.

Mutation-verified: making the validator linear still **fails** the control; removing the pre-size drops it 15.61x → 14.27x. The **linear** control is unaffected at 4.10x – 4.15x, so this doesn't buy detection with false failures on correct code — the trade this guard has already made twice (#546, #579).

## 3. CI: a cache step that fails on every cache hit

`actions/setup-go` already caches **both** paths a manual step would name — `~/go/pkg/mod` and `~/.cache/go-build`, keyed on `go.sum` — **by default**. The separate `actions/cache` over the same paths therefore always loses the race to extract. From an ubuntu log:

```
setup-go:  Cache hit ... Cache Size: ~678 MB ... Cache restored successfully
           /home/runner/go/pkg/mod
           /home/runner/.cache/go-build
cache:     path: ~/.cache/go-build, ~/go/pkg/mod ... Cache Size: ~677 MB
           /usr/bin/tar: ...: Cannot open: File exists          (many)
           Failed to restore: "/usr/bin/tar" failed with exit code 2
```

Every cache-hit run **re-downloaded ~677 MB and then failed to unpack it**. Removed.

It also correlates with the ubuntu jobs that die: **all 3** killed by `SIGTERM` at ~245 s carried this failure, against **1 of 5** that passed. I am **not** claiming it is the cause — n is small, and I falsified my other three hypotheses for that failure by measurement (suite peak **3.6 GB** vs a 16 GB runner; `-race` compile **377 MB**; build cache **375 MB** vs ~14 GB free). Removing a step that always fails is right on its own terms.

## 4. CI: make the guard's ratios visible

`go test` hides `t.Logf` for a **passing** package unless `-v` is set (`fmt.Println` too — both measured), which is why the 8.0 threshold has been resting on one machine's numbers with no per-platform data to check it against. A `continue-on-error` step now prints all 18 target ratios plus both controls on all three runners.

Anchored after **Build**, not after the test step: the granularity step in #597 anchors there, and two insertions at the same point conflict — **measured with a real merge, not guessed.** That is what lets this ship now rather than waiting for #597.

## Test plan

`make pr-checklist BASE=origin/main` → **10 ran, 0 failed, 1 needs manual**. Test/CI-only; no production Go file touched.

```
4-7 redaction/suppression/timing/formats  N/A — no production code changed
12 flake (3 repeats)              RAN — ./internal/detector ./internal/goldencorpus
```

- [x] **10 — non-vacuity**: the four mutations above (`+Inf` guard both ways; linear validator; pre-size removal).
- [x] **actionlint** on the workflow: 1 issue, the **pre-existing** SC2046 on the unchanged `go test -race` line — verified identical on `origin/main`.

**Worth flagging:** an earlier checklist run of this branch failed `cmd` under `-race`. It did **not** reproduce — 3/3 clean in isolation, then a full `-race` suite at exit 0, then a clean checklist. It was load from my own concurrent measurements, not this change. Reported rather than quietly re-run.

**Merge test:** `git merge --no-commit --no-ff` against **#597** — **CLEAN MERGE**, re-verified after every commit including the CHANGELOG edit, and the merged workflow carries **both** reporting steps.

Supersedes #598 and #600. Related: #596, #599, #581, #586.


sectoramen:
## Scope reduced: the quadratic-control change is reverted, and windows-latest is why

`windows-latest` failed on this branch and **passes on `main`**, so the quadratic-control fixture change (a707cea) was harmful. Reverted in 74efa10 — that file is now byte-identical to `main`.

The same CI run answered the question **#596** was filed to ask. Every CPU reading on windows-latest is an exact multiple of **15.625 ms**:

```
linear control   base=15.625ms   big=31.25ms     ->  1 tick / 2 ticks  = "2.00x"
ssn              base=46.875ms   big=234.375ms   ->  3 ticks / 15 ticks = "5.00x"
also: 78.125  93.75  125  140.625  156.25  546.875   (all exact multiples)
```

So `MinMeasurableCPU = 2ms` is **0.128 of a single tick**, and every Windows ratio this guard produces is a ratio of small integers — the linear control's `2.00x` is literally two ticks over one. On that platform the guard has been comparing tick counts, not measuring code.

That **inverts** what the reverted change was for:

| platform | what the base needs | making the fixture cheaper |
|---|---|---|
| ubuntu | a larger quadratic **share** (`f` was 0.29) | helps |
| windows | a longer **absolute time**, to span many ticks | **hurts** |

The two platforms want opposite fixture changes, so no fixture change fixes both. The real fix is a floor expressed in **ticks of the measured granularity** rather than a fixed 2 ms — now filed with hard numbers on #596, and #599 for the threshold margin.

## What remains in this PR

Three things, all unambiguous and unaffected by the above:

1. **The unguarded ratio divisor** in `internal/detector/line_span_complexity_test.go` — `x/0 = +Inf > 2.5` reports a false O(n²) on correct code, while the two sibling assertions in the same file already guard it. Reproduced (`cost +Infx more (base=0s big=276.084µs)`), and the guard still catches the real regression at 6.28x.
2. **The redundant `actions/cache` step**, which fails on every run `setup-go` gets a cache hit — proven from the log, re-downloading ~677 MB and then failing to unpack it.
3. **The ratio-reporting CI step**, anchored after Build so it survives a killed test step.

`internal/goldencorpus` is untouched now. The second Windows failure in that run, `TestATinyWorkloadFallsBackToWallClock`, is the **pre-existing** break that **#597** fixes — this branch is based on `main`, which does not have it yet, so that failure is expected here and not caused by this PR.

sectoramen:
Superseded by #602, which carries this content (cherry-picked with `-x` provenance) plus the **root-cause fix** underneath it.

Rolled into one PR at the maintainer's request. The reason it is one and not three: the windows-latest failure here turned out to be a symptom of `MinMeasurableCPU` being a fixed 2 ms while Windows' clock tick is **15.625 ms** — so the gate was 0.128 of a single tick and the guard was dividing integer tick counts (a linear control reporting "2.00x" from 2 ticks over 1). Fixing that in #602 is also what makes the ubuntu fixture fix safe, which is why splitting them was the wrong shape.

Nothing is dropped: #602's checklist is 10 ran / 0 failed with race green at 84 packages, and every mutation from this PR is re-verified there.

PR #602, opened 2026-09-04
fix(tests, ci): close the clock-measurement gaps — Windows was dividing tick counts, not measuring code

### LINKED ISSUE #596: perfguard: MinMeasurableCPU is 2ms while the constant's own doc says Windows granularity is 15.6ms, so a quantised reading can clear the gate

sectoramen:
Supersedes **#597** and **#601**, which are closed in favour of this. One PR for the whole class, as requested — and it fixes the **root cause** rather than the three symptoms, which is what let a fix I'd previously reverted come back safely.

It started as a one-line Windows test failure and ended up showing that the O(n²) guard has never measured anything on windows-latest.

---

## The root defect: the resolution gate was 0.128 of a single tick

`MinMeasurableCPU` was a fixed **2 ms**, and the constant's own doc comment asserted *"15.6ms on Windows by default"* with no evidence behind it. That figure is now **measured, and exact** — every CPU reading on windows-latest is an integer multiple of **15.625 ms**:

```
linear control   base=15.625ms   big=31.25ms      ->  "2.00x"   = 2 ticks / 1 tick
ssn              base=46.875ms   big=234.375ms    ->  "5.00x"   = 15 ticks / 3 ticks
also observed:   78.125  93.75  125  140.625  156.25  546.875   (all exact multiples)
```

The readings that are *not* multiples in the same log (13.6137, 18.0832, 51.7005, 4.4544, 5.0413) are precisely the **wall-clock fallbacks** — itself confirmation that the CPU path quantises and the wall path does not.

So **the gate was 0.128 of one tick**, and the guard was doing arithmetic on integer tick counts. Neither number above measured code, and a 1-tick base can quantise in either direction — so on that platform the guard could **fail on correct code and pass a real regression**.

### The fix: a floor relative to the measured tick

- **`perfguard.ClockResolution()`** finds each clock's granularity by **tight-polling until it changes** — measuring the *clock*. An earlier attempt timed small workloads instead, and reported `866µs` then `1.912ms` on the same machine depending on the fixture, which is how that error was caught.
- **`Growth.Ticks()`** reports how many ticks the base spans, against **`MinTicks = 8`** — so one tick of quantisation per side moves the ratio by at most ~1/8. Chosen against the numbers: at 2 ticks a single tick is 50% error and the 4x-vs-16x decision is a coin flip; at 32 a Windows base would need half a second and every fixture in the repo would need resizing.
- **All four ratio-asserting guards** — the 18-target validator guard, both confirmation controls, and the two SVG guards — assert **only** when the base clears the floor, and log `ResolutionNote()` (which always states the tick count) when it does not.

**A skip is not a silent hole**, which was the obvious risk:

| still asserts everywhere | asserts where the clock allows |
|---|---|
| absolute wall-clock ceilings | growth ratios (Linux, macOS — a ~3.4 ms base against a 1 µs tick is **~3,400 ticks**) |
| match-count non-vacuity floors | |

and the tick count now reaches a CI log on every runner. Asserting on 1-tick readings was the hole.

## What this unlocked: the ubuntu fix, back and now safe

The synthetic-quadratic control measured **7.46x and 7.85x on ubuntu — below its own 8.0 threshold**, i.e. the guard could not detect the regression it exists for there either. Cause: its base was dominated by fixture overhead, not the algorithm. For a 4x step `ratio = 4·(1+3f)` where `f` is the quadratic share of the base — **`f` was 0.29 on ubuntu against 0.80 locally**.

Pre-sizing the output slice and scanning with `strings.Index` lifts it to **14.60x – 15.31x**. I shipped exactly that in #601 **without** the gate and it **broke windows-latest**, because a cheaper base spans *fewer* ticks:

| platform | base needs | cheaper fixture |
|---|---|---|
| ubuntu | larger quadratic **share** | helps |
| windows | longer **absolute** time | **hurts** |

The two are opposite, so no fixture change fixes both — which is why it needed the gate underneath it. With the gate, Windows declines to assert and ubuntu asserts on a real measurement.

**Contributions isolated rather than assumed**, three runs each under `-race`, because my first two explanations were both wrong:

| variant | ratio |
|---|---|
| as it was | 13.49x – 14.06x |
| pre-size only | 14.65x – 14.89x |
| `strings.Index` only | 13.30x – 13.78x ← **nothing alone** |

`strings.Index` does nothing by itself because slice growth hides the scan until it is pre-sized. **A third change was written and dropped for lack of evidence** — emitting every 64th match measured 15.87x against 15.82x.

## The other three fixes

**The Windows coarse-clock test** (was #597). `TestATinyWorkloadFallsBackToWallClock` broke `main` deterministically: it required a ~90 µs workload to be measurable on the wall clock, and Windows reads zero on *both* clocks. The premise was unsatisfiable, not unlucky — the window it aimed at can be **empty**. It now asserts on the **clock named in the result**, accepting both correct outcomes and failing only on a `cpu` reading being divided. Verified three ways: passes the simulated Windows state, still catches the forced-CPU mutation, and catches it **even on a coarse clock**.

**An unguarded ratio divisor** (was #601). `internal/detector` asserted `ratio > 2.5` with **no `tBase > 0`** while the two siblings in the same file already had it. `x/0 = +Inf > 2.5` is **true**, so a zero base reports *"the per-match line hash is back"* on correct code — reproduced as `cost +Infx more (base=0s big=276.084µs)`. The guard still catches the real regression at **6.28x**. All six divisions in the file were mapped; the other unguarded three feed `t.Logf` only.

**CI** (was #601). Removed a redundant `actions/cache` step: `actions/setup-go` already caches both paths by default, so the second restore always lost the extract race — every cache-hit run **re-downloaded ~677 MB and then failed to unpack it**, proven from the log. Plus two `continue-on-error` reporting steps for granularity and the guard's ratios, at **different anchors** because two insertions at the same point conflict (measured with a real merge).

## Test plan

`make pr-checklist BASE=origin/main` → **10 ran, 0 failed, 5 need manual**.

```
12 flake (3 repeats)              RAN — detector, goldencorpus, perfguard, svgtextlib
```

The 5 manual items are flagged because `perfguard/resolution.go` is a non-test file. Discharged by establishing it **cannot reach the scanner**:

- [x] **4 / 5 / 7 — redaction, suppression, formats: N/A, proven not asserted.** `internal/perfguard` has **zero non-test importers** (all 6 are `_test.go`), and `go list -deps ./cmd | grep -c perfguard` returns **0** — it is not in the shipped binary. No validator, redactor, formatter or CLI file is in the diff.
- [x] **6 — timing:** the added measurement is the thing to watch, since it burns CPU deliberately. **6.792 µs** on first call, **41 ns** cached (`sync.Once`), once per test binary. On Windows it costs ~1 tick per attempt (~47 ms total), bounded. The guards' own ratios are measured throughout and reported above.
- [x] **10 — non-vacuity, mutation-verified:** `MinTicks` at 1 admits the Windows reading; at 128 admits nothing; a gate that always reports "sufficient"; a removed degenerate-input guard (`base/0` is `+Inf`, and **`+Inf >= MinTicks` is true**, so an unmeasurable reading would read as *ample*) — all **caught**. Plus the earlier mutations for the divisor guard and the Windows fallback test.
- [x] **Windows test binaries cross-compile** (`GOOS=windows go test -c`) for both changed packages — the platform that broke must at least build.

**A mistake worth recording:** a mutant making the gate always report "sufficient" initially **survived**, because my test computed tick counts with its *own* arithmetic instead of calling the function. The decision is now a pure `ticksAt(base, res)`, which both fixes the vacuity and lets windows-latest's real numbers be exercised from a darwin host. 4/4 caught after.

Closes #596. Supersedes #597 and #601. **#599** stays open — the emit-heavy threshold margin is a separate design question, and this PR's CI step is what will supply the per-platform data to settle it.


sectoramen:
## Correction: the cache-restore correlation is REFUTED

The PR body says all 3 ubuntu jobs killed by `SIGTERM` at ~245s carried the failed `tar` restore, against 1 of 5 that passed — described as *"suggestive, NOT established, and deliberately not claimed as the cause."*

**It is now refuted, by this PR's own run.** [Job 101151917461](https://github.com/awslabs/ferret-scan/actions/runs/33912500088/job/101151917461) on this branch:

```
tar restore failed:  0        <- the redundant cache step is gone, so no failure
packages ok:         22
exit:                143 + "The runner has received a shutdown signal"
elapsed:             20:02:34 -> 20:07:36  (302s)
```

The cache failure is absent and the SIGTERM still happened. So removing that step does **not** fix Mode A, and the 4-of-4 correlation was coincidence — three of those runs simply shared a cache-key state.

**The cache step removal stays in this PR on its own merits**, unchanged: `actions/setup-go` already caches both paths by default, so the second restore always lost the extract race and every cache-hit run re-downloaded ~677 MB before failing to unpack it. That is a real misconfiguration whether or not it relates to the runner deaths.

### Hypotheses now eliminated by measurement

| hypothesis | measured | verdict |
|---|---|---|
| test-suite OOM | 3.6 GB peak vs 16 GB runner | ✗ |
| `-race` **compile** OOM | 377 MB | ✗ |
| disk exhaustion | 375 MB build cache vs ~14 GB free | ✗ |
| concurrency cancellation | no `concurrency` block; sibling jobs survive the same run | ✗ |
| configured timeout | none in the workflow; default is 360 min | ✗ |
| **failed cache restore** | **absent in this run, SIGTERM anyway** | **✗** |

Also corrected earlier: *"it always dies at `internal/goldencorpus`"* was an artifact of `go test ./...` printing results in **list order** — that package is just the long pole, so the death point says nothing about the cause.

### What is left

`ubuntu-latest` only, never windows or macos; consistently **244–306s** in; mid-suite; `SIGTERM` with no test having reported `FAIL`. Six hypotheses eliminated, and the remaining ones are not visible from a log — hosted-runner reclamation or an agent crash.

Two things that would make the next occurrence diagnosable, neither of which I have added here because they are churn on a hunch rather than fixes:

1. A step sampling free memory and disk every 30s during the suite, so the next SIGTERM arrives with data instead of an absence.
2. `timeout-minutes` on the job, which would at least distinguish "killed at a boundary" from "killed at a duration".

Say which, if either, is worth having and I will add it. Meanwhile the practical unblock is unchanged: re-run the ubuntu job, since it passes roughly 5 times in 12.

rustic (approved):

Issue #293, opened 2026-08-12
Trust-boundary decision for a config discovered inside the scan target: opt-in gating, a THREAT_MODEL entry, and the pre-commit gate that silences the provenance disclosure


sectoramen:
Found while reviewing #285, tracing why a `pkg/scan` result depended on the process working directory.

## What happens

`FindConfigFile` checks the **current working directory first** (`config.yaml`, `ferret.yaml`, `ferret.yml`, `.ferret-scan.yaml`, `.ferret-scan.yml`) and only then the user config dir (`internal/config/config.go:352-376`). A config discovered that way can switch off detection — for example `validators.intellectual_property.disabled_types` — and nothing in the output says which config file governed the scan.

Reproduced on `main` @ 1824563, with `FERRET_CONFIG_DIR` pointed at an empty directory so only built-in defaults are in play:

```console
$ cat > notice.txt <<'EOF'
Copyright (c) 2026 Acme Corporation. All rights reserved.
Internal wiki: https://w.amazon.com/team
EOF

$ ferret-scan --file notice.txt --checks INTELLECTUAL_PROPERTY
[MEDIUM] INTELLECT... COPYRIGHT              86.00% line     1 [HIDDEN]   notice.txt
Files: 1 scanned | Findings: 1 (0 high, 1 medium, 0 low)

$ cat > .ferret-scan.yaml <<'EOF'
validators:
  intellectual_property:
    disabled_types:
      - copyright
      - internal_url
EOF

$ ferret-scan --file notice.txt --checks INTELLECTUAL_PROPERTY
No matches found.
```

Same binary, same flags, same file: 1 finding → 0, decided by a file sitting next to the scanned content. No stderr line names the config that was applied.

Project-local config is clearly deliberate — the code comment calls `.ferret-scan.yaml` "project-specific config" — so the feature is not the problem. Two consequences of it look worth a decision.

## 1. The scanned tree is a trust boundary the threat model does not evaluate

`THREAT_MODEL.md` TB-7 already names the case: "an outside contributor's PR run through a maintainer's pre-commit/CI". TM-11 covers that attacker driving confidence to zero through attacker-authored **content**. The same attacker can add a **config file** to the same PR and disable whole detection categories for the scan that is supposed to review them — a shorter path to the same outcome, and one no TM entry covers.

[...]

PR #603, opened 2026-09-04
fix(cli): stop pre-commit mode silencing the config-provenance disclosure (TM-13)

sectoramen:
A config file **inside the scanned tree** can switch detection off, and in pre-commit or CI the note that would name it was **suppressed** — so the scan reviewing a pull request could be silenced by a file in that same pull request, with nothing on stderr and exit 0.

## Measured on `main`

Scanning a copyright notice sitting beside a `.ferret-scan.yaml` that disables `copyright` and `internal_url`:

| environment | findings | disclosure | output | exit |
|---|---|---|---|---|
| no variable set | 1 → **0** | ✅ names the config | 126 B | 0 |
| `PRE_COMMIT=1` | 1 → **0** | ❌ | **0 B** | 0 |
| `_PRE_COMMIT_RUNNING=1` | 1 → **0** | ❌ | 0 B | 0 |
| `PRE_COMMIT_HOME=1` | 1 → **0** | ❌ | 0 B | 0 |
| `PRE_COMMIT_HOOK=1` | 1 → **0** | ❌ | 0 B | 0 |
| `GIT_HOOK_TYPE=1` | 1 → **0** | ❌ | 0 B | 0 |

`FindConfigFile` searches the **working directory first** (`config.yaml`, `ferret.yaml`, `ferret.yml`, `.ferret-scan.yaml`, `.ferret-scan.yml`) before the user config dir. That is deliberate — the code calls it "project-specific config" — and it stays. What was wrong is that the note naming a *discovered* config was gated on pre-commit mode, and **`IsPrecommitEnvironment()` is true from environment variables alone**, so silencing it took no flag at all.

The comment justifying that gate argued against itself:

> *"CI is exactly where a config that silently fails to apply is most dangerous, and CI is never interactive."*

…and then switched the warning off in the CI and git-hook path.

This is **THREAT_MODEL TB-7** (an outside contributor's PR run through a maintainer's pre-commit/CI) reaching **TM-11**'s outcome by a shorter route — shipping a config file is easier than crafting suppression padding — and no TM entry covered it. Added as **TM-13**, which the issue explicitly asked for.

## The fix, and why it is not noisy

The provenance note is no longer gated on pre-commit, `--quiet` or non-interactive. Noise is bounded by the function itself:

| situation | note |
|---|---|
| config **discovered** next to the working directory | printed |
| `--config` named a file explicitly | **silent** |

So a repo that deliberately keeps a project config acknowledges it with a flag and stays quiet, while an unexpected one is always named. Both directions are verified by test. It writes to stderr, so scripts parsing stdout are unaffected, and it never changes the exit code.

## What I got wrong, and how it was caught

My first version **also** ungated `warnUnknownConfigKeys` — which broke `TestUnknownConfigKeysSilentInPrecommit`, an existing deliberate decision with a test behind it. Narrowed to the provenance note only: a typo'd key means a setting the operator wrote did not apply, which is a correctness annoyance rather than an attacker's lever, and overturning that needs its own justification.

**That test lives in `tests/integration/`, which CI excludes** (`go list ./... | grep -v /tests/integration`). It surfaced only because the full local suite runs it — CI would have merged the regression.

## Test plan

`make pr-checklist BASE=origin/main` → **10 ran, 0 failed, 5 need manual**; all five discharged below with an A/B against a binary built from `origin/main`.

```
12 flake (3 repeats)              RAN — ./cmd
```

- [x] **4 — redaction, all 3 strategies**: redacted output **byte-identical** to base for `simple` and `format_preserving`, and **0 of 3 reported values survive** in any strategy for either binary. `synthetic` is not byte-comparable — a **positive control** shows the base binary differs from *itself* there — so the value-survival check is the oracle, and a **sanity control** confirms all 3 values *are* findable in the original file, so the check is not blind. *(My first leak probe used `find -exec grep -q`, which returns find's status rather than grep's and so reported "3 surviving" unconditionally. Fixed before drawing any conclusion.)*
- [x] **5 — suppression, base-generate / this-binary-apply**: 3 rules generated by base (all `enabled: false`, flipped on for the test) — **3 of 3 still match** under this binary, same as base. No hash-input change.
- [x] **6 — timing**: the change adds one conditional stderr write and no loop. Min of 3 full scans: **62 ms base, 63 ms here**.
- [x] **7 — all 7 formats**: stdout equivalent for all seven. `json` and `yaml` differ **only** in `duration_seconds`, and the base binary differs from itself in the same field — verified rather than assumed, since this change touches `cmd/main.go`.
- [x] **10 — non-vacuity, mutation-verified**: restoring the pre-commit gate **fails** the new guard (`PRE_COMMIT=1 suppressed the config-provenance disclosure`); neutering the fixture config so it disables nothing **fails** the non-vacuity floor beneath it (`did NOT suppress the finding, so the disclosure tests are describing a config with no effect`).

The new guard is **end-to-end against the built binary**, deliberately: the defect was in the caller's gate, not in `reportConfigProvenance`, and a unit test on that function passes with the gate wrong — which is how this shipped in the first place.

**Merge test:** `git merge --no-commit --no-ff` against **#602** — CLEAN MERGE, union builds. CHANGELOG entry is in 🔒 Security; #602's is in 🐛 Bug Fixes.

## Deliberately left open

**#293 stays open.** This makes the substitution *visible*, not *impossible*. The two remaining directions from the issue are unchanged:

1. **Require opt-in** for a config discovered inside the scan target (e.g. `--allow-project-config`, or honour it only when outside the scanned path). A behaviour change that needs a decision, since it would break existing project-config users.
2. **Report config-disabled types as a coverage signal** through the existing `Incomplete`/`IncompleteReason` plumbing — "validator sub-types disabled by config" is the same class of fact as the unexamined-files disclosure.

Refs #293. Adds TM-13 to `THREAT_MODEL.md`.


sectoramen:

|---|---|---|

rustic (approved):

PR #604, opened 2026-09-04
feat(rtf): redact a value split across formatting runs, via a source span map

### LINKED ISSUE #593: rtf: a value split across formatting runs is detected but cannot be redacted, so redaction is refused for that file
Follow-up to #421, which is where this limitation is introduced — deliberately, and disclosed, rather than shipped as a silent leak.

## The situation after #421

The RTF extractor reassembles a value a producer split across formatting runs. macOS `textutil` writes `452-11-9384` as

```
Employee SSN: 452-11-\f1\b 9384
```

so the value is now **reported**, where before #421 it was silently missed. That is the leak #421 closes.

But the reassembled value then occurs **nowhere literally in the file**, so the byte substitution `internal/redactors/rtf` performs (via the plaintext redactor's `RedactDocument`) finds nothing to replace. The residue check at the sink catches that and **refuses to write the file** rather than emit something an operator would read as redacted while it still holds the value.

## Measured, current behaviour

```
$ ferret-scan --file formatted.rtf --checks SSN,EMAIL,CREDIT_CARD \
    --enable-redaction --redaction-strategy simple --fail-on-incomplete
findings reported: 3 (SSN, BUSINESS, VISA)
stats: files_not_redacted=1  values_not_redacted=3
WARNING: redaction incomplete — 1 of 1 file(s) have findings but no redacted copy was written;
         the original values remain in cleartext:
  formatted.rtf: redacted RTF formatted.rtf still holds reported value(s) of type VISA, BUSINESS, SSN;
                 refusing to write it
exit=3
```

Identical for all three strategies. This is the correct end state given the constraint — detected, disclosed, non-zero exit, no destroyed file, no false all-clear — but it is not a redaction.

For comparison, before #421 the same file gave **0 findings, exit 0, no warning**. So this is strictly better; it is just not finished.

## What a fix needs

A byte-span map from the extractor's output back to the source. `parse` in `text-extract-rtftextlib` builds its output with a `strings.Builder`; it would need to also record, per contiguous run of literal text copied, a `{outStart, srcStart, srcEnd}` span. Recording per *run* rather than per byte keeps the memory proportional to the number of formatting runs rather than to the document size.

The redactor then locates a match by offset in the decoded text, maps it to one or more source ranges, and substitutes across them. The design question worth settling first: when a value spans runs, the intervening control words (`\f1\b`) sit *inside* the replaced region. Options are to write the replacement into the first range and delete the remaining value bytes while **preserving** the control words between them, or to collapse the whole span. The former keeps the document's formatting intact and is probably right, but it means the replacement is not contiguous.

## Verification a fix should show

- The `textutil` bold fixture redacts, with `files_not_redacted` back to 0 and exit 0.
- The redacted output is still valid RTF — `textutil -convert txt -stdout` reads the surrounding prose back, and the surrounding formatting survives.
- The residue check still fires for any value it genuinely cannot remove, i.e. the guard is not simply deleted once the common case works.
- `\'hh`-escaped values (`452\'2d11\'2d9384`) redact too — they are the same class, since the reported value does not appear literally either.
- The existing tests in `internal/redactors/rtf` continue to pass, in particular `TestRedactContentIsNotOffered`: the fix must stay on the `RedactDocument` path, because `RedactContent` writes the extracted prose over the file.

Related: #421 (introduces the extractor and this limitation), `internal/redactors/svg` (the same lossy-extraction pattern; its numeric-character-reference case is the same shape and is also a loud refusal).

sectoramen:
Closes the gap **#421 shipped deliberately and filed as #593**: an RTF value the producer split across formatting runs was detected and disclosed, but could not be removed, so the whole file was refused.

## Before

```
findings reported:      2
values_not_redacted:    2
redacted files written: 0
--fail-on-incomplete:   exit 3
```

The extractor reassembles `452-11-\f1\b 9384` into `452-11-9384`, so the value is **reported** — but it occurs nowhere literally, so the byte substitution found nothing and the residue check refused rather than write a file an operator would read as redacted. Honest, but not a redaction.

## The fix

The extractor now records a **span map** from each run of extracted text back to the source bytes that produced it, so the redactor rewrites the bytes that *made* the value even though the value never appears contiguously.

**The control words between the halves are left alone:**

```
Employee SSN: 452-11-\f1\b 9384\b0\par
Employee SSN: [SSN-REDACTED]\f1\b \b0\par
```

`\f1\b` survives. Collapsing the whole span would delete the run and change how the rest of the paragraph renders — data loss of a different kind.

**Measured across 5 fixtures, all three strategies:**

| | base | this PR |
|---|---|---|
| redacted files written | **1 of 4** | **4 of 4** |
| newly redactable | — | `split.rtf`, `escaped.rtf`, `twice.rtf` |
| joined value surviving | 0 | 0 |
| `9384` fragment surviving | — | **0** |
| RTF signature + `textutil` readback | ok | ok on all 4 |

Hex-escaped spellings redact too (`452\'2d11\'2d9384` → `[SSN-REDACTED]`), which needed a second fix: the escape emit sites passed an **already-advanced** offset, so every escape got an empty source range, its span was dropped, and any value containing one stayed unmappable.

## Two more defects the tests found, not review — both silent

1. **Merging spans whose output and source lengths differ.** An escape emits one byte from four, so a merged span looked like output `-11` against source `11`; slicing it computed an offset **inside a control word**.
2. **`strings.TrimSpace` at the end of `parse` shifts every output offset.** Clipping a span has to move **both** ends of the mapping — adjusting only the output end left it off by exactly the trimmed width, which a document with no leading whitespace would never reveal.

## Two mutations initially SURVIVED, and that reshaped the tests

This is the part worth reviewing. My first test suite validated the map *structurally*, and:

- **A structural round-trip test cannot see a span that is missing.** Emptying an escape's source range left the remaining spans self-consistent, so it passed.
- **Its escape allowance was too lenient** — "the lengths may differ if the source contains a backslash" waved through a wrongly merged span.

Replaced with the **functional property**: for **7 producer spellings** (contiguous, split once, split twice, hex-escaped, hex-escaped *and* split, leading-trimmed whitespace, value appearing twice), redacting must make the value absent from a **re-extraction** of the redacted bytes **and leave its neighbours intact**. That second half matters: an over-broad source range removes the value *along with the surrounding text* and passes every other check — which is exactly how the merge mutation escaped.

Mutations now caught: empty escape range, merge-regardless-of-1:1, TrimSpace re-basing removed, only-the-first-range-rewritten (leaves the tail digits), gap detection removed.

**One emit site is documented as uncovered rather than faked.** The literal-escape site (`\\`, `\{`, `\}`) survives every assertion, because catching it needs a value containing a brace or backslash and no realistic PII value has one. Its failure mode is also the safe one — a missing span is a *gap*, `SourceRanges` refuses a window covering a gap, and the redactor falls back to refusing rather than corrupting.

## Test plan

`make pr-checklist BASE=origin/main` → **10 ran, 0 failed, 5 need manual**; all five discharged against a binary built from `origin/main`.

```
12 flake (3 repeats)              RAN — rtftextlib, redactors/rtf
```

- [x] **4 — redaction, all 3 strategies**: the table above. `format_preserving` yields `***-**-9384`; that is the strategy's documented masking and is **byte-identical to what it produces for the same value in a plain `.txt`** — verified, because a preserved suffix looks like a leak until compared against the non-RTF path.
- [x] **5 — suppression, base-generate / this-binary-apply**: 7 rules from base, **7 of 7 suppressed** under both binaries. No hash-input change.
- [x] **6 — timing**: the map is recorded per **run**, not per byte — adjacent 1:1 emits extend the previous span — so it is proportional to the number of formatting runs, not to document size. The extractor's existing linearity guard (a 4x input allocating 3.97x) still passes.
- [x] **7 — all 7 formats**: equivalent. `yaml` differs **only** in `duration_seconds`, and the base binary differs from itself in the same field — checked with a positive control rather than assumed.
- [x] **10 — non-vacuity**: the five mutations above, plus the pre-existing `TestAValueSplitAcrossRunsIsRefusedNotSilentlyLeft` **rewritten** to assert the fix instead of the limitation. Its safety property is not lost: `TestResidueCheckStillRefusesAValueItCannotRemove` now exercises `residueTypes` directly, including a value visible only in the **decoded** rendering, since the end-to-end refusal path is no longer easy to reach and an unreachable guard is one nobody notices has broken.

**Merge test:** clean against **#602** and **#603**. *(My CHANGELOG entry first collided with #603 at the top of 🔒 Security — my own anchor rule — and is re-anchored deeper in 🐛 Bug Fixes.)*

Closes #593.


sectoramen:

|---|---|---|

rustic (approved):

PR #607, opened 2026-09-04
perf(socialmedia): compile the validator's regexes once, not once per match

### LINKED ISSUE #606: socialmedia: the validator recompiles regexes once per match, allocating ~34 KB per finding and ~1,730x the input size
Found while diagnosing the `ubuntu-latest` OOM kills. This is the underlying product inefficiency; the CI symptom is handled separately in #602.

## Measured

`internal/validators/socialmedia`, configured from the shipped `config.yaml`, **without** the race detector — so this is the path a user is on:

| input | matches | allocated | per match | ×input |
|---|---|---|---|---|
| 7.6 KB | 15 | 35.8 MB | — | 4,946× |
| 31 KB | 1,600 | 52.4 MB | 34,351 B | 1,774× |
| **126 KB** | **6,400** | **207.8 MB** | **34,045 B** | **1,732×** |

**2,384,151 allocations for a single `ValidateContent` call** — 372 per match.

## Cause

`regexp` compilation in per-match code paths. From the allocation profile, `regexp.compile` accounted for **522 MB of 861 MB (60%)**, reached via `calculateConfidenceForPlatform` → the per-platform username validators.

Two distinct shapes:

1. **43 inline `regexp.MustCompile` calls over only 16 distinct literal patterns**, in `validateTwitterSpecific`, `validateGitHubSpecific`, `validateLinkedInSpecific`, `validateInstagramSpecific`, `validateFacebookSpecific`, `validateTikTokSpecific`, `validateDiscordSpecific`, `validateRedditSpecific`, `validateCharacterSet`, `findInvalidCharacters`, `isFalsePositiveHandle`. Each is reached once per match, so every finding recompiled several regexes from source.

2. **5 uses of the package-level `regexp.MatchString(pattern, s)`**, which **compiles its pattern on every call** — in `isOverlyBroadMatch` and `isPlaceholderURL`. This is the easier one to miss, because the call site reads like a simple predicate. After fixing (1), the *remaining* 333 MB of a 667 MB profile was all reached through this.

## Why it matters beyond CI

A user scanning a file dense with profile URLs pays this directly — 126 KB in, 208 MB allocated. There is no race detector to blame.

It did also nearly break CI, which is how it was found: under `-race` the same allocation amplifies **~22×** (to ~663 KB per match), so the complexity guard's 4,800-match reading reached 3.0 GB, four readings with GC disabled reached ~12 GB, and `goldencorpus.test` was measured on the runner growing **1.2 GB → 14.4 GB in 45 s**, exhausting 16 GB of RAM and all 3 GB of swap.

## Fix, and what it achieves

Hoist every literal pattern to package scope, compiled once at init. `MustCompile` is correct there: each is a literal, so a failure is a programming error that should panic at startup rather than be silently discarded per call.

| | before | after |
|---|---|---|
| allocated (126 KB input) | 217.4 MB | **84.4 MB** (−61%) |
| allocations | 2,384,151 | **689,928** (−71%) |
| per match | 34,045 B | **13,249 B** (−61%) |
| ×input | 1,732× | **674×** |
| time | 287 ms | **218 ms** (−24%) |
| guard's socialmedia target under `-race` | 6,930–8,142 MB | **3,241 MB** |

**No behaviour change**, proven rather than asserted: 3,087 findings **byte-identical** across 40 fixtures built with deliberately awkward usernames (1–2 chars, 101 chars, `...`, `___`, all-digits, `UC…` channel IDs, `[USERNAME]`, `{user}`, `<name>`, `${env}`, `{{tpl}}`, dotted, dashed, mixed case), plus 0 golden files changed.

## Left alone deliberately

The one genuinely **dynamic** pattern — an email-domain regex built with `regexp.QuoteMeta` in `isPartOfEmailAddress` — cannot be hoisted, and it sits behind an early return (`!strings.HasPrefix(match, "@")`) that a URL match never passes. Caching it would need a design decision about cache growth on attacker-influenced input, and it is not on the hot path measured here.

## What remains after this

674× input is still high, and the profile's next contributors are `getValidationDetails` (132 MB cum) and `getPlatformDisplayName` (60 MB) — string building per match rather than regex compilation. Worth a follow-up, but out of scope for a change whose point is that it provably does not alter output.

sectoramen:
Found while diagnosing the `ubuntu-latest` OOM kills. This is the underlying **product** inefficiency — a user pays it with no race detector to blame.

## Measured

Shipped `config.yaml`, **without** `-race`, so this is the path a user is on:

| input | matches | allocated | per match | ×input |
|---|---|---|---|---|
| 7.6 KB | 15 | 35.8 MB | — | 4,946× |
| 31 KB | 1,600 | 52.4 MB | 34,351 B | 1,774× |
| **126 KB** | **6,400** | **207.8 MB** | **34,045 B** | **1,732×** |

**2,384,151 allocations for one `ValidateContent`** — 372 per match. `regexp.compile` was **522 MB of an 861 MB** profile (60%).

## Two causes

1. **43 inline `regexp.MustCompile` calls over only 16 distinct literal patterns**, inside the per-platform username validators (`validateTwitterSpecific`, `validateGitHubSpecific`, `validateLinkedInSpecific` and eight more) — each reached once per finding, so every match recompiled several regexes from source.

2. **5 uses of the package-level `regexp.MatchString(pattern, s)`, which compiles its pattern on every call.** This is the one that hides: the call site reads like a plain predicate. After fixing (1), the *remaining* **333 MB of 667 MB** was all reached through it.

## Result

| | before | after |
|---|---|---|
| allocated (126 KB input) | 217.4 MB | **84.4 MB** (−61%) |
| allocations | 2,384,151 | **689,928** (−71%) |
| per match | 34,045 B | **13,249 B** (−61%) |
| ×input | 1,732× | **674×** |
| time | 287 ms | **218 ms** (−24%) |

**Side effect on CI:** under `-race` the same allocation amplifies **~22×**, so the complexity guard's socialmedia target falls from **6,930–8,142 MB to 3,241 MB**. That compounds with #602's separate `-race` fix rather than duplicating it.

## No behaviour change — proven, not asserted

- **3,087 findings byte-identical** across 40 fixtures built with deliberately awkward usernames: 1 and 2 characters, 101 characters, `...`, `___`, all-digits, `UC…` channel IDs, and the placeholder forms `[USERNAME]`, `{user}`, `<name>`, `${env}`, `{{tpl}}`.
- `UPDATE_GOLDEN=1` changes **0 files**.

## The regression guard, and why it parses the AST

My first guard **scanned lines, and a re-inlined `MustCompile` survived it** — a bare `}` closing an inner `if` reset its "am I inside a function" flag, so most of every function body was invisible. Brace counting cannot work here either: the patterns themselves contain literal braces (`{22}`, `\{.*\}`) inside backticks, which no counter can distinguish from code.

Rewritten with **`go/ast`**, it caught a site I had missed on the first run. It flags any `regexp.Compile`/`MustCompile` inside a `FuncDecl` body, and any package-level `regexp.MatchString` (distinguished from the method on a compiled value by the selector base being the `regexp` identifier).

**Four functions are exempted by name with a stated reason:**

| function | reason |
|---|---|
| `isPartOfEmailAddress` | pattern embeds `regexp.QuoteMeta(domainPart)`, so it is per-match by construction; also behind an early return a URL match never passes |
| `Configure` | validates configured patterns once at startup |
| `compileOptimizedPattern` | *is* the configure-time compile helper; its result is cached |
| `compileAllowlistPatterns` | compiles the configured allowlist once |

**A stale exemption is itself a test failure**, so one cannot outlive its justification. And the exemptions are not a loophole — with all four in place, an inlined compile in a per-match validator is still reported.

## Test plan

`make pr-checklist BASE=origin/main` → **10 ran, 0 failed, 5 need manual**; all five discharged against a binary built from `origin/main`.

```
12 flake (3 repeats)              RAN — ./internal/validators/socialmedia
```

- [x] **4 — redaction, all 3 strategies**: 40/40 files redacted by both binaries, **0 files still holding a profile URL** in any strategy, with a **sanity control** confirming all 40 originals *do* contain one so the check is not blind.
- [x] **5 — suppression, base-generate / this-binary-apply**: 3,087 rules from base, **3,087 of 3,087 suppressed** under both. No hash-input change.
- [x] **6 — timing / O(n²)**: N/2N/4N with distinct values and `--limit 0`. Findings grow exactly 2× per doubling (**1,600 → 3,200 → 6,400**, non-vacuity floor satisfied) and time is linear and consistently *faster*: 192/311/544 ms base against **177/286/490 ms** here.
- [x] **7 — all 7 formats**: equivalent. `gitlab-sast` differs **only** in `start_time`/`end_time`, and base differs from **itself** the same way — verified with a positive control, not assumed.
- [x] **10 — non-vacuity, 4 of 4 mutations caught**: re-inlining a `MustCompile`; reintroducing package-level `MatchString`; widening a hoisted username rule (the dangerous direction — a looser pattern admits handles the platform rejects); and making an exemption stale.

**Merge test:** clean against **#602**, **#603** and **#604**. This PR touches only `internal/validators/socialmedia/` plus `CHANGELOG.md`.

## Left for a follow-up

674× input is still high. The profile's next contributors are `getValidationDetails` (132 MB cum) and `getPlatformDisplayName` (60 MB) — per-match **string building**, not regex compilation. Out of scope for a change whose point is that it provably does not alter output.

Closes #606.


sectoramen:

|---|---|---|

rustic (approved):

Issue #619, opened 2026-09-09
complexity guard: the O(n²) guard asserts 0 of 18 targets on windows-latest — its 15.625ms CPU tick declines every base


sectoramen:
Found while answering #599 with real per-platform data. The O(n²) complexity guard **asserts nothing at all on windows-latest** — 0 of 18 validator targets, and neither control.

## Measured

Quoted from [run 34281522336](https://github.com/awslabs/ferret-scan/actions/runs/34281522336) on `main`, the `Report the complexity guard's measured ratios` step. Every one of the 18 targets logs a variant of:

```
ssn: growth ratio NOT asserted — base spans only 3.0 ticks of the cpu clock
  (tick=15.625ms), under the 8 required
ipaddress: ... only 1.0 ticks ...
phone: ... only 1.0 ticks ...
```

Base tick spans across the 18: **1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 2.0, 3.0, 3.0, 3.0, 5.0, 6.7, 6.9, 7.0, 7.7** (two on the 722.7µs wall clock, the rest on the 15.625ms CPU clock). Both controls are declined too — `TestGrowthRatioStillCatchesAGenuineQuadratic` at 6.9 ticks and `TestGrowthRatioStaysLowOnLinearCode` at 1.0.

By contrast the same run asserts all 18 on ubuntu (3.30x–4.64x) and macos (3.86x–4.65x).

## The tick gate is right, and that is the point

This is not a gate to relax. Windows' CPU clock advances in **exactly 15.625ms** steps, so a 1-tick base makes the "ratio" one integer over another. `vin`'s raw reading on that run:

```
vin: 4x input took 7.00x longer on the cpu clock
  (min base=15.625ms big=109.375ms over 2 pairs [9.00x 3.50x])
```

`base=1 tick`, `big=7 ticks`. The per-pair samples are **9.00x and 3.50x** — pure quantisation, and the 9.00x sample is above `maxGrowthRatio` (8.0), so without the gate this would **fail correct code**. That is #546 exactly, and it is why #602 added `perfguard.MinTicks`.

So the gate converts a false failure into a silent skip. Correct trade, but the consequence is that **an O(n²) regression merged on a Windows-only code path, or one that only manifests there, has no guard in front of it** — and the guard's own output is the only place that is visible.

## What a fix would need to weigh

[...]

PR #624, opened 2026-09-09
feat(perfguard): a process-wide CPU cycle counter for Windows, and the probe that decides whether to use it

sectoramen:
**Refs #619.** Additive — nothing changes what any guard measures today. This adds the counter and the probe whose Windows numbers decide whether the O(n²) guard should switch to it, because that measurement cannot be taken on any developer's machine.

## The problem

The complexity guard **asserts 0 of 18 targets on windows-latest**. Its CPU-time accounting advances in steps of exactly **15.625 ms**, so every base reading spans 1–7 of them, under the 8 `Growth.Ticks` requires, and the resolution gate declines all 18 plus both controls.

The gate is right, not over-strict. `vin`'s raw reading there:

```
vin: 4x input took 7.00x longer on the cpu clock
  (min base=15.625ms big=109.375ms over 2 pairs [9.00x 3.50x])
```

`base = 1 tick`, `big = 7 ticks`. Per-pair samples of **9.00x and 3.50x** — one integer over another — and the 9.00x is above the 8.0 threshold, so without the gate this **fails correct code**.

## The API, chosen from the documentation

[`QueryProcessCycleTime`](https://learn.microsoft.com/en-us/windows/win32/api/realtimeapiset/nf-realtimeapiset-queryprocesscycletime) returns *"the sum of the cycle time of all threads of the specified process"*, including *"cycles spent in both user mode and kernel mode"* — **exactly** the semantics `getrusage(RUSAGE_SELF)` and `GetProcessTimes` already have. So it would substitute without disturbing any assumption the guard rests on; `AssertNoParallelTests` stays necessary for the reason it is today.

**Deliberately not `QueryThreadCycleTime`**, which is the API that looks right and the one I named in #619's body. It is per-**thread**, and the Go runtime migrates goroutines between OS threads, so one goroutine's work lands on several and a per-thread count under-reports unpredictably. Correcting it needs `runtime.LockOSThread` around every measurement, which changes the scheduling of the code being measured.

It returns a **count, not a `time.Duration`**. Microsoft: *"Do not attempt to convert the CPU clock cycles ... to elapsed time."* A ratio is unitless so the prohibition does not touch what the guard needs, but a `Duration` would print fabricated milliseconds in every log line and make `ResolutionNote` claim a resolution the counter lacks.

**`timeBeginPeriod` refuted from the documentation, not measured.** It requests a minimum resolution for *periodic timers*, says nothing about CPU-time accounting, *"no longer affects global timer resolution"* since Windows 10 2004, and *"can also reduce overall system performance, because the thread scheduler switches tasks more often"* — it would perturb the thing being measured. One document instead of a CI round-trip.

## How the numbers get out, without touching a file another PR holds

The probe publishes through **`$GITHUB_STEP_SUMMARY`**. This is what makes #619 answerable at all: `go test` hides `t.Logf` for a passing package unless `-v` is set (and `fmt.Println` too, both measured), which is exactly why the issue sat open — the guard has logged its ratios all along and no CI log ever contained one. The job summary needs no workflow edit, so this does not touch `.github/workflows`, which **#620** holds. No-op locally, and an unwritable path does not fail the test.

Four properties, so one Windows run settles the design: the counter advances and never goes backwards; its smallest non-zero delta (against 15.625 ms today); that **sleeping is not charged** — the property a raw invariant TSC would lose, and the whole reason the CPU statistic exists (#579); and that a 4x workload reads ~4x while a quadratic one reads ~16x.

## Three things I got wrong, and how they were caught

**1. `cycleSink` does not prevent elision.** I claimed it did. Measured:

| | base | clock | ratios |
|---|---|---|---|
| as written | 5.094ms | cpu | 4.41x / 15.97x |
| `_ = acc` instead of the sink | **1.047ms** | **wall** | 4.41x / 14.75x |
| body reduced to a constant store | **1.625ms** | **wall** | 3.67x / 15.54x |

Go keeps the loop either way; the sink keeps each iteration *expensive*. And **a ratio bound can never detect this** — a ratio is invariant to how expensive the work is, so bounds of 2.5–6.0x and ≥10x let both mutations through. What changes is the **clock falling back to wall**, which on Windows means a base spanning zero ticks while reporting a tidy 4x/16x. `assertFixtureIsMeasurable` now checks the clock and a 2x-`MinMeasurableCPU` floor; both mutations are caught.

**2. That floor failed the suite on the commit that introduced it.** The quadratic fixture's base was 3.635ms against a 4ms floor. `burnQuadratic` spends most of its budget in loop bookkeeping, so equal nominal unit counts do **not** give equal readings — sizing the two fixtures symmetrically is what put it under. Resized from measurement: quadratic base **33 ms** (8x margin), linear **8.3–9.1 ms**. The linear case mattered for the opposite reason: a *faster* CPU lowers the base, and 1.2x margin is within noise of the value the floor exists to detect.

**3. My mutation harness reverted my own uncommitted fix.** `git checkout -- <file>` in the helper discarded it, so one mutation ran against the old file and read as SURVIVED. Committing first made both CAUGHT.

```
1 gofmt / vet / build             RAN — clean
4 redaction                       RAN — all 3 strategies, 0 leaks, positive control
5 suppression                     RAN — parent-generated rules, cross-binary
6 timing / O(n^2)                 RAN — 5 runs, stable; see the fixture table above
7 all 7 formats                   RAN — byte-identical to the parent binary
10 non-vacuity                    RAN — 5 mutations, 5 caught (2 only after fixing the test)
12 flake                          RAN — 0 failures in 6 consecutive whole-package runs
-- tree clean                     RAN — this run dirtied nothing
=== 12 ran, 0 failed, 0 outstanding ===
```

**Dimensions 4, 5 and 7 rest on a stronger argument than running them piecemeal: the shipped binary is unchanged.** Nothing outside the new files references `ProcessCPUCycles`, `CyclesSupported` or `ErrCyclesUnsupported` (grepped, non-test code), and all seven output formats are **byte-identical** to the parent binary once the timestamp and `duration_seconds` are normalised. Redaction: 0 leaks across `simple`/`format_preserving`/`synthetic`, each with a positive control that the file changed. Suppression: rules generated by the **parent** binary take 5 findings → **0** under mine, matching parent-on-parent.

**Dimension 10 — 5 mutations, 5 caught:**

| mutation | which assertion fired |
|---|---|
| stub returns `(0, nil)` instead of an error | `want ErrCyclesUnsupported` |
| drop the sink, making the fixture cheap | the clock check — fell back to wall at 1.212ms |
| loop body reduced to a constant store | the clock check — wall at 1.302ms |
| make the "quadratic" fixture linear | `want ~16x` |
| claim the counter exists where it does not | `CyclesSupported()` |

## Also

`ErrCyclesUnsupported` lives in an **untagged** file because putting it in the `!windows` one broke `GOOS=windows go vet` with `undefined: ErrCyclesUnsupported` — a break that would have surfaced only on the Windows runner, in a change whose entire point is Windows.

The stub returns an **error, not `(0, nil)`**: a zero count would be divided, or read as a genuine measurement of instantaneous work. `syscall.NewLazyDLL` rather than `golang.org/x/sys/windows`, which is an *indirect* dependency today — promoting it to direct for one call would add a supply-chain edge to `go.mod` to save a dozen lines.

## Conflicts

Union merge-tested with **all six** other open PRs — #616, #617, #618, #620, #622, #623 all merge clean alongside this, the union **builds**, **vets clean under both `GOOS=linux` and `GOOS=windows`**, and `perfguard` + `goldencorpus` + `validators` + `intellectualproperty` + `formatters/sarif` + `cmd` pass together. Only `CHANGELOG.md` is shared; anchor line 113, chosen with ≥6 lines of clearance from the other five ranges by a script, after I collided six times picking them by eye.


sectoramen:
**Correction pushed.** The probe I added here failed windows-latest, and its message was untrue there:

```
linear fixture: 4.06x on the wall clock (base=10.077ms big=40.8777ms)
the linear fixture was measured on the wall clock, not cpu: its base of
10.077ms did not clear MinMeasurableCPU (2ms)
```

**10.077ms plainly does clear 2ms.** Windows reports process CPU time in 15.625ms steps, so a 10ms workload reads as 0 and `Measure` correctly falls back to wall. The fixture was fine — my assertion was **unsatisfiable on that platform**, which is the same defect #625 fixes in the tick-sufficiency guard, reproduced in the change meant to characterise it.

Judged on `BaseWallMin` now: available everywhere and finely resolved on all of them (41ns darwin, 722.7µs windows). The floor is **per fixture**, sized from each one's own measurement, because one shared number goes vacuous against the cheaper of the two:

| | correct | sink dropped | body a constant store | floor |
|---|---|---|---|---|
| linear (8M) | 9.09ms | 2.15ms | 2.48ms | **4ms** |
| quadratic (12M) | 32.71ms | 8.26ms | 8.03ms | **16ms** |

Both clear windows-latest (linear 10.077ms, quadratic ~31ms). Which clock was chosen is now **reported, not asserted** — on windows the CPU clock cannot resolve a 10ms workload at all, and that is #619, not a fixture fault.

**A second mutation then exposed a gap:** lowering the floor from 4ms to 1ns **survived**, because nothing pinned the floor — it was the only thing between the check and vacuity. The base must now stay within 4x of the floor, the same both-directions treatment `MinTicksBudget` gets in #625. Three mutations, three caught.

The other windows failure on this PR (`TestTicksIsSufficientOnThisPlatformForATypicalFixture`) is inherited from `main` and is what #625 fixes.

sectoramen:
**Correcting a claim in this PR's description: I cannot read the numbers back.**

The description says the probe "publishes them through `$GITHUB_STEP_SUMMARY`" so that "one Windows run settles the design". The first half is true and the windows job now passes, but **step summaries are UI-only** — there is no summary field on the Actions jobs API, and `gh run view --log` does not contain them either. Confirmed by looking:

```
gh api repos/.../actions/runs/<id>/jobs --jq ".jobs[0]|keys"   -> no summary field
gh run view --log --job=<windows job>  | grep PROPERTY          -> nothing
```

The second grep returns nothing for the reason this whole issue exists: `go test` hides `t.Logf` for a **passing** package. So the measurements are in the job summary where a human can read them in the browser, and nowhere a script can.

That does not break the channel — it was chosen because `.github/workflows` is held by open PR #620 and this measurement cannot be taken on any developer machine — but it does mean **the numbers need a human glance at the job summary rather than arriving automatically**, and I should not have implied otherwise.

**Better follow-up once #620 merges:** add `./internal/perfguard/` to the existing `Report the complexity guard's measured ratios` step, which already runs `-v` and therefore lands in the log where both a person and a script can read it. The step-summary write can stay as a second channel or be dropped then.

Everything else in the description stands; the windows job on this PR passing is what verifies the probe correction itself.

rustic (approved):

PR #645, opened 2026-09-09
fix(perfguard): choose a clock that can resolve the workload — 15 of 18 O(n^2) targets were asserting nothing on windows

### LINKED ISSUE #644: perfguard: the growth guard picks a clock that cannot resolve the workload, leaving 15 of 18 O(n^2) targets asserting nothing on windows
The growth estimator prefers the CPU clock whenever every base reading clears `MinMeasurableCPU` (2 ms). On windows-latest the CPU clock advances in steps of **15.625 ms**, so a reading of a *single tick* clears that 2 ms floor, the CPU clock is chosen — and `Growth.Ticks` then correctly refuses to divide a 1-tick base.

The result is that the guard chooses a clock that cannot resolve the workload while a clock that can sits unused.

## Measured, one windows-latest run

**15 of the 18** goldencorpus complexity targets declined to assert. Every one of the 15 had a perfectly divisible wall reading:

```
ssn                   base = 1 CPU tick   wall base = 57.0221ms   =  79 wall ticks
socialmedia           base = 1 CPU tick   wall base = 102.6083ms  = 142 wall ticks
cloudresources        base = 1 CPU tick   wall base = 4.2600ms    =   6 wall ticks
...
```

So the O(n²) guard was **83% vacuous** on that platform — it ran, printed a ratio, and asserted nothing. `internal/perfguard/perfguard.go` line ~223 is the choice:

```go
for _, b := range bases {
    if b.CPU < MinMeasurableCPU {   // 2ms, against a 15.625ms tick
        useCPU = false
        break
    }
}
```

The comment above `MinMeasurableCPU` already identifies the hazard ("2ms is 0.128 of a SINGLE TICK on that platform, so it admitted 1-tick readings and divided them") and resolves it by declining to assert. What it does not account for is that declining is unnecessary whenever the wall clock — 20x finer there — can carry the same measurement.

## The wall clock is not merely usable, it is tighter

Recomputing all 18 targets from the same run's wall readings:

```
18 targets: wall ratio min=3.28x  max=4.72x  mean=4.08x   (bound 8.0)
over the bound: 0            margin at the worst target: 1.70x
```

4.08x mean is exactly what a linear validator should measure for a 4x input. The quantised CPU readings for the same run ranged 1.00x–6.00x — both noisier *and* closer to the bound (personname and secrets both read 6.00x, a margin of only 1.33x). Two integer tick counts divided by each other were the less accurate statistic.

## Knock-on: this package's own tests

`internal/perfguard/perfguard_test.go` asserts on growth ratios from a fixed 4-unit fixture and never consults `Growth.Ticks` — the gate this very package defines. 4 units costs ~3.9 ms, which is ~3,900 ticks of darwin's 1 µs CPU clock and **0.25 of one tick** on windows. That is the direct cause of the `TestTheRatioUsesMinimumsNotTheFirstReading` failure seen on PR #642's windows job (a PR touching only `internal/config`):

```
with one 8x-inflated base sample: 1.00x on the cpu clock
  (min base=15.625ms  big=15.625ms  over 4 pairs [0.33x 3.00x 3.00x 3.00x])
```

Base and big had quantised onto the *same single tick*. The minimum-taking was working perfectly; the clock could not show it. Note also that `min(base)` came from a clean pair while `min(big)` came from the inflated one — taking each side's minimum independently pairs readings from different runs, which is only harmless while the clock can resolve them.

Once the clock choice is fixed, sizing that fixture becomes cheap: against windows' 722.7 µs wall tick, `MinTicks` costs 5.78 ms, so the fixture needs **6 units, not 127**.

## Trade-off, stated

The CPU clock is preferred for a real reason: contention only ever inflates wall time, so a descheduled sample cannot fake a quadratic there, and that asymmetry is why this package replaced a wall-clock statistic. Moving windows onto the wall clock does reintroduce that sensitivity.

Weighed against the measurements: today those 15 targets assert **nothing**, so the comparison is not "CPU clock vs wall clock" but "no guard vs a wall-clock guard". The min-of-pairs design already discards one-signed contamination, and the measured margin to the bound is 1.70x. No case gets worse — a target either becomes assertable or stays exactly as silent as it is now.

## Relationship to open work

- **#625** added `MinTicksAffordable(clock, cpuTick, wallTick)`, which answers whether a platform can clear `MinTicks` within budget *given the clock a measurement already used*. It built the predicate; nothing wired it into the choice. Its own numbers agree with this fix: on windows the CPU clock needs 125 ms (over the 50 ms budget) while the wall clock needs 5.78 ms (well under).
- **#643** is a different cause — a macOS *contended base*, where the clock resolves fine (9,200 ticks) and the base reading itself doubled.
- `TestATinyWorkloadFallsBackToWallClock`, which failed 3 of 14 main runs, no longer exists at HEAD (removed in ba3b60f / 93ebe46).

Found while separating inherited CI failures from genuine ones across the open PRs.

sectoramen:
Closes #644

## The defect

`Measure` preferred the CPU clock whenever every base reading cleared `MinMeasurableCPU` (2 ms). windows-latest advances its CPU clock in **15.625 ms** steps, so a reading of a *single tick* cleared that floor, the CPU clock was chosen, and `Growth.Ticks` then correctly refused to divide a 1-tick base.

The guard was choosing a clock that could not resolve the workload while a clock that could sat unused.

**Measured on one windows run: 15 of the 18 goldencorpus O(n²) targets asserted nothing** — the guard ran, printed a ratio and checked nothing. Every one of the 15 had a wall base between 4.26 ms and 102.6 ms, i.e. 6–288 wall ticks.

## The finer clock is also the tighter one

Recomputing all 18 targets from the same run's wall readings:

```
18 targets: wall ratio  min=3.28x  max=4.72x  mean=4.08x     (bound 8.0)
over the bound: 0                 margin at the worst target: 1.70x
```

4.08x mean is what a linear validator should measure for a 4x input. The quantised CPU readings for that same run spanned 1.00x–6.00x — noisier *and* nearer the bound (personname and secrets both read 6.00x, a 1.33x margin). Two integer tick counts divided by one another were the worse statistic.

## What changed

`clockForRatio` (new file `internal/perfguard/clock_choice.go`) prefers the CPU clock **only while it can resolve the workload**, falls back to the wall clock **only on evidence** that the wall clock can, and otherwise returns the historical choice so the existing disclosure is unchanged. Pure, taking both ticks as parameters — for the reason `ticksAt` is pure: on darwin the CPU tick is 1 µs, so the first branch is taken every time and no local run exercises the fallback.

Second commit: `perfguard_test.go` asserted on ratios from a fixed 4-unit fixture and never consulted `Growth.Ticks` — the gate this very package defines. 4 units is ~3.9 ms: ~3,900 ticks on darwin, **0.25 of one tick** on windows. That is the direct cause of the `TestTheRatioUsesMinimumsNotTheFirstReading` failure on #642's windows job (a PR touching only `internal/config`), which reported `1.00x` because base and big had quantised onto the same tick. The fixture is now sized from the measured clock, and the three ratio assertions gate on `Ticks` like every other guard in the repo.

Sizing against the finer clock is what makes this cheap: on windows `MinTicks` costs 5.78 ms of wall time, so the fixture grows to **6 units, not 127**.

## Trade-off, stated plainly

The CPU clock is preferred for a real reason — contention only inflates wall time, so a descheduled sample cannot fake a quadratic there, and that asymmetry is why this package replaced a wall-clock statistic. Moving windows onto the wall clock reintroduces that sensitivity.

Weighed against the measurements: those 15 targets assert **nothing** today, so the comparison is not "CPU vs wall" but "no guard vs a wall-clock guard". The min-of-pairs design already discards one-signed contamination, and the measured margin is 1.70x. No case gets worse — a target either becomes assertable or stays exactly as silent as it is now.

```
1 gofmt / go vet / go build       RAN — clean
12 flake (3 repeats)              RAN — ./internal/perfguard
-- -short mode / tree clean       RAN — ok
```

The five manual dimensions, worked rather than waived:

**4 redaction · 5 suppression · 7 formats — N/A, proven by measurement not assertion.** `internal/perfguard` has **0 non-test importers** and appears in **0 packages of `go list -deps ./cmd/...`**; it is test-only and is never linked into the shipped binary. No detection, redaction or formatter path can observe this change.

**6 timing / O(n²) — this change *is* the timing guard.** The 18-target recomputation above is the substantive evidence. Every consumer package also run green: `goldencorpus` (the 18-target guard, 4.5 s), `detector`, `text-extract-svgtextlib`, `text-extract-rtftextlib`.

**10 non-vacuity — 11 mutants, 11 caught, 0 survivors.** Both new functions are new symbols, so the parent commit cannot execute their tests; the equivalent proof is that each function's *old* behaviour is expressed as a mutant and caught. M2 ("always cpu") is precisely the old rule and S3 ("drop the floor") is precisely the old fixed fixture:

| mutant | caught by |
|---|---|
| M1 remove the wall fallback entirely | TheCoarseClock, Table, ActuallyDiffers |
| M2 always cpu *(= the old rule)* | Table |
| M3 wall fallback taken unconditionally | Table, ActuallyDiffers |
| M4 cpu chosen even when it cannot resolve | TheCoarseClock, Table, ActuallyDiffers |
| M5 judge the wall reading with the CPU tick | TheCoarseClock, Table, ActuallyDiffers |
| M6 drop the MinMeasurableCPU fallback | Table |
| S1 size against the COARSER clock | BaseUnitsFor |
| S2 truncate instead of rounding up | BaseUnitsFor |
| S3 drop the floor *(= the old fixed fixture)* | BaseUnitsFor |
| S4 drop the zero-probe guards | BaseUnitsFor |
| S5 require one tick fewer than MinTicks | BaseUnitsFor |

`TestTheChoiceActuallyDiffersFromTheOldRule` is the standing non-vacuity floor: it replays the 15 measured windows readings and asserts that **14** move off the unusable CPU clock, so a future change that quietly reverts to the old rule fails rather than passing silently. (14 not 15 — at the coarser of the two measured windows wall ticks, 722.7 µs, the 4.26 ms `cloudresources` reading does not clear 8 wall ticks either. Stated as measured rather than rounded up.)

**And the gate is not a silent skip here.** Under both plain and `-race` runs all three gated assertions execute, with 2,926–3,654 ticks of margin — verified in verbose output, since a gate that always skips would turn this fix into the vacuity it removes.

## Conflicts

Base is `main`. Touches `internal/perfguard/perfguard.go`, `perfguard_test.go` and two new files — **no open PR touches any of them** (#624 holds `cycles*.go`, #625 holds `resolution*.go`, both different files in the same package). `resolution.go` was deliberately left alone for that reason, which is why `clockForRatio` lives in a new file. CHANGELOG entry inserted at line 85, ≥14 lines from all 15 insertion points of the other open PRs.


sectoramen:
## Engaging the prior refutation of the wall clock, which the PR body did not

This package's doc comment rejects the wall clock outright, and that objection deserves answering directly rather than being left to the "trade-off" paragraph:

> Measured under 28 external busy-loop processes on 14 CPUs with -race (#579, #581):
> `quadratic median 9.94x` / `linear median 10.20x` — **the populations INVERTED**

If that applied to what this PR does, the PR would be wrong. It does not, and the distinction is a real one rather than a convenient one: **that measurement was the wall clock with a MEDIAN OF RATIOS.** The estimator no longer uses a median. `Measure`'s own table shows why the statistic, not the clock, was doing the damage:

```
                     median of ratios          ratio of minimums
quadratic, idle      13.66x (13.05-14.17)      14.17x
quadratic, loaded    12.48x ( 7.24-15.19)      14.00x     <- 1.2% move
linear,    idle       3.65x ( 2.41- 4.11)       3.75x
linear,    loaded     2.92x ( 1.67- 3.73)       3.18x
```

So the 2×2 is:

| | median of ratios | ratio of **minimums** |
|---|---|---|
| **wall** clock, loaded | INVERTED, 9.94x vs 10.20x → rejected (#579/#581) | never measured |
| **cpu** clock, loaded | moves 9%, worst sample halves | moves 1.2% — the current design |

The bottom-right cell is what the guard runs today. The **top-right cell was never measured** — and it is the one this PR relies on.

## That cell, filled from the same windows run

It turns out not to need a new experiment. The #642 windows job measured both, on a contended hosted runner, with the current min-of-pairs estimator:

```
quadratic control, WALL clock:   23.95x   (min base=2.4975ms big=59.8268ms, per-pair [16.87x 23.95x])
18 linear targets, WALL clock:   3.28x - 4.72x,  mean 4.08x
```

**23.95x against 4.72x is 5.1x of separation, straddling the 8.0 bound with 1.70x of margin below and 3.0x above.** No inversion, and no overlap — on the loaded platform, with the estimator the repo actually ships.

Two honest caveats:

1. **One run.** The 18 linear readings are 18 independent samples on that runner, but the quadratic control is a single measurement. It is the same control #619 quotes, so a second data point arrives with the next windows job on any branch.
2. That quadratic control **declined to assert** on the same run — its wall base spanned 7.0 ticks against the 8 required. It is quoted here as a *measurement*, which it validly is; the second commit's fixture sizing is what makes it clear the gate, and 7.0 → 8+ is the 1.5x bump `baseUnitsFor` produces.

## Relationship to #619 and #624

**#619** is the parent issue and this is a fourth option its body does not list. It weighed `QueryThreadCycleTime`, `timeBeginPeriod`, and declaring the guard a Linux/macOS gate — all three about finding a *finer CPU clock* or giving up. The wall clock was already being measured and reported in every `Reading`, 20x finer than the CPU clock on that platform, and 15 log lines per run were quoting its values while declining to use them.

**#624** is complementary, not overlapping. It states plainly that it is additive — it lands `QueryProcessCycleTime` plus the probe whose Windows numbers decide a later switch, and changes nothing any guard measures today. Different files in the same package (`cycles*.go` vs this PR's `perfguard.go` + new `clock_choice.go`), and the union of all 17 open PRs merges clean and passes every gate.

The two also compose in the right direction: if #624's cycle counter later becomes the Windows CPU clock, it will resolve `MinTicks` easily, `clockForRatio` will simply keep choosing `"cpu"`, and this PR's fallback goes quiet on its own. It is the host for that outcome rather than a competitor to it — which is also why the fallback is written as "prefer CPU while it resolves" and not as "use wall on windows".

rustic (approved):

PR #646, opened 2026-09-09
fix(ci): unblock the three release paths branch protection was holding up, and mirror releases to Test PyPI

sectoramen:
`main` was unprotected on 2026-09-09 so the weekly metrics push and the formula commit work unattended again. Three things follow from that, plus the TestPyPI mirror.

## 1. Homebrew — revert the open-a-PR workaround

`.goreleaser.yml` opened a PR (`branch: chore/brew-formula-{{ .Version }}` + `pull_request:`) instead of committing the formula. That was added because, under protection, goreleaser's direct push was rejected:

```
homebrew formula: could not update "HomebrewFormula/ferret-scan.rb":
409 Could not update file: Changes must be made through a pull request.
```

…and goreleaser then exited 1 **after the release had already published**, so v2.3.3 shipped every asset while the formula on `main` still said 2.3.2 and `brew install` served the previous version. The PR flow fixed that failure and became the only reason a release needs a human.

Now reverted to a direct commit. The comment records the history and says explicitly: **if `main` is ever protected again, restore the PR flow** — a direct push starts failing after the release has published, which is the worst possible ordering.

`brew-smoke` is unaffected either way; it installs the *generated* formula from a workflow artifact, so it validates the release's formula whether or not the commit has landed.

## 2. TLD snapshot — two independent faults

**(a) Drift was decided by IANA's version serial**, which is a publication timestamp bumped roughly daily, not a content hash. So the check reported drift with an identical TLD set. Measured on its first run, 2026-09-07:

```
IANA:     # Version 2026090601, 1438 TLDs
embedded: # Version 2026090400, 1438 TLDs
WARNING: TLD snapshot has drifted from the IANA root zone
```

Same 1438 TLDs; only the header differed. A check that cries wolf weekly is a check nobody reads.

Drift is now decided by the TLD **set**, and it names what moved. Counts alone were not enough either — one TLD delegated and one withdrawn in the same week leaves the count unchanged while the set has moved, and it is the set that decides whether a real address gets capped into the LOW band.

Verified against the live root zone, four cases:

| case | result |
|---|---|
| unmodified (control) | `✅ current (1438 TLDs match the root zone exactly)` |
| a TLD removed from the embedded list | `drifted` — `delegated since the snapshot (1): zone` |
| a bogus TLD added | `drifted` — `withdrawn since the snapshot (1): notarealtld` |
| **one added + one removed, count 1438 vs 1438** | `drifted`, both named — invisible to the old test |

The real-world case now passes cleanly and says why:

```
✅ TLD snapshot is current (1438 TLDs match the root zone exactly)
   (IANA serial has moved on to # Version 2026090900; the TLD set is unchanged, so nothing to do)
```

**(b) The `tld-snapshot` label did not exist**, and `gh issue create --label` hard-fails on a missing label:

```
could not add label: 'tld-snapshot' not found
##[error]Process completed with exit code 1.
```

So no issue was ever filed, the job went red every Monday, and the drift report it had correctly produced sat unread in the log. The dedup query is equally blind, so it could never converge. The workflow now creates the label itself (idempotent), so the check carries its own prerequisite rather than depending on repo state a fork or a tidy-up could remove.

## 3. Test PyPI — mirror every tagged release

**TestPyPI had drifted to 2.4.0 while PyPI was on 2.4.4** — four releases behind — because the only route to it was a manual `pypi-publish.yml` dispatch with `action=publish-testpypi`, which nobody runs on a release. A stale rehearsal index is worse than an empty one: it is what you reach for when reproducing an install problem, and it was quietly answering for a version nobody runs.

Note the correction this rests on: the **live** publisher for a tagged release is `python-package.yml`'s `publish-to-pypi` job, not `pypi-publish.yml` — nothing invokes the latter's `workflow_call`/`is_tag_push` path, and its history shows only `workflow_dispatch` runs. So the mirror goes where the real publish is.

Design, and why each part:

- **A separate job, not a second step.** `environment:` is per-job and PyPI's trusted publishing binds the OIDC token to it. One job cannot be in two environments, so a second upload inside `publish-to-pypi` would present the `pypi` environment to TestPyPI.
- **Parallel with production** (`needs: test-and-release`, not `needs: publish-to-pypi`), so mirroring never delays the real publish.
- **`continue-on-error: true`.** TestPyPI is a convenience index with no availability promise and a *separate* trusted-publisher config; neither an outage nor a misconfiguration there may turn a good release red.
- **The same artefact.** It downloads `python-package-release`, the exact bytes that went to PyPI, not a rebuild that could differ.

Resulting job graph:

```
build → test-and-release → publish-to-pypi      env=pypi
                        └→ publish-to-testpypi  env=testpypi  continue-on-error=true
```

### ⚠️ One manual step needed before this works

TestPyPI's trusted publisher must accept **workflow `python-package.yml`, environment `testpypi`**. The existing TestPyPI uploads came from `pypi-publish.yml`, so the publisher on TestPyPI most likely names *that* workflow. Add a second trusted publisher (or update the existing one) at <https://test.pypi.org/manage/project/ferret-scan/settings/publishing/>.

Both repo environments (`pypi`, `testpypi`) already exist, so nothing is needed on the GitHub side. Until the publisher is added the mirror job will fail auth — and `continue-on-error` means that is visible in the run without affecting the release.

```
=== summary: 9 ran, 0 failed, 0 need manual work ===
1 gofmt / go vet / go build      RAN — clean
2 full suite (-count=1)          RAN — 84 packages ok
3 golden regen                   RAN — 0 files changed
11 cross-platform vet            RAN — linux darwin windows
13 race (whole module)           RAN — 84 packages ok
-- -short mode / tree clean      RAN — ok
4,5,6,7,10                       N/A — 0 production .go files changed
```

Config and shell only; no Go code. Additional verification beyond the checklist:

- **YAML**: all three edited files parse (`YAML.safe_load`); `brews.repository` now has exactly `name, owner` — asserted, not eyeballed.
- **`goreleaser check`**: `configuration is valid`. It exits non-zero only on a **pre-existing** deprecation, `brews should not be used anymore` — not introduced here, and flagged below rather than fixed in a release-path PR.
- **`shellcheck scripts/check-tlds.sh`**: clean apart from SC2018/SC2019 on `tr 'A-Z' 'a-z'`. Suppressed with a reason on the new line: IANA publishes ASCII only (IDNs as punycode), and an explicit range is locale-independent where `[:upper:]` is not. The pre-existing occurrence in `update` mode is deliberately left alone, and the new comparison reuses that exact expression — normalising differently from the list-generator would report every IDN as drift.
- **`make check-tlds`** runs green against the live root zone.

## Noted, not changed here

- **`brews` is deprecated in goreleaser.** Migrating it is a change to the one path that publishes to users, so it wants its own PR and a release to verify against — not this one.
- **`release.yml` commits `CHANGELOG.md` and never pushes it** (zero `git push` in the file), and `.goreleaser.yml` sets `changelog: use: github`, so release notes come from commits and that file is never read. The commit is discarded every release. Harmless, but it means `git-chglog`'s output never lands. Happy to remove it or make it real — it needs your call on which.


sectoramen:
## Added: remove `git-chglog` entirely

Following the discussion on whether to make the `CHANGELOG.md` commit push — **no**, and the reason turned out to be stronger than "it would be untidy".

`.chglog/CHANGELOG.tpl.md` was `{{ range .Versions }}` over **tags**, emitting one line per commit subject, with **no `[Unreleased]` section at all**. So `git-chglog --output CHANGELOG.md` rewrote the entire file. Measured against the current file, that replaces:

```
[Unreleased] holds:  181 lines   114 bullets   ~43,000 words
```

…with commit subjects, and drops everything not yet tagged. Concretely:

```
HAND: - **perfguard:** the growth guard now picks a clock that can RESOLVE the workload rather than
        one that merely clears a floor. windows-latest advances its CPU clock in 15.625ms steps, so a
        single-tick reading cleared MinMeasurableCPU (2ms), the CPU clock was chosen, and…
TPL:  - fix(perfguard): choose a clock that can resolve the workload — 15 of 18 O(n^2) targets were…
```

The subject survives; the measurement, mechanism and reasoning don't.

### Why removal rather than leaving it alone

Two CI jobs generated and committed the file and never pushed, so the damage was discarded with the runner. That made the missing `git push` read as an obvious one-line bug — which is how it came up. The comment in `.gitlab-ci.yml` records the real purpose: *"Commit the changelog to avoid dirty state"*. Not publication; a clean tree for goreleaser — and `--skip=validate` already waives that check, so even that purpose was moot.

**`make changelog` had no such accident protecting it.** `Makefile:735` was `git-chglog --output CHANGELOG.md`, and `make changelog-next` the same with `--next-tag`. Either would have clobbered the file on whoever ran it, today. That is the part that made a partial fix untenable: deleting only the workflow steps leaves the actual footgun in the one place a human would trigger it.

### Also resolves a HIGH audit finding

`.gitlab-ci.yml:279` ran `go install github.com/git-chglog/git-chglog/cmd/git-chglog@latest` — **unpinned** — in a job with a `write_repository`-scoped token in scope. The repo's own audit flagged this HIGH (CWE-829) and noted that the GitHub side had pinned it (`@v0.15.4`, "MED-4") while the GitLab copy never was. Deleting the step resolves it outright instead of pinning it.

### Scope

| file | change |
|---|---|
| `.github/workflows/release.yml` | 3 steps removed, replaced by a do-not-re-add note |
| `.gitlab-ci.yml` | unpinned install + generate/commit blocks removed |
| `Makefile` | `changelog`, `changelog-next` and the help line removed |
| `scripts/install-release-tools.sh` | install / verify / dry-run blocks removed |
| `docs/RELEASE_PROCESS.md` | 7 corrections — it claimed the release used git-chglog |
| `.chglog/` | deleted |

`-189 / +36` lines. The `release.yml` comment deliberately does **not** cite `.chglog/` as an existing path, since this change deletes it.

`docs/RELEASE_PROCESS.md` now leads with the fact that was nowhere in the repo: **CHANGELOG.md is hand-maintained, nothing generates it, add your entry in the PR that makes the change.**

### Verification

```
=== summary: 9 ran, 0 failed, 0 need manual work ===   (make pr-checklist BASE=origin/main)
```

Plus, since this is all config and shell:

- `release.yml` and `.gitlab-ci.yml` both still parse (`YAML.safe_load`).
- `bash -n scripts/install-release-tools.sh` — clean. It was **not** clean at first: removing the verify block left an orphan `fi` and the script failed to parse. Caught and fixed before commit.
- `make help` — zero `changelog` mentions; `make release-snapshot` still resolves.
- No `git-chglog` reference remains anywhere outside the two explanatory comments.

### The real gap, filed separately

`[Unreleased]` has swallowed **27 releases** — v1.7.0 (2026-05-08) is still the newest versioned heading, while v2.4.4 shipped today. So the CHANGELOG cannot answer "what changed in v2.4.4?". Filed as a separate issue: the fix is *promotion* (rename the heading, insert a fresh empty `[Unreleased]`), which preserves the prose, and it now can push since `main` is unprotected. With the caveat that the 27 already-folded releases can't be separated automatically — nothing records which bullet shipped in which tag.

rustic (approved):

Issue #649, opened 2026-09-10
complexity guard: two maxGrowthRatio controls sit under the 1.5x margin convention on macos, and the linear control's 6.40x is unexplained


sectoramen:
Two of the four `maxGrowthRatio` controls sit **under this repo's own ≥1.5x margin convention** on macos-latest. Neither is failing today; both are the next failure in the same class as #620's and #643's.

## Measured, non-race step (the path that still asserts after #648)

```
test                                     platform  reading  bound            margin
TestGrowthRatioStaysLowOnLinearCode      macos      6.40x   must stay <=8.0  1.25x
TestGrowthRatioMissesASparseQuadratic    macos      6.04x   must stay <=8.0  1.32x
TestGrowthRatioCatchesAnEmitPerMatch     ubuntu    14.25x   must exceed 8.0  1.78x
TestGrowthRatioStillCatchesAGenuineQuad  macos     15.69x   must exceed 8.0  1.96x
```

The two "must stay low" controls are the thin ones, and the linear control is the outlier worth understanding: it reads **6.40x on macos-latest against ~4.0x everywhere else**.

## The linear control's 6.40x is not noise, and not a size effect

Its per-pair samples give the mechanism away:

```
macos-latest  6.40x  min base= 7.316ms  big=46.788ms  per-pair [3.87x 6.40x]
ubuntu-latest 3.99x  min base=11.147ms  big=44.508ms  per-pair [3.98x 3.99x]
darwin (dev)  3.97x  min base= 4.627ms  big=18.377ms  per-pair [4.00x 3.97x]
```

Reconstructing macos's two pairs from `min base`, `min big` and the pair ratios: the bases were **~12.5ms and 7.32ms — a 1.7x spread inside one run** — while the two bigs were within 3% of each other. `Ratio = min(big)/min(base)` therefore divides the *cleanest* base into a big that both samples agree on, and reports 6.40x.

There is a second inconsistency in the same numbers. macos's base is 1.6x the dev machine's (7.32 vs 4.63ms), so a uniform slowdown predicts a big of ~29ms. It measured **46.8ms** — 1.6x above prediction, on *both* samples. So the big readings look uniformly inflated on that runner, which a minimum over two samples cannot remove.

**A size/memory threshold is refuted.** `detector.Match` is 232 bytes and the big side emits 160,000 of them (~35 MB) with GC off, so memory pressure was the obvious candidate. Measured on darwin/arm64 across a size sweep:

```
[...]

PR #655, opened 2026-09-10
test(ci): measure the #619 cycle clock and the #649 control margins on the real runners

sectoramen:
Measurement-only branch for the two questions that need CI hardware. Assertion-free, gated on `exp/*` + `FERRET_CI_EXPERIMENT`, `continue-on-error`. Not for merge as-is — the point is the numbers in the job logs.

Refs #619, #649.


sectoramen:
Measurement complete — the numbers are recorded on #619 and #649. Closing rather than merging: the workflow step and the two experiment tests exist only to gather CI data, and both issues now have what they needed. The branch stays available if either question reopens.

PR #703, opened 2026-09-17
test(cmd): guard the ASH integration contract, and pin the pre-commit opt-out ASH depends on

### LINKED ISSUE #702: ASH integration: five environment variables silently change what ASH gets, our next minor release would strand every ASH user, and nothing in this repo referenced ASH at all
## Summary

AWS Labs' [`automated-security-helper`](https://github.com/awslabs/automated-security-helper) (ASH) ships a **first-party plugin that runs this binary as a subprocess and parses its SARIF**:

```
automated_security_helper/plugin_modules/ash_ferret_plugins/
  ferret_scanner.py     1,099 lines — the consumer
  DEVELOPMENT.md          58 KB    — its prose contract
  ferret-config.yaml      46 KB    — the config it ships and passes to us
```

**Nothing in this repository referenced ASH before this issue.** Every CLI-surface change we have shipped went out with no signal that a downstream consumer parses it — and it has already cost them. ASH's own source records it, in the `get_installation_commands` docstring:

> CI ran `pip install ferret-scan`, which resolves to whatever is newest [...] It added an `API_KEY_OR_SECRET` detector that matches `session: Optional[Session]` in generated Pydantic schemas at "95% HIGH" confidence, and **every open PR went red with no source change**.

Their `spec/` delta log records four more rounds of chasing us: the version window, the check inventory (**20 → 19 checks, `KEYWORD_MATCH` dropped**), empty results changing from `null` to `[]`, and narrowing accepted exit codes from `{0,1,3}` to `{0,3}`.

Everything below was measured against **merged `main` (78e8e17)**, which contains PRs #693–#701, using ASH's exact invocation. The contract was read from **ASH's plugin code, not its documentation** — the two disagree, and the code is what runs.

## 1. Live break: five environment variables silently change ASH's results

Pre-commit detection fires **from the environment alone** (`internal/precommit/detector.go`, `detectEnvironment`). It takes no flag and no argument. With ASH's flag set byte-identical and **only the environment changing**:

| environment | exit | `ruleId`s in the SARIF |
|---|---|---|
| *(none)* | **0** | `AWS_ACCESS_KEY`, `VISA` |
| `PRE_COMMIT=1` | **1** | `VISA` |
| `_PRE_COMMIT_RUNNING=1` | **1** | `VISA` |
| `PRE_COMMIT_HOME=…` | **1** | `VISA` |
| `PRE_COMMIT_HOOK=1` | **1** | `VISA` |
| `GIT_HOOK_TYPE=pre-commit` | **1** | `VISA` |

Two independent consequences, from a variable ASH does not set and cannot see:

1. **Exit 1 is outside ASH's `success_exit_codes = {0, 3}`**, so ASH reports the scanner as **FAILED** instead of reporting findings. Total, not partial.
2. **The result set shrinks**, because detection also switches the active profile — `generateOptimizedConfig` sets `ProfileName: "precommit"` and `ExitOnFindings: "high"`. The dropped finding here is `AWS_ACCESS_KEY` at confidence **15**, so this is a confidence filter rather than a HIGH-finding leak. But it means ASH's results depend on ambient environment.

`PRE_COMMIT_HOME` is the one most likely to bite: **`pre-commit` itself sets it**, so any CI image or developer shell with pre-commit installed can carry it into an unrelated ASH run.

**The remedy is one line, on ASH's side.** `FERRET_PRECOMMIT=0` (added for #353) measurably restores exit 0 and the full result set for **all five** signals. ASH should set it in the subprocess environment.

> A correction on my own working: I first measured the opt-out as *not* working. That was a zsh word-splitting mistake in my probe — unquoted `$1` holding `PRE_COMMIT=1 FERRET_PRECOMMIT=0` became a single variable named `PRE_COMMIT` with the value `1 FERRET_PRECOMMIT=0`. Re-measured with proper splitting, the opt-out works completely.

## 2. Our next minor release would strand every ASH user

ASH pins `>=2.4.5,<2.5.0`, and the latest tag is exactly **v2.4.5** (PyPI latest 2.4.5, uploaded 2026-09-11 — the wrapper is genuine and owned by `rustic_aws`/`sectoramen`). So the window admits only 2.4.x.

This is not a cosmetic warning. `get_installation_commands` runs

```
pip install "ferret-scan>=2.4.5,<2.5.0"
```

**unconditionally**, and its docstring says the unconditional install is the point, because *"an already-installed out-of-range build is the case that needs correcting, and pip will downgrade it to satisfy the constraint."*

**Publishing 2.5.0 does not give ASH users a newer ferret-scan — it pins them back to 2.4.5**, and they stop receiving anything we ship afterwards, including leak fixes. Raising ASH's plugin range has to happen *before* our next minor bump, not after.

## 3. ASH ships four config keys we silently ignore

Running ASH's own bundled `ferret-config.yaml` through this binary produces exactly four warnings and exit 0:

| key path in ASH's config | line | consequence |
|---|---|---|
| `preprocessors.text_extraction.types` | 34 | **the one that matters** — ASH lists `pdf` and `office` here to choose which extractors run, and the setting has no effect |
| `redaction.memory_scrub` | 92 | inert |
| `redaction.audit_trail` | 93 | inert |
| `redaction.strategies` | 96 | inert |

The three under `redaction.` are doubly inert, since ASH also blocks every redaction option outright. `preprocessors.text_extraction.types` is the live one.

The mechanism is worth naming separately: **we only *warn* on an unknown config key, and ASH passes `--quiet` and captures stderr** into its own progress rendering. So a renamed key does not fail anything — the user's setting just stops working, in a place they will not look.

## 4. `disabled_types` is inert outside `intellectual_property`

ASH discovered this independently and documented it (`DD-1`: the plan to disable the generic secret type is *"impossible"*, verified on v2.4.5). Confirmed here: `disabled_types` under any validator section other than `intellectual_property` produces **no error, no warning, and no effect**. ASH's workaround was to keep the detector enabled and suppress six false positives on its side.

That is a real defect of ours, not an ASH quirk: a user disabling a type gets silence and no behaviour change.

## 5. `--exclude`, which ASH folds into a single value

ASH concatenates **every** `exclude_patterns` entry plus every `global_ignore_paths` entry into one comma-separated `--exclude`. The whole list therefore shares a fate:

- **A malformed pattern kills the run.** An unmatched `[` — legal in a filename, and `filepath.Match`'s only syntax error — gives `rc=1` and **no SARIF file at all**, with the reason only on the stderr ASH captures. One typo in one entry of a user's ASH config costs the entire scan. *Failing loudly is the right behaviour;* the ask is that ASH validate patterns before folding them.
- **Globstar is a silent no-op.** `**/*.pyc` and `**/node_modules/**` are accepted with `rc=0`, no warning, and no exclusion. Over-scanning is the safe direction for a scanner, but an ASH user writing globstar syntax gets silence, and silence reads as success.
- **ASH's documented `strings.Contains` substring arm no longer exists.** `DEVELOPMENT.md` says `--exclude` does *"PLUS a `strings.Contains` substring match"*, verified against v2.4.5. It does not any more, so ASH's short folded patterns exclude less than its documentation claims. (This one is a docs-vs-code divergence on ASH's side; our current behaviour is the narrower and safer one.)

## 6. Smaller divergences worth recording

- **`SECRET-SECRET-KEYWORD` is a `ruleId` this binary can never emit.** ASH's doc names it. If any ASH suppression or test keys on that string, it matches nothing — and a suppression that matches nothing is indistinguishable from one that was not needed.
- **`tool.driver.name` is `"Ferret Scan"`, not `"ferret-scan"`.** ASH's doc says it matches the literal `"ferret-scan"`. **This is not a live break** — I checked ASH's code and the plugin never reads `tool.driver.name`; it identifies the scanner by its own `name: Literal["ferret-scan"]`. Recording it as latent: if ASH ever did key on `driver.name`, it would fail.
- **The real exit-code surface is `{0, 1, 2, 3}`.** ASH's doc mentions only 1 as our error code, but this repo documents **2** as "the tool failed" and 3 is `exitCodeIncompleteCoverage`. Harmless in effect, since ASH treats everything outside `{0,3}` as failure — but its doc is incomplete.
- **ASH's block list is load-bearing and should stay.** `--pre-commit-mode` is the only mode here that turns findings into a non-zero exit, which is exactly why ASH blocks it. Three entries it still blocks (`memory_scrub`, `extract_text`, `preprocess_only`) are not ferret-scan flags at all.

## What this issue delivers

`cmd/ash_integration_contract_test.go` — ten tests pinning **only what ASH observes**: the flag surface, exit codes, and the SARIF fields it parses.

The contract is **vendored as constants read from ASH's plugin source**, with the ASH line that establishes each. It deliberately does not fetch ASH at test time: CI has no network for it, and a test that skips when a fetch fails is worse than no test. When ASH changes its expectations, these constants are what a human updates.

| test | what breaks if it fails |
|---|---|
| `TestASHEmittedFlagsAllStillExist` | all 23 flags ASH's code emits. Go's flag package exits non-zero on an unknown flag *without scanning*, so one rename means no SARIF for anybody |
| `TestASHRequiresExitZeroWhenSensitiveDataIsFound` | findings must not make the exit non-zero on the ordinary path |
| `TestASHParsedSARIFFieldsAreStable` | version `2.1.0`; `results` an empty **array** and never `null`; non-empty `ruleId` on every result (ASH suppressions key on `(path, rule_id)`) |
| `TestASHAlwaysGetsSARIFRegardlessOfConfigOrProfile` | `--format sarif` beating a *contending* `format: text` in config |
| `TestASHReferencedChecksAreStillAccepted` | the four `--checks` names ASH passes |
| `TestOurVersionIsInsideASHsSupportedWindow` | §2 above, at release time |
| `TestASHBundledConfigProducesNoNewUnknownKeyWarnings` | the *set* of inert keys, so a rename cannot add a fifth silently |
| `TestASHBlockedOptionsThatMakeFindingsFatalStillRequireAnOptIn` | `--pre-commit-mode` staying opt-in |
| `TestPrecommitEnvVarsHijackASHsInvocation` | characterises §1 — fails if we ever fix it, which is correct: ASH's workaround would then be unnecessary and its docs should say so |
| `TestFerretPrecommitOptOutNeutralisesEveryPrecommitSignal` | **protects ASH's remedy.** If `FERRET_PRECOMMIT=0` stopped covering one of the five signals, ASH would keep setting it, keep believing it was protected, and silently report failed scans again |
| `TestASHFoldedExcludeFailureModes` | §5 — that a malformed pattern fails *loudly* rather than being silently ignored |

Every test carries a non-vacuity check, because most of them could pass on an empty probe: the flag test fatals on a truncated `--help`, the exit-code test fatals if the fixture found nothing, the SARIF empty-results subtest fatals if its "clean" fixture produced findings, the config test asserts a **known** key is still recognised, and the pre-commit tests require a multi-finding baseline at `rc=0` and prove each signal actually hijacks before checking the opt-out undoes it.

The version check is deliberately soft — it logs on a dev build (where `-ldflags` has not set `internal/version.Version`) and fails only on a real release outside the window. A guard that fails on every developer machine is a guard that gets deleted.

## Asks

**This issue is closed by #703**, which delivers the guard described above.

Everything that needs a **decision or a code change** rather than a guard has moved to **#704** so it is not closed with this one:

1. whether an explicit `--format`/`--output` invocation should suppress pre-commit auto-detection (§1)
2. `disabled_types` being silently inert outside `intellectual_property` (§4), and the `validators:` subtree being the one blind spot in the unknown-key warning
3. a coverage gap living only in `invocations`, the field an aggregator is most likely to rewrite
4. a nonexistent `--file` path exiting 0 with a SARIF identical to a clean scan
5. a relative `--exclude` pattern containing a path separator silently matching nothing
6. six `ruleId`s we emit that are absent from `docs/checks.md`, which claims to be complete
7. coordinating ASH's version range before the next minor bump (§2), plus `version-helper.sh` being unable to propose a next version at all

#704 also carries the items for the ASH plugin owners, which are not ours to change.

sectoramen:
## Why

AWS Labs' [`automated-security-helper`](https://github.com/awslabs/automated-security-helper) (ASH) ships a **first-party plugin that runs this binary as a subprocess and parses its SARIF**. Nothing in this repository referenced ASH, so every CLI-surface change we shipped went out with no signal that a downstream consumer parses it.

ASH's own source records what that has cost. From the `get_installation_commands` docstring:

> CI ran `pip install ferret-scan`, which resolves to whatever is newest [...] It added an `API_KEY_OR_SECRET` detector that matches `session: Optional[Session]` in generated Pydantic schemas at "95% HIGH" confidence, and **every open PR went red with no source change**.

Their `spec/` delta log records four more rounds of chasing us: the version window, the check inventory (**20 → 19 checks, `KEYWORD_MATCH` dropped**), empty results changing from `null` to `[]`, and narrowing accepted exit codes from `{0,1,3}` to `{0,3}`.

## The live break this found

Pre-commit detection fires **from the environment alone** (`internal/precommit/detector.go`). With ASH's flag set byte-identical and only the environment changing:

| environment | exit | `ruleId`s in the SARIF |
|---|---|---|
| *(none)* | **0** | `AWS_ACCESS_KEY`, `VISA` |
| `PRE_COMMIT=1` | **1** | `VISA` |
| `_PRE_COMMIT_RUNNING=1` | **1** | `VISA` |
| `PRE_COMMIT_HOME=…` | **1** | `VISA` |
| `PRE_COMMIT_HOOK=1` | **1** | `VISA` |
| `GIT_HOOK_TYPE=pre-commit` | **1** | `VISA` |

Exit 1 is outside ASH's `success_exit_codes = {0, 3}`, so ASH reports the scanner **FAILED** instead of reporting findings. The result set also shrinks, because detection sets `ProfileName: "precommit"` (`detector.go:138`) which `cmd/main.go:1259-1265` silently activates. `PRE_COMMIT_HOME` is the one to worry about: **`pre-commit` itself sets it**, so any CI image or shell with pre-commit installed carries it into an unrelated ASH run.

The exit-code half is conditional — the default policy is `ExitOnFindings="high"`, so it needs a ≥90 finding that survives pre-commit's check narrowing; a 9-finding fixture including two at confidence 100 exits 0 because the survivors were lower. The **result-set change is unconditional**. Both are stated at that precision in the test.

**The remedy is one line on ASH's side: `FERRET_PRECOMMIT=0`**, which restores exit 0 and the full result set for all five signals.

## What this adds

`cmd/ash_integration_contract_test.go` — twelve tests pinning **only what ASH observes**: the flag surface, exit codes, and the SARIF fields it parses.

The contract is **vendored as constants read from ASH's plugin code, not its prose**, because the two disagree: `DEVELOPMENT.md` names three hardcoded flags and leaves four spellings unstated, while the code emits **23**. It deliberately does not fetch ASH at test time — CI has no network for it, and a test that skips when a fetch fails is worse than no test.

| test | what breaks if it fails |
|---|---|
| `TestASHEmittedFlagsAllStillExist` | all 23 flags. Go's flag package exits non-zero on an unknown flag *without scanning*, so one rename means no SARIF for anybody |
| `TestASHRequiresExitZeroWhenSensitiveDataIsFound` | findings must not make the exit non-zero on the ordinary path |
| `TestASHParsedSARIFFieldsAreStable` | version `2.1.0`; `results` an empty **array**, never `null`; non-empty `ruleId` on every result |
| `TestASHAlwaysGetsSARIFRegardlessOfConfigOrProfile` | `--format sarif` beating a *contending* `format: text` |
| `TestASHReferencedChecksAreStillAccepted` | the four `--checks` names ASH passes |
| `TestOurVersionIsInsideASHsSupportedWindow` | the release-time version window |
| `TestASHBundledConfigProducesNoNewUnknownKeyWarnings` | the *set* of inert config keys |
| `TestASHBlockedOptionsThatMakeFindingsFatalStillRequireAnOptIn` | `--pre-commit-mode` staying opt-in |
| `TestPrecommitEnvVarsHijackASHsInvocation` | characterises the break above |
| `TestFerretPrecommitOptOutNeutralisesEveryPrecommitSignal` | **protects ASH's remedy** |
| `TestASHFoldedExcludeFailureModes` | a malformed `--exclude` failing *loudly* |
| `TestOnlyShowMatchPutsMatchedValuesIntoASHsSARIF` | the disclosure boundary |

Two deserve their reasoning stated.

**The opt-out guard.** ASH's remedy is invisible from ASH's side. If `FERRET_PRECOMMIT=0` stopped covering one of the five signals, ASH would keep setting it, keep believing it was protected, and silently report failed scans again. Nothing else here ties the opt-out to the five signals it must cover.

**The disclosure guard.** ASH aggregates our SARIF into a report commonly archived as a CI artifact for a whole organisation, and it warns operators about `--show-match` **only** — not about `--explain`, which it also emits. Measured: `--explain` reveals nothing and `--show-match` reveals all three test values, so ASH's warning is on the right flag and only that flag. `--show-match` is the test's **positive control**, so a run that stopped detecting cannot read as "no leak".

## Dim 10, and a vacuous first attempt

**The flag guard was wrong first.** It grepped `--help`. Mutating `flag.Bool("respect-gitignore")` to `"honour-gitignore"` left it reporting `ok` — because this binary prints a **hand-written** help page from `internal/help`, so the documentation and the real flag set are independent things. It was asserting the former while ASH depends on the latter.

It now invokes the binary and looks for Go's own `flag provided but not defined`, which the flag package emits regardless of what we document, plus a **sentinel probe** with a deliberately bogus flag so a run where that message never appears fatals instead of passing. Against the same mutation it now reports:

```
1 flag(s) ASH's plugin passes are no longer REGISTERED: [--respect-gitignore]
```

**The opt-out guard was mutated too**, by making `precommitOptOut` apply only when `PRE_COMMIT` itself is set. It failed and named each newly-unprotected signal *while the `PRE_COMMIT` subtest kept passing* — so it discriminates per signal rather than passing or failing wholesale. Both mutations were confirmed to **compile** first.

Every test carries a non-vacuity check, because most could pass on an empty probe: the exit-code test fatals if the fixture found nothing, the empty-results subtest fatals if its "clean" fixture produced findings, the config test asserts a **known** key is still recognised so "unknown" means something, and the pre-commit tests require a multi-finding baseline at `rc=0` and prove each signal actually hijacks before checking the opt-out undoes it.

**Correction to an earlier revision of this PR.** The version check originally ran the built binary's `--version` and skipped on the dev default, and I described it here as failing "only on a real release outside the window". That was **false**: `buildForASHContract` does a plain `go build` with no `-ldflags`, so `internal/version.Version` is always `0.0.0-development` in a test run and the check **skipped 100% of the time**. A guard that cannot fire is worse than no guard, because the comment stands in for coverage that does not exist.

It now reads the **git tag**, which is what a release publishes — releases are cut by pushing `v*` and goreleaser derives the version from it. Verified both ways: it passes on `v2.4.5` (`tag 2.4.5 is inside ASH's window`) and, on a throwaway local `v2.5.0` tag, fails with the exact consequence.

**And its limitation is now stated in the test rather than implied: it cannot gate a release as CI stands.** `.github/workflows/go-test.yml` checks out with the default depth and fetches **no tags**, and does not run on tag pushes at all; `.github/workflows/release.yml` checks out with `fetch-depth: 0` but runs goreleaser with **no test job and no `needs:` on tests**. So nothing test-shaped can block a tag today. It is effective in a full clone — which is the person cutting the release. Closing that gap is a workflow change, tracked in #704.

`=== summary: 10 ran, 0 failed, 1 need manual work ===`

| dim | result |
|---|---|
| 1 gofmt / vet / build | RAN — clean |
| 2 full suite `-count=1` | RAN — **88 packages ok** |
| 3 golden regen | RAN — **0 files changed** |
| 4–7 redaction / suppression / timing / formats | N/A — **production (non-test) .go files: 0** |
| 8 web / CSP | N/A |
| 11 cross-platform vet | RAN — linux, darwin, windows |
| 12 flake (3 repeats) | RAN — `./cmd` |
| 13 race (whole module) | RAN — 88 packages ok |
| `-short` / tree clean | RAN — ok, dirtied nothing |
| 10 non-vacuity | the **three** mutations above |

Re-run after the version-guard rewrite, on `bad3b9a`, with `production (non-test) .go files: 0`.

The third mutation is the version guard itself, and it is the one that caught my own error: with a throwaway local `v2.5.0` tag on HEAD it fails naming the consequence, and on `v2.4.5` it logs `tag 2.4.5 is inside ASH's window`. Before the rewrite it did **neither** — it skipped unconditionally. The tag was local only and deleted; nothing was pushed.

No open PRs remain, so `main` is the union — this branch is based on `78e8e17` with #693–#701 all merged, and every measurement above was taken against that tree.

## Verified separately, and reassuring

PRs #693–#701 changed **nothing ASH reads**. Twelve generated SARIF documents validate against **ASH's own generated pydantic model** (the exact `SarifReport.model_validate` it calls), and the `ruleId` multiset and `tool.driver.rules` id set are **identical** pre- and post-merge — even though #695/#662 is demonstrably live (14 rule `shortDescription`s and 3 `rank` values changed, 28 diff lines). `--format sarif` won in all 20 profile × config × binary combinations.

## Scope, and what is deliberately not here

This PR pins the contract **as it is today**. It changes no behaviour, and that is the point: a guard that also fixes things cannot tell you whether the fix or the guard is what moved.

The seven findings that need a **decision or a code change** are split into #704 so that closing #702 does not close them — the coverage-disclosure default, a nonexistent `--file` path exiting 0, relative `--exclude` patterns with a separator silently matching nothing, six `ruleId`s missing from `docs/checks.md`, `disabled_types` being inert outside `intellectual_property`, the version-range coordination, and whether an explicit `--format`/`--output` should suppress pre-commit auto-detection at all.

#702 also records three corrections to my own first reading, and one claim that **survived adversarial verification and was still false** (that ASH never passes `--limit`; it does, unconditionally, at `ferret_scanner.py:739` with a default of 0).

Closes #702


sectoramen:
**Heads-up before you merge this: one test in it is now failing, and it is failing correctly.**

`v2.5.0` was tagged and published today from `78e8e17`. `TestOurVersionIsInsideASHsSupportedWindow` reads the git tag, so on any clone that has tags it now reports:

```
the most recent tag reachable from HEAD is 2.5.0, which is OUTSIDE the range ASH's
plugin installs and supports (>=2.4.5,<2.5.0).
```

That is the guard doing exactly what it was built for. ASH's plugin pins `>=2.4.5,<2.5.0` and installs it unconditionally so pip *downgrades* out-of-range builds, which means every ASH user is now resolved back to 2.4.5 and will not receive 2.5.0 — including the redaction and suppression leak fixes from #690, #693, #694, #698 and #699.

**It should stay red until ASH's range is raised.** The remedy is four constants in `ash_ferret_plugins/ferret_scanner.py`, spelled out in #704; when that lands, the vendored constants in `cmd/ash_integration_contract_test.go` get updated in the same change and this goes green.

Two things that make it safe to merge red:

- **It does not fail CI.** `.github/workflows/go-test.yml` checks out with the default depth and fetches no tags, so `git describe` finds nothing and the test skips there with an explicit message. The checklist run recorded above was taken before the tag existed and was green; the full suite in CI is unaffected.
- **It fails only in a full clone** — i.e. for whoever is cutting a release, which is the person who needs to see it.

If you would rather not carry a red test in the tree while ASH catches up, the honest alternative is to widen the vendored `ashMaxSupportedVersion` **only after** ASH actually widens theirs — not before, because the constants are supposed to describe what ASH really accepts, and softening them to go green would make this file fiction. I would leave it red.

rustic (approved):

PR #706, opened 2026-09-18
chore(build): adopt the republished golang:1.27.1-alpine digest (Alpine 3.24.1 → 3.24.2)

sectoramen:
## What happened

`make` was failing for everyone — `all: check-go-version fmt vet build`, and `check-go-version` refused to proceed:

```
❌ Dockerfile digest: sha256:cf6fca66... (registry now serves sha256:4cb7ac97... for 1.27.1-alpine)

❌ The upstream builder image moved: golang:1.27.1-alpine now resolves to a
   different digest than the one pinned in the Dockerfile.
```

That is the pin doing its job (TM-10): a mutable upstream tag was republished under the same name, and the build declines to follow it silently.

## Verified before moving the pin, not after

Adopting a new base image is a supply-chain decision, so the new image was checked against the registry rather than trusted because the failure message offered a one-line fix. No `crane`/`docker`/`skopeo` on the host, so this went through the ECR Public registry API directly.

| | |
|---|---|
| **served digest** | `sha256:4cb7ac979db5fcc41cae44b2227ba5ab8a51e8807f40d9ba4dee20a0ad960b5b` — **computed from the manifest bytes**, not read back from the check's own output, so a stale or misreported value could not carry the decision |
| **Go version** | `GOLANG_VERSION=1.27.1`, `GOTOOLCHAIN=local` — **unchanged**, verified on `linux/amd64` *and* `linux/arm64/v8` |
| **platforms** | amd64 and arm64/v8 both present, plus six others. Those two are what `docker-multiarch.yml` builds |
| **blast radius** | the previously pinned digest **still returns HTTP 200** with all eight platforms, so nothing was broken — this was drift, not an outage, and the v2.5.0 images built fine on the old base |

## What actually changed: exactly one of ten build steps

```
step 0: ADD alpine-minirootfs-3.24.1-x86_64.tar.gz /   →   3.24.2-x86_64.tar.gz
step 1: CMD ["/bin/sh"]                                    SAME
step 2: RUN apk add --no-cache ca-certificates             SAME
step 3: ENV GOLANG_VERSION=1.27.1                          SAME
step 4: ENV GOTOOLCHAIN=local                              SAME
step 5: ENV GOPATH=/go                                     SAME
step 6: ENV PATH=...                                       SAME
step 7: COPY /target/ /                                    SAME   ← the Go toolchain
step 8: RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" ...       SAME
step 9: WORKDIR /go                                        SAME
```

Same on `aarch64` (`3.24.1-aarch64` → `3.24.2-aarch64`). The Go install step is byte-identical, which is the strong signal: **the toolchain layer was not rebuilt, only the OS beneath it.** Alpine 3.24.1 → 3.24.2 is a patch release, and republishing `golang:X.Y.Z-alpine` on a new Alpine patch is the ordinary reason this digest moves.

## What is NOT verified here

**That the new base actually builds this project.** There is no container runtime on this host, so the image was not built locally, and I am not going to imply otherwise.

`docker-multiarch.yml` runs on `pull_request` with `Dockerfile` in its `paths`, so the build on this PR is the real proof, and **it passed in 3m47s**. The log confirms the new base was genuinely used, not cached past:

```
#2 [internal] load metadata for public.ecr.aws/docker/library/golang:1.27.1-alpine@sha256:4cb7ac97...
#5 [builder  1/15] FROM public.ecr.aws/docker/library/golang:1.27.1-alpine@sha256:4cb7ac97...
```

**Correcting an overclaim in the first revision of this description:** I wrote that this job proves "amd64 and arm64". It does not. The workflow builds **`linux/amd64` only on pull requests** — *"Only build for amd64 on PRs for faster feedback"* — and takes `linux/amd64,linux/arm64` only on main and tags.

So the honest state is:

| platform | Go version verified via registry | image actually built |
|---|---|---|
| `linux/amd64` | yes — `GOLANG_VERSION=1.27.1` | **yes, this PR, passed** |
| `linux/arm64/v8` | yes — `GOLANG_VERSION=1.27.1` | **not yet** — happens on merge to main |

The arm64 residual risk is low: its manifest is present in the served index, its `GOLANG_VERSION` and `GOTOOLCHAIN` are identical to amd64's, and its only differing build step is the same Alpine bump (`3.24.1-aarch64` → `3.24.2-aarch64`). But it is unbuilt until this merges, and that is worth knowing rather than glossing.

## Scope note

The final stage is `FROM scratch`, so the builder base is a **build-provenance** question rather than a runtime CVE surface in the published image. That lowers the urgency of the Alpine bump itself; the pressing part was that `make` was blocked for every developer.

## Test plan

Diff is **one line** in `Dockerfile` (plus a CHANGELOG entry) — no Go code, so the dimensions that exercise detection, redaction, suppression and output formats have nothing to act on.

| check | result |
|---|---|
| `make check-go-version` | **8/8 pins consistent**, was 7/8 with the digest failing |
| `make` (`check-go-version fmt vet build`) | passes end to end; **was failing before this** |
| `go vet` cross-platform | darwin, linux, windows — all ok (via `make vet`) |
| full `make pr-checklist BASE=origin/main` | see the comment below |
| multi-arch container build | **deferred to this PR's `Multi-Architecture Docker Build` job** — the only place it can actually run |

Fixes the `make` failure on `main`.


sectoramen:
## Test plan results

`make pr-checklist BASE=origin/main` on `72abf26`, run in a detached worktree:

```
=== scope ===
  CHANGELOG.md
  Dockerfile

=== summary: 9 ran, 0 failed, 0 need manual work ===
```

| dim | result |
|---|---|
| 1 gofmt / vet / build | RAN — clean |
| 2 full suite `-count=1` | RAN — **88 packages ok** |
| 3 golden regen | RAN — **0 files changed** |
| 4–8 redaction / suppression / timing / formats / web | N/A — no production Go code changed |
| 10 non-vacuity | N/A — no test files added |
| 11 cross-platform vet | RAN — linux, darwin, windows |
| 13 race (whole module) | RAN — 88 packages ok |
| `-short` / tree clean | RAN — ok, dirtied nothing |

Nothing needed manual work, which is unusual and is simply because a one-line base-image digest change cannot move detection, redaction or output.

**The container build is the check that matters, and it passed** (`docker`, 3m47s). Two things I verified in its log rather than inferring from the green tick:

1. The new base was genuinely resolved and used — `#5 [builder 1/15] FROM ...@sha256:4cb7ac97...`. A cache hit on the old layer would have made a pass meaningless.
2. **It built `linux/amd64` only.** PRs deliberately restrict to amd64 for speed; arm64 builds on merge to main. I have corrected the description above, which claimed both — the registry data for arm64 is verified but the image is not yet built.

Local gates for completeness: `make check-go-version` now reports **8/8 pins consistent** (was 7/8), and `make` completes end to end — it was failing for every developer before this, since `all` depends on `check-go-version`.

rustic (approved):

Issue #704, opened 2026-09-18
ASH integration follow-ups: seven ferret-scan decisions the contract guard deliberately does not make


sectoramen:
Split out of #702, which reported the ASH integration contract and is closed by the guard in #703. Everything here is a **decision or a defect on our side that the guard does not fix** — #703 pins the contract as it is today, deliberately; these items change it.

All measurements were taken against `78e8e17` (PRs #693–#701 merged), using the flag set ASH's plugin actually sends.

## 1. A subprocess consumer is steered by ambient environment

Pre-commit detection fires from environment variables alone (`internal/precommit/detector.go`, `detectEnvironment`): `PRE_COMMIT`, `_PRE_COMMIT_RUNNING`, `PRE_COMMIT_HOME`, `PRE_COMMIT_HOOK`, `GIT_HOOK_TYPE`. With ASH's flag set byte-identical and only the environment changing, the exit code goes 0 → 1 and the result set shrinks.

`PRE_COMMIT_HOME` is set by `pre-commit` itself, so any CI image or developer shell with pre-commit installed carries it into unrelated runs.

**The mechanism, precisely** (this took a correction to get right): the finding loss is **profile substitution**, not the pre-commit config block. `detector.go:131-138` drops nothing and does not even win on format — `cmd/main.go:1354` applies its `"text"` default only when `!isFlagSet("format")`. The actual path is `detector.go:138` setting `ProfileName: "precommit"`, which `cmd/main.go:1259-1265` silently activates.

**Decision needed:** should an explicit `--format` / `--output` invocation suppress auto-detection? A caller that names its output format and destination is not a git hook. `FERRET_PRECOMMIT=0` already exists as an opt-out (#353) and works, but it requires the consumer to know it exists — and `#703`'s `TestFerretPrecommitOptOutNeutralisesEveryPrecommitSignal` exists only because that remedy is invisible from the consumer's side.

Note the exit-code half is conditional: the default policy is `ExitOnFindings="high"`, so it needs a finding at confidence ≥90 that also survives the check-set narrowing. A 9-finding fixture including two at confidence 100 exits **0**. The result-set change is unconditional.

## 2. `disabled_types` is silently inert outside `intellectual_property`

Under any of the other 18 validator sections it produces **no error, no warning, no SARIF notification, no stats key, and no effect**. `validator.go:444` is the sole non-test read. Identical on HEAD and the merged tree.

[...]

Issue #637, opened 2026-09-09
redaction: a literal backslash in a POSIX filename collapses two output paths, losing one redacted copy and mislabelling the other


sectoramen:
`makeRelativePath` rewrites every backslash to a forward slash unconditionally. On POSIX a backslash is a **legal filename character**, so two different scanned files map to one output path: one redacted copy is lost, and the survivor is **labelled with the wrong file's name**.

## Reproduced

A tree holding both `a\b.txt` (a single file whose name contains a backslash) and `a/b.txt` (a file inside directory `a`), plus two controls, all with findings:

```console
$ ferret-scan --file tree --recursive --checks ssn --confidence all \
    --enable-redaction --redaction-strategy simple --redaction-output-dir out
rc=0, 0 bytes of stderr, 0 lines mentioning a collision
findings: 4
redacted copies written: 3
```

Reading the outputs back (escape-safe, via `cat -v` — see the note at the end):

```
out/.../tree/x:y.txt  -> colon-file       correct
out/.../tree/a/b.txt  -> backslash-file   WRONG DOCUMENT
out/.../tree/y.txt    -> plain-y          correct
                         nested-file      GONE
```

**The path `out/.../tree/a/b.txt` names `a/b.txt` and contains the content of `a\b.txt`.** The redacted copy of the real `a/b.txt` does not exist. An operator forwarding that file believes it is the redacted `a/b.txt`; it is a different document. A mislabelled artefact is worse than a missing one, because nothing about it looks wrong.

## The code

`internal/redactors/output_manager.go:118`:

```go
// Convert backslashes to forward slashes for consistency
path = strings.ReplaceAll(path, "\\", "/")
```

The intent — normalise the platform separator — is right; the spelling is unconditional. `filepath.ToSlash` is the same intent correctly expressed: it rewrites the separator on Windows, where a backslash **is** one, and leaves a POSIX path alone.

## This function already documents the same class of bug

Immediately below the offending line:

> Neutralise a `..` SEGMENT, not every occurrence of two dots. This was `strings.ReplaceAll(path, "..", "parent")`, which rewrites the middle of a FILENAME. Until #562 no path containing `..` ever reached a scanner … so the substitution was unreachable for real input and the collision below could not happen.

Same class, still live: a whole-string substitution applied to a path, producing a collision between two legitimate names. The `..` case was fixed once its inputs became reachable; the backslash case never needed anything to become reachable, because a backslash in a filename has always been legal on POSIX.

[...]

PR #707, opened 2026-09-18
fix(gitlab-sast): stop collapsing an absolute finding path to its basename

### LINKED ISSUE #705: gitlab-sast: every finding is reported at the repository root
With `--format gitlab-sast`, every finding's `location.file` is just the filename, whatever directory the file is actually in.

On a tree with `src/first/config.py` and `src/second/config.py` (one SSN each), `--file . --recursive --confidence all` at `056518b`:

| format | paths |
|---|---|
| `json` | `src/first/config.py`, `src/second/config.py` |
| `gitlab-sast` | `config.py`, `config.py` |

```json
"location": { "file": "config.py", "start_line": 1, "end_line": 1 }
```

GitLab resolves `location.file` from the repository root, so the dashboard link points at a file that isn't there, and the two findings are indistinguishable.

As far as I can tell it's this in `internal/formatters/gitlab-sast/mapper.go`:

```go
if filepath.IsAbs(cleaned) {
    cleaned = filepath.Base(cleaned)
}
```

and it hits every finding, not just some, because `cmd/main.go` runs each input through `filepath.Abs` before walking it, so the formatter only ever sees absolute paths.

I think the formatter needs to know the root and use `filepath.Rel` against it. Two things I'd want to be careful about, based on what's already in the tree: the root probably shouldn't come from `CI_PROJECT_DIR` inside the formatter given #704, and a path that isn't under the root shouldn't be an error, since the mapper's errors get skipped by the loop in `formatter.go` and #562 covers what that looks like.

Happy to put up a PR.

2018-lonely-droid:
Closes #705

When I scan a repo with `--format gitlab-sast`, every finding comes out with `location.file` set to the bare filename. The mapper does `filepath.Base` on any absolute path, and since `cmd/main.go` resolves inputs with `filepath.Abs` before walking, every path the formatter sees is absolute. So on a tree with `src/first/config.py` and `src/second/config.py` (one SSN each), `--file . --recursive` gives:

| format | paths |
|---|---|
| `json` | `src/first/config.py`, `src/second/config.py` |
| `gitlab-sast` | `config.py`, `config.py` |

GitLab resolves `location.file` from the repo root, so the dashboard link goes nowhere and the two findings collapse into one.

This passes the working directory as `FormatterOptions.SourceRoot` from the CLI and has the mapper use `filepath.Rel` against it, forward-slashed. Same scan now gives `src/first/config.py` and `src/second/config.py`.

## Changes since the first push

Mostly from the review below, plus a couple I'd already started on:

- **Root comes from the caller.** I originally read `CI_PROJECT_DIR` in the formatter with an `os.Getwd` fallback. Given #704 that seemed like the wrong place for it, so `cmd/main.go` sets it and the formatter has no environment access. `CI_PROJECT_DIR=/nonexistent` now changes nothing.
- **Out-of-root paths degrade instead of dropping.** The first revision returned a `MappingError`, which the loop in `formatter.go` skips past. On this repo's `docs/` (262 findings) scanned from `docs/metrics` that gave 9 vulnerabilities at `status: success`; from a sibling dir, 0. Now those paths fall back to the basename they always had, and both invocations give 262, matching `--format json`. The test for this asserts the count before the path.
- **Symlinked roots.** Lexical `Rel` first, then both sides through `EvalSymlinks` if that fails, same as `withinRoot` in `cmd`. Lexical first because resolution can succeed on one side and not the other. `CI_PROJECT_DIR=/tmp/sym/link` against `/tmp/sym/real` now gives `src/app.py`.
- **Backslash in a POSIX filename left alone.** Dropped the `ReplaceAll`; `filepath.ToSlash` alone. `we\ird.txt` stays `we\ird.txt` on Linux/macOS.
- **Nested golden fixture.** The corpus fixtures were all flat, so `SourceRoot` in the harness wasn't testing anything. Added `file_nested_path` two directories down; reverting the fix now fails the golden. `writeFixture` creates the parent dir for it.

Tests cover nested paths, out-of-root (sibling, parent, no root), a string-prefix lookalike root (`/repo` vs `/repo-archive/x.py`), symlinked root in both directions, relative paths, `<stdin>`, and the separator handling. Reverting to `Base`, swapping `Rel` for `HasPrefix`, erroring on out-of-root paths, and reintroducing the backslash rewrite each fail a test.

`go test ./...` (89 packages, including `TestABlockingRunAlwaysDisclosesWhy`), `go vet`, `gofmt` clean.

Two things from the review I didn't take on here: a documented override for the root outside GitLab CI, and the `sarif`/`gitlab-sast` path disagreement that comes from `cmd/main.go` absolutizing every input. Both seem like separate issues; happy to file them.


sectoramen:
Thanks for this — **the bug you found is real, your diagnosis is correct, and it is worse than #705 claims.** I reproduced everything and the core fix works. There is one blocker that has to be resolved before merge, and it is a small change.

Everything below is measured against two binaries built from `main` and from your head `6ed5e6b`, driven through the CLI.

---

## What you got right

**#705's core claim, confirmed.** Nested paths collapse, and duplicate basenames become genuinely indistinguishable:

```
tree: probe705/src/nested/config.py   (AWS key)
      probe705/lib/config.py          (SSN)

main:  config.py, config.py            → 1 distinct location for 2 findings
#707:  probe705/src/nested/config.py
       probe705/lib/config.py          → 2 distinct, correct
```

**The intended GitLab CI path works exactly as designed:**

```
$ CI_PROJECT_DIR=/tmp/scantarget ferret-scan --file /tmp/scantarget --recursive --format gitlab-sast
main:  files=['config.py', 'other.py']
#707:  files=['src/nested/config.py', 'src/other.py']     ← correct, status=success, all 3 findings
```

**The interface change is clean.** One implementer, one caller, no mocks left stale, and `FormatterOptions` is not public API — `pkg/scan` and `pkg/redact` never touch it, so adding a field breaks no consumer. `IsVirtual()` still bypasses normalization, so stdin scans are unaffected. Sibling directories sharing a prefix are handled correctly rather than mangled. And your three tests are non-vacuous — I checked `TestAbsoluteLocationOutsideSourceRootIsRejected` with a mutation and it is the one test that pins the new error return.

---

## 🚫 Blocker: findings are dropped from the report, silently

`MapToGitLabVulnerability` now returns an error for a path outside the source root, and the formatter's handler is:

```go
vuln, err := f.mapper.MapToGitLabVulnerability(match, sourceRoot)
if err != nil {
    errorCount++
    if options.Verbose { log.Printf(...) }
    continue          // ← the finding leaves the report
}
```

`errorCount` is only ever printed under `--verbose`, so by default the finding just disappears. On a secrets scanner that inverts the tool's purpose: **only reported findings reach the consumer, so a dropped finding is a cleartext value nobody is told about.** That is categorically worse than a wrong path.

**The dangerous variant is a *partial* drop, and it is completely silent.** Run from a subdirectory while scanning the parent — an ordinary thing to do:

```
$ cd /tmp/part707/sub && ferret-scan --file /tmp/part707 --recursive --format gitlab-sast --limit 0

main:  exit=0  scan.status='success'  vulns=3  files=['in1.txt','in2.txt','out1.txt']
#707:  exit=0  scan.status='success'  vulns=2  files=['deep/in1.txt','in2.txt']
                                       ↑ the credit-card finding in ../other/ is gone
       stderr: '' (empty)
```

Exit 0, `status: "success"`, nothing on stderr. The pipeline is green and a finding is missing.

The *total* drop is at least signalled — scanning an absolute path wholly outside cwd gives `vulns=0` with `scan.status='failure'` (3 → 0 on a tree holding an AWS key, an SSN and a card). But it is still exit 0, so CI passes.

**Your own CI would have caught this.** The repo has a guard for exactly this failure mode, and it is red on your branch and green on `main`:

```
main:   ok      github.com/awslabs/ferret-scan/v2/cmd
#707:   --- FAIL: TestABlockingRunAlwaysDisclosesWhy/blocking/gitlab-sast
        gate_integrity_guard_test.go:289: the run BLOCKED (rc 1) and its gitlab-sast
        output names no finding and no file.
```

Its fixture lives in `t.TempDir()`, which is outside cwd, so every finding is dropped while the run still exits 1 — "the report asserts the run was clean while the commit was rejected", which is the sentence that guard was written to prevent. The PR description lists `make test` under Testing; I think that run did not actually complete, because the full suite fails: **88 packages ok, 3 FAILED.** Worth re-checking on your side — note GitHub has not run CI here at all, because the PR is from a fork and needs maintainer approval.

### Why this happens, and the fix

The root cause is the half that is missing rather than anything wrong in what you wrote. `#705` asked to *"pass the scan root to the formatter"*, and you added exactly the right field for it — but **nothing in production ever sets it:**

```
$ grep -rn 'SourceRoot' --include='*.go' .
PROD  internal/formatters/formatter.go:59          (declaration)
PROD  internal/formatters/gitlab-sast/formatter.go:76  (reads it)
TEST  .../repository_relative_path_test.go:75
TEST  internal/goldencorpus/golden_file_test.go:46
```

All three CLI construction sites — `cmd/main.go:1939`, `cmd/stdin.go:369`, `cmd/stdin.go:585` — leave it empty, so every real invocation falls through to `CI_PROJECT_DIR` or `os.Getwd()`. The issue even warned against the cwd fallback specifically.

Two changes make this safe, and I'd suggest both:

1. **Set `SourceRoot` from the resolved scan root** in `cmd/main.go` (and the stdin paths). The scan root is by definition an ancestor of every scanned file, so `filepath.Rel` can't escape it and the drop becomes unreachable. Watch the single-file case — `--file a/b.txt` needs `filepath.Dir`, or `Rel` yields `"."`.
2. **Never drop a finding on a path problem.** Degrade instead: fall back to the old basename (or the cleaned absolute path) and keep the finding. A wrong path is a bad link; a missing finding is a missed secret. If you want the mismatch visible, surface `errorCount` unconditionally rather than only under `--verbose`.

---

## Other findings

**`major` — a symlinked or differently-spelled root drops everything.** `filepath.Rel` is purely lexical, so a root that is equivalent but spelled differently fails:

```
$ CI_PROJECT_DIR=/tmp/sym707/link  ferret-scan --file /tmp/sym707/real ...
main:  vulns=2  files=['app.py']
#707:  vulns=0  status='failure'
$ CI_PROJECT_DIR=/tmp/sym707/real  ferret-scan --file /tmp/sym707/real ...   # control
#707:  vulns=2  files=['src/app.py']    ← works when spelled consistently
```

This is not exotic: macOS `/tmp` is a symlink to `/private/tmp`, and container volume mounts commonly introduce one. `filepath.EvalSymlinks` on both sides before `Rel` would close it. Case-only differences on a case-insensitive filesystem fail the same way.

**`major` — re-introduces #637.** The final line rewrites backslashes unconditionally:

```go
return strings.ReplaceAll(filepath.ToSlash(cleaned), `\`, "/"), nil
```

On Linux/macOS `ToSlash` is a no-op, so this rewrites backslashes in *POSIX filenames*, where a backslash is a legal character. Measured on one real file named `we\ird.txt`:

```
main:  location.file = 'we\ird.txt'     ← correct
#707:  location.file = 'we/ird.txt'     ← invents a directory that does not exist
```

[#637](https://github.com/awslabs/ferret-scan/issues/637) fixed precisely this class in the redaction path — *"a literal backslash in a POSIX filename no longer collapses two output paths into one"*. `filepath.ToSlash` alone is correct and already platform-aware; the `ReplaceAll` should go.

**`minor` — the golden gate was adjusted rather than exercised.** `golden_file_test.go` gains `SourceRoot: tmpDir`, which makes the harness agree with the new code instead of testing it. The fixtures are flat, so `Rel(root, root/f)` returns the basename either way and the goldens can't move — meaning the corpus can't see this defect class at all. A nested fixture would make it bite.

**`minor` — `CI_PROJECT_DIR` is now load-bearing for correctness and nothing documents it.** Outside GitLab CI there is no way to set the root at all: no flag, no config key, no doc. Fixing (1) above mostly removes the need, but a documented override would help anyone embedding the formatter.

**Root-cause note, pre-existing and not yours:** `cmd/main.go:1624` absolutizes every input path, which is why no formatter ever receives a repository-relative path. `sarif` has the same underlying shape, so after this lands `sarif` and `gitlab-sast` will disagree about the same finding's path. Worth a follow-up issue rather than expanding this PR.

**One review claim I withdrew:** I initially thought nothing exercised the new error return. That was wrong — `TestAbsoluteLocationOutsideSourceRootIsRejected` does, confirmed by mutation. The real gap is narrower: there is no **formatter-level** test asserting how many vulnerabilities come out when a path is outside the root, and that is the level the drop lives at.

---

## Suggested tests

Two would have caught the blocker, and both belong at the formatter level:

- scan a tree from a subdirectory and assert the vulnerability **count** is unchanged from the count reported for the same tree by `--format json`
- a source root reached via a symlink, asserting the count is preserved

Plus the two the issue asked for that aren't covered yet: Windows separators (the `ToSlash`/`ReplaceAll` line is untested), and a relative path containing a climbing segment.

---

**Verdict:** the diagnosis and the core fix are right, and I'd like to merge this once findings can no longer be dropped. The two changes in "Why this happens, and the fix" are small and I'm happy to help with either — say the word if you'd prefer I push them onto your branch.

sectoramen:
## Follow-up: the blocker measured on a real tree, not a toy fixture

My review above used a 3-finding fixture. I re-measured on this repository's own `docs/` tree (262 findings on `main`) because the magnitude changes how the blocker reads. All four figures below are from the CLI, `--config /dev/null --limit 0`, base vs your `6ed5e6b`.

**The silent partial loss is the one to fix.** Scanning `docs/` while cwd is a subdirectory of it — `docs/metrics` — which is an entirely ordinary thing to do:

| | exit | `scan.status` | vulnerabilities | stderr |
|---|---|---|---|---|
| main | 0 | `success` | **262** | 0 bytes |
| #707 | 0 | `success` | **9** | 0 bytes |

**253 of 262 findings deleted, 96.6%, reported as a successful scan with nothing on either stream.** No `--verbose`, no non-zero exit, no `scan.status` change. A GitLab dashboard shows green with 9 findings where there are 262.

The total-loss variant, scanning `docs/` from a sibling cwd, is at least signalled — `262 → 0` with `scan.status='failure'` — though still exit 0, so the job passes.

**Two things that sharpen the diagnosis:**

**The tool's own remedy routes users straight into it.** `cmd/path_traversal.go:94` prints:

```
path climbs above the working directory; pass an absolute path instead
```

with the comment *"An absolute path always passes the gate."* That advice is correct for the traversal gate and, after this change, walks the user into the 100%-loss case. Anyone who hits the traversal refusal and does what the tool tells them gets an empty report.

**The blast radius is confined to `gitlab-sast`,** which is worth knowing for scoping the fix — `--format sarif` reports **262 on both binaries** for the identical sibling-cwd invocation. So nothing in the scan or detection layer changed; this is purely the new path resolution in the one formatter. That also means `sarif` and `gitlab-sast` will disagree about the *existence* of 253 findings for the same scan, not merely about their paths.

None of this changes my recommendation or the fix — clamp instead of erroring, and wire `SourceRoot` from the scan root. It only changes the urgency: at 96.6% silent loss this cannot go in as-is, and with those two changes I think it's mergeable.

One process note in fairness: I ran seven independent review passes and then adversarially re-checked each finding. That pass **refuted one criticism outright** — I had been told nothing exercised your new error return, and that was wrong; `TestAbsoluteLocationOutsideSourceRootIsRejected` does, which I confirmed by mutation. It also downgraded 16 of 36 findings for stating conditional failures as unconditional. I've only sent you what I reproduced myself.

2018-lonely-droid:
Thanks for the thorough pass, and for reproducing on `docs/`. You're right about the blocker and about `make test`; I'd run the formatter packages and not the full suite. My mistake.

Pushed `bd2b094`, which replaces the earlier commit:

- `SourceRoot` is now set from `cmd/main.go` (working directory) and the formatter no longer reads `CI_PROJECT_DIR` or calls `os.Getwd`. I went with cwd rather than the resolved scan root because with several `--file` inputs there isn't one root, and GitLab resolves against the checkout anyway; happy to change that if you'd prefer the scan root.
- Out-of-root paths fall back to the basename instead of erroring. Your `docs/metrics` and sibling-cwd cases both give 262 now, matching `--format json`, and `TestABlockingRunAlwaysDisclosesWhy` passes.
- Symlinked root: lexical `Rel` first, then both sides through `EvalSymlinks` like `withinRoot` does. Your `/tmp/sym/link` case gives `src/app.py`.
- Dropped the `ReplaceAll`; `filepath.ToSlash` only. `we\ird.txt` stays as is on POSIX.
- Added a nested fixture to the golden corpus (`file_nested_path`, two dirs down) so the harness actually sees this. Reverting the fix fails the golden.

Tests added for each of those, and I checked that each of the four regressions you listed fails a named test when reintroduced.

I left the documented-override and the `sarif`/`gitlab-sast` path disagreement alone for now since they seemed like their own issues; I can open them if that's useful.

sectoramen:
## Re-reviewed `bd2b094` — every item verified fixed, by measurement. This looks good to me.

I re-ran the exact battery from my earlier comments against a binary built from your head, versus one from `main`. All four regressions are gone, the original fix is intact, and the suite is clean.

| what I reported | before | `bd2b094` |
|---|---|---|
| partial silent loss (`docs/` from `docs/metrics`) | 262 → **9**, `status=success` | **262 → 262** ✅ |
| total loss (`docs/` from a sibling cwd) | 262 → **0**, `status=failure` | **262 → 262** ✅ |
| symlinked root (`/tmp/sym/link` vs `/tmp/sym/real`) | 2 → **0** | **2 → 2**, and `src/app.py` ✅ |
| #637 backslash (`we\ird.txt` on disk) | `we/ird.txt` | **`we\ird.txt`** ✅ |
| `TestABlockingRunAlwaysDisclosesWhy/blocking/gitlab-sast` | FAIL | **ok** ✅ |
| full suite | 88 ok / **3 FAILED** | **89 ok / 0 FAILED** ✅ |

The #705 fix itself is intact — duplicate basenames go from `distinct=1/2` to `distinct=2/2` with correct nested paths. And the symlink case is now *better* than `main`, which reported `app.py`.

**Your golden fixture is non-vacuous, which I checked rather than assumed.** Reverting `relativeToRootOrBase` to the basename collapse (keeping it compiling) fails `TestGoldenFileScanFormats/file_nested_path` on `file_nested_path.gitlab-sast.json`, plus `TestAbsolutePathsInsideTheRootKeepEverySegmentBelowIt` and `TestARootReachedThroughASymlinkStillContainsItsFiles`. Exactly the gate that was missing before.

---

## I gave you bad advice, and you were right to push back

You asked whether to use the working directory or the resolved scan root, and said you'd change it if I preferred the scan root. **Don't — your choice is correct and my suggestion was wrong.** Scanning a subdirectory from the repo root:

```
$ ferret-scan --file ./dq/src --recursive --format gitlab-sast
main:        ['config.py']
bd2b094:     ['dq/src/nested/config.py']     ← correct for GitLab
```

With the scan root instead, `Rel("./dq/src", "./dq/src/nested/config.py")` gives `nested/config.py` — which GitLab would resolve against the checkout root and get wrong. Using the scan root would have reintroduced a variant of #705 for every subdirectory scan. The working directory is the checkout in GitLab CI, and it is the right anchor.

One small correction to your reasoning, which doesn't change the conclusion: `--file` isn't repeatable — passing it twice honours only the last (`--file ./dq/src --file ./dq/other` reported only `dq/other/x.txt`). So "several `--file` inputs" isn't the constraint; the subdirectory case above is the real reason, and it's a stronger one.

**Dropping the `CI_PROJECT_DIR` read was the right call too,** and I verified it: output is now byte-identical with and without `CI_PROJECT_DIR` set. A formatter whose report contents are steered by an ambient env var is the shape #704 is filing against, so moving the decision to the caller is the better design, not just a smaller diff.

---

## Two things worth recording, neither blocking

**The fallback is lossy, and that is the correct trade — just so it's stated.** When cwd is not an ancestor of the scan target, all 262 findings are reported but **every one by basename** (0 with a directory segment). So #705 remains unfixed for that invocation, degraded to pre-#705 output rather than losing data. Your code comment says exactly this, which is why I'm noting it rather than objecting: *"A finding at the wrong path can still be found; a finding that is not in the report cannot."* That's the right principle and it's the one I'd want applied.

**`cmd/stdin.go`'s two `FormatterOptions` sites don't set `SourceRoot`,** which is harmless today — stdin findings are virtual and bypass normalization (`location.file = "<stdin>"`, verified). Worth a one-line comment there someday so nobody "fixes" it by wiring a root into a virtual path.

Yes please on both follow-up issues — the documented override and the `sarif`/`gitlab-sast` path divergence. The second is the more interesting one: `cmd/main.go` absolutizes every input before any formatter runs, so `sarif` has the same underlying shape and will now disagree with `gitlab-sast` about the same finding's path. That's a pre-existing scan-layer gap, not something this PR should grow to cover.

---

Thanks for the fast, precise turnaround, and for owning the `make test` slip — that was a useful signal, not a problem. Note CI still hasn't run here because the PR is from a fork and needs maintainer approval, so the 89/0 above is from my local run of your head.

**Recommending merge.** Leaving the formal approval to @sectoramen.

rustic:
Thank you for this contribution to Ferret Scan. This highlighted a few other discrepancies as to how we report file locations in other output formats and the web UI. We will merge this PR and the forthcoming PRs to correct those formatters at the same time before a version bump. As we are currently at version 2.5.0, we also need to reach out to ASH to increase the  MAX_SUPPORTED_VERSION to greater than 2.5.0.

sectoramen (approved):

